PR #934, opened 2013-08-31
Waf build system

bjourne:
This PR is more of an idea and a proof of concept and not ready to be merged.

The suggestion is to replace Factors build system based on batch and shell scripts with waf: https://code.google.com/p/waf/ waf is a "real buildsystem" like autotools and both much faster and easier to maintain than hand-coded makefiles and scripts. The difference becomes more apparent when you add advanced build features, like making distribution tarballs, automated test running, window installer and installation targets. It's really hard writing all that by hand. Currently the wscript I've written only builds and installs, but I plan to add more features to it.

waf only requires Python and the script itself is included in the pr. So when you have applied it, you can build using:

```
python waf.py --prefix=/blah configure
python waf.py build
sudo python waf.py install
```


razetime:
is this still a thing? the lik to waf here doesn't work, and i'm not able to find it in any ordinary search engine.

catb0t:
> is this still a thing? the link to waf here doesn't work, and i'm not able to find it in any ordinary search engine.

https://waf.io/

This PR has probably bitrotted pretty bad in the last 9 years, I think modern Waf might be somewhat different, and Waf is still maintained https://gitlab.com/ita1024/waf/activity

mrjbq7:
Waf is cool, but not going to merge this in present form -- we need to get 0.99 released. Closing.

PR #1011, opened 2014-03-08
ui.text.pango: have pango wrap lines longer than 10 000 pixels

bjourne:
If you run something in the graphical listener that prints long strings, such as `"http://www.reddit.com" http-get* .`, the listener crashes with a cairo-error. It happens because cairo doesn't want to create surfaces that are wider than 32767 pixels. So this patch fixes that by just enabling pangos line wrapping.


mrjbq7:
For now, fixed by 94cbe2b4796fd97d99e8ab32244f1fd3f7362721.

PR #1046, opened 2014-04-27
Moving help lint

bjourne:
The code added to check for leaks caused an impossible to solve circularity in the bootstrap from help.lint -> tools.destructors and then back to help.tips. You would have to change the order in which `bootstrap/stage2.factor` loads vocabs but then you would likely just cause more dependency cycles between other vocabs. 

So if the mountain won't come to Muhammed, Muhammed has to come to the mountain. So I moved the help.lint vocabs to tools.help-lint. This fixes the bootstrap issues. Which I don't think is so bad because a help lint tool sounds (to me) as a better fit in `tools`  than `help` anyway.


mrjbq7:
I fixed this a little bit differently, by moving the ``TIP:`` from ``tools.destructors`` to ``help.tips``, which fixes the cycle.

a33fc84de7fa5189e1ff1135736beed9753e33cc

mrjbq7:
Anyway, merged, thanks!!!

bjourne:
sounds good to me!

PR #1091, opened 2014-07-04
Fix portability nits

sstewartgallus:
Also set a language standard for the GNUMakefile build and separate
CFLAGS and CXXFLAGS.


erg:
Some of these have been applied.

PR #1179, opened 2014-10-31
Refactor read request

bjourne:
So I was trying to get the http server to server a webapp for me and then I found that it is childs play to break it. By forging malicious requests you can get the server to crash and/or hang. E.g factorcode.org could probably easily be crashed if it is not behind a web proxy.

To fix that, I made a new vocab http.server.requests with all the request parsing moved into there. I've tried to make it safer, eg it shouldn't crash if a `string>number` call fails. The idea is that if someone makes a naughty request, that vocab will throw a `request-error` and the server will log and respond with 400 Bad Request which is consistent with what other web servers do. 

Other errors will still be bubbled up like they are now, which I think is nice because then if the web server logs `ERROR`, you know there is a bug in the code, but if it logs `NOTICE httpd-bad-request:` you know the client sent a bogus request.

### DIFF
--- a/basis/furnace/actions/actions-tests.factor
+++ b/basis/furnace/actions/actions-tests.factor
@@ -1,4 +1,4 @@
-USING: kernel furnace.actions validators tools.test math math.parser
+USING: kernel furnace.actions io.crlf validators tools.test math math.parser
 multiline namespaces http io.streams.string http.server http.server.requests
 sequences splitting accessors ;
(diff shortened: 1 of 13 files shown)


mrjbq7:
Reminder to myself that I want to look at this later.

mrjbq7:
Merged! Still not sure about the ``mime.multipart`` change, going to look at that in more detail later.

PR #1205, opened 2014-11-18
Test suite for Factor's C++ code

bjourne:
Here is a small test suite for the VM I've been working on. Testing that things like garbage collection works is often too hard to do in Factor code. I based it on googletest (https://code.google.com/p/googletest/) because it seemed simple enough (although there are 5e9+ other c++ test frameworks to choose from).


mrjbq7:
So I like this idea but I noticed hard coded paths to ``gtest``. Is there a way to make this more flexible or is there standard installation paths that are fine to use?

bjourne:
I'm not sure. You can override the paths using `make GTEST=/path/here`. My setup in the pr was just a convenient hack and to set it up correctly I think you need something more elaborate.

PR #1220, opened 2014-11-23
Added new option "thiscommit" to allow building using current commit whe...

tgunr:
...n trying to use git bisect.


bjourne:
Should we reject this pr then? No reason to keep it open if it has no chance of going in.

mrjbq7:
I guess close it for now and we can revisit it - the use case seems useful but @erg made a good point about boot images.

Issue #1187, opened 2014-11-05
VM crash on split-slice


bpollack:
In the Factor GUI, this crashes the VM hard, leaving a stack trace into the garbage collector:

```factor
USE: splitting
"Hello world" split-slice
```

whereas, in the CLI, the same code simply throws a stack underflow error as expected.  I'm not familiar enough with the VM to debug this much further, but happy to provide more details if you tell me what you need.

PR #1271, opened 2015-01-19
Fix 1187 attempt3

bjourne:
Here is another fix for the really annoying #1187 bug. E.g

    10 [ [ minor-gc split-slice ] [ drop ] recover ] times

doesn't crash. :) So the problem is that there may be uninitialized holes in the stacks. Which works fine until a `##peek` instruction looks at a stack slot above the known height and causes a stack underflow. Then the GC code tries to trace all slots on the stacks and that crashes because of the uninitialized slots. It's the same kind of problem that the gc maps solves.

So I've added a step compiler pass called `clear-uninitialized` which finds `##peek`s that can cause underflows and inserts `##replace-imm` instructions preceding those to initialize the stack slots with dummy values. I also had to detach the `normalize-height` and `schedule-instructions` passes because the way they reordered the instructions caused to many uninitialized slots which caused to many ##replace-imm's to be inserted. 

There is some overhead, but not to much I think. The mean size per word increases from 386 to 391 bytes on 32 bit and 459 to 460 on 64 bit.

### DIFF
--- a/basis/compiler/cfg/finalization/finalization.factor
+++ b/basis/compiler/cfg/finalization/finalization.factor
@@ -2,15 +2,13 @@
 ! See http://factorcode.org/license.txt for BSD license.
 USING: compiler.cfg.build-stack-frame compiler.cfg.gc-checks
 compiler.cfg.linear-scan compiler.cfg.representations
-compiler.cfg.save-contexts compiler.cfg.scheduling
-compiler.cfg.ssa.destruction compiler.cfg.stacks.vacant
-compiler.cfg.utilities compiler.cfg.write-barrier ;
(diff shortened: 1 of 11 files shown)


bjourne:
You can see the effect of the pr:

```factor
USE: compiler.cfg.debugger
[ dup ] regs.

! Old code generation
...
=== Basic block #1

##inc-d 1
##peek RCX D 1
##replace RCX D 0
##branch

Successors: 2
...

! New code generation
...
=== Basic block #1

##peek RCX D 0
##inc-d 1
##replace RCX D 0
##branch

Successors: 2
```

erg:
Merged, let's see if everything still works. Thanks!

PR #1274, opened 2015-01-28
Refactor debug

bjourne:
I noticed that the build bot fails with the error `critical_error: Bad symbol specifier in compute_dlsym_address: b5e437d7` so this pr tries to improve on that by making the error message better.

The first (big) commit adds an `ostream& out` printing methods because that makes them easier to use when building error message. The second refactors `store_external_address` in such a way so that it outpus the name of the word that caused the bad symbol specifier error which should make it easier to debug. E.g `critical_error: unix.ffi:getservbyport: Bad symbol specifier in store_external_address: 7f88469c9a39`


mrjbq7:
Merged, thanks!

PR #1276, opened 2015-01-30
Remove collapse from sequences.extras

bpollack:
collapse has the same functionality as filter-as with a reversed conditional,
so there's no reason to keep it.


bpollack:
Bah, I'm wrong, this is subtly different.  Beefing up docs for `compact` and `collapse` instead and abandoning this PR.

mrjbq7:
@erg wrote it so he's a good one to bug if you're curious about it.

PR #1278, opened 2015-02-03
Make fuel more discoverable when browsing Factor help

bpollack:
Fuel's one of Factor's best parts, but it's almost totally undiscoverable. This starts towards fixing that.


mrjbq7:
Merged, thanks!

PR #1280, opened 2015-02-05
Stack checker docs tests

bjourne:
Here is some docs and tests I've written for the stack-checker hierarchy of vocabs.


mrjbq7:
Merged, thanks!

PR #1282, opened 2015-02-11
Fix typo and add find-library links in alien docs

bpollack:
This is really minor, but...yeah.


mrjbq7:
Merged, thanks!

Unfortunately, ``find-library`` is awesome but had the problem of locating the library on the build machine so that when we distributed binaries for 0.97, they included the path on the build machine rather than the path on the deployed machine.  I made a patch that makes a version of ``add-library`` that is like an ``update-library``, but haven't pushed it yet.

PR #1291, opened 2015-02-28
Gc fixes afl

bjourne:
Here is a bunch of gc refactorings I've been working on. The `trace_cards` function is rewritten so that you get rid of the goto. And I've also replaced some bitshifts with easier to read multiplications and divisions. It is an optimization that doesn't make sense anymore because these days it is as fast or even faster to multiply or divide instead of bit shifting.

### DIFF
--- a/vm/collector.hpp
+++ b/vm/collector.hpp
@@ -99,35 +99,34 @@ template <typename TargetGeneration, typename Policy> struct collector {
     }
   }
 
-  inline cell first_card_in_deck(cell deck) {
+  cell first_card_in_deck(cell deck) {
     return deck << (deck_bits - card_bits);
   }
 
-  inline cell last_card_in_deck(cell deck) {
+  cell last_card_in_deck(cell deck) {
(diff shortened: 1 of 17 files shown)


vm/collector.hpp
```diff
@@ -90,123 +76,100 @@ template <typename TargetGeneration, typename Policy> struct collector {
...
-  inline cell first_card_in_deck(cell deck) {
-    return deck << (deck_bits - card_bits);
-  }
-
-  inline cell last_card_in_deck(cell deck) {
-    return first_card_in_deck(deck + 1);
-  }
-
-  inline cell card_deck_for_address(cell a) {
-    return addr_to_deck(a - data->start);
```
mrjbq7: Do you want to remove the ``addr_to_deck`` function if not used anymore?

mrjbq7: Actually, maybe leave it for now -- its documentation even if unused.

mrjbq7:
Merged, thanks!

PR #1294, opened 2015-03-08
Correct DLL name typo in GObject

hyphz:
Not sure why GitHub thinks the previous pull wasn't also applied..


erg:
Applied the docs and the DLL gobject patches. Thank you!

PR #1309, opened 2015-04-24
FUEL: better font lock highlighting for some factor constructs

bjourne:
Here is some simpel fuel fixes for some syntax highlighting regexps that were wrong.


mrjbq7:
Merged, thanks!

PR #1310, opened 2015-04-30
Gc fixes afl part3

bjourne:
This should fix the 32bit test failures the previous pr caused. The windows build crashes in bootstrap but maybe that's just some stale files or something?


basis/compiler/constants/constants-docs.factor
```diff
@@ -9,7 +9,45 @@ HELP: rt-decks-offset
 { $description "Relocation offset type for the decks table." }
 { $see-also rel-decks-offset } ;
 
+HELP: string-offset
+{ $values { "n" integer } }
+{ $description "hm" } ;
```
mrjbq7: :-)

mrjbq7:
Merged, thanks!

PR #1314, opened 2015-05-08
Add a couple minor extra docs for math.extras

bpollack:
This is really minor, but I realized it was still sitting on my computer while I was waiting to do more, so, here you go.


mrjbq7:
Merged, thanks!

PR #1315, opened 2015-05-09
Gc fixes afl part4

### LINKED ISSUE #1289: gc crash when reading from files
Here is a minimal example of the crash Doug found when stress-testing with afl:

```factor
USING: accessors arrays byte-arrays formatting io io.buffers
io.encodings.binary io.files io.private kernel memory
namespaces sequences sequences.private ;
IN: test3
QUALIFIED: unix.ffi

: some-alloc ( y -- w )
    drop 80 f <array> ;

: my-read-ifd ( -- ifd )
    7 read 2 read 4 read 4 read
    19 f <array>
    2nip 2nip ;

: my-read-ifds ( tiff -- )
    input-stream get
    [ buffer>> 0 >>fill 0 >>pos drop ]
    [ handle>> fd>> 256 unix.ffi:SEEK_SET unix.ffi:lseek drop ] bi
    2 read drop
    13 [
        my-read-ifd
    ] replicate
    some-alloc drop
    555 f <array>
    over [ length ] [ set-nth ] bi
    my-read-ifds ;

"/etc/passwd" binary <file-reader>
[
    V{ } clone my-read-ifds
] with-input-stream
```
The code runs an infinite loop which allocates more and more memory so it should end when Factor can't grow the heap further. Instead it crashes:

    assertion "(parent->current_gc && parent->current_gc->op == collect_growing_heap_op) || parent->data->seg->in_segment_p((cell)obj)" failed: file "../vm/collector.hpp", line 28

bjourne:
Here is another pr to try and fix #1289. I've created a new vocab `compiler.cfg.stacks.padding` which more precisely traces which stack locations are live so you get more accurate gc maps. It supersedes `compiler.cfg.stacks.map`, but I've kept the old one in the history so it's easy to go back to that one if something doesn't work. But it should work. :) You probably need new boot images too. 

I've also written a fix for the "Generic word lookup-base-pointer*..." test failure that the build machines got. And some general improvements to the cpp code too.


basis/compiler/cfg/builder/blocks/blocks.factor
```diff
@@ -35,11 +35,12 @@ IN: compiler.cfg.builder.blocks
 : call-height ( #call -- n )
     [ out-d>> length ] [ in-d>> length ] bi - ;
 
+: emit-call-block ( word height -- )
+    dup adjust-d ##call, make-kill-block ;
```
mrjbq7: Isn't this wrong, since its ``( word height -- )``, this calls ``adjust-d`` and ``##call,`` on height?  When it should instead be:

```[ ##call, ] [ adjust-d ] bi* make-kill-block```

?

File: basis/compiler/cfg/ssa/destruction/destruction.factor [hunk 2]
```diff
@@ -132,6 +131,7 @@ M: insn cleanup-insn , ;
 PRIVATE>
 
 : destruct-ssa ( cfg -- )
+    H{ } clone leader-map set
```
mrjbq7: Does it need to be a ``H{ }`` or can you just call ``leader-map off``?

basis/compiler/cfg/stacks/clearing/clearing.factor
```diff
@@ -1,11 +1,10 @@
 USING: accessors arrays assocs combinators.short-circuit
 compiler.cfg.instructions compiler.cfg.registers compiler.cfg.rpo
-compiler.cfg.stacks.map kernel math sequences ;
+compiler.cfg.stacks compiler.cfg.stacks.padding kernel math sequences ;
 IN: compiler.cfg.stacks.clearing
 
 : state>replaces ( state -- replaces )
-    [ stack>vacant ] map first2
-    [ [ <ds-loc> ] map ] [ [ <rs-loc> ] map ] bi* append
+    [ stack>vacant ] map { ds-loc rs-loc } [ swap create-locs ] 2map concat
```
mrjbq7: Does the order of this matter, previously it would be all ds's before all rs's.  Now it is ds/rs alternating?

basis/compiler/cfg/stacks/clearing/clearing.factor
```diff
@@ -20,6 +19,6 @@ IN: compiler.cfg.stacks.clearing
     [ [ clearing-replaces ] keep suffix ] with map V{ } concat-as ;
 
 : clear-uninitialized ( cfg -- )
-    [ trace-stack-state ] keep [
+    [ trace-stack-state2 ] keep [
```
mrjbq7: Is there a better name for this, or is intended to replace the old one and eventually be just called ``trace-stack-state2``?

basis/cpu/architecture/architecture.factor
```diff
@@ -234,7 +234,7 @@ HOOK: %replace-imm cpu ( src loc -- )
 HOOK: %inc cpu ( loc -- )
 
 HOOK: stack-frame-size cpu ( stack-frame -- n )
-HOOK: %call cpu ( word -- )
+HOOK: %call cpu ( word height -- )
```
mrjbq7: Does any implementation of ``#call`` use the height argument? It looks like they always pass ``0`` and always ``drop`` the height?

vm/collector.hpp
```diff
@@ -97,18 +100,12 @@ template <typename TargetGeneration, typename Policy> struct collector {
   void trace_partial_objects(cell start, cell card_start, cell card_end) {
     object* obj = (object*)start;
     cell end = start + obj->binary_payload_start();
-    if (card_start < end) {
-      start += sizeof(cell);
-
-      start = std::max(start, card_start);
-      end = std::min(end, card_end);
+    start += sizeof(cell);
```
mrjbq7: Is ``card_start < end`` always true?

basis/compiler/cfg/ssa/destruction/destruction.factor
bjourne: missed that!

basis/compiler/cfg/stacks/clearing/clearing.factor
bjourne: It's still ds before rs

vm/collector.hpp
bjourne: No, but if `card_start >= end` then it must be that `start >= end` and then `visit_object_array` won't visit anything.

bjourne:
Fixed some small things. The `emit-call-block` is right because I've added a height slot to `##call`. You need it to know how much a call increases or decreases the datastack. It's not needed in `%call` as you noted, but the parameters given to the %-word needs to be the same as the slots in the ##-tuple due to some metaprogramming. Yes the idea is that `trace-stack-state2` will become `trace-stack-state` if the new scheme works as it should.

mrjbq7:
Okay, merged! Thanks!!

bjourne:
Damnit. Something seem to break when you run the benchmarks. I'll make a small patch to go back to the old stack tracing.

mrjbq7:
What issue are you seeing with the benchmarks?

bjourne:
`benchmark.terrain-generation` crashes if you compile factor in debug mode.

PR #1316, opened 2015-05-10
sequences.extras: add start-all and count-subseq for searching/counti…

jonenst:
Add words for searching/counting all possibly-overlapping or non-overlapping subseqs of a seq. help-lint ok

This uses start which uses the naive approach, not a more complex algorithm like the ones from http://en.wikipedia.org/wiki/String_searching_algorithm


mrjbq7:
Merged, thanks!

(We also have a ``boyer-moore`` vocabulary, not sure about the other algorithms).

PR #1320, opened 2015-05-16
Shutdown quot

bjourne:
I noticed that the shutdown quot special object is not used. So I think you can safely remove it. It doesn't affect the normal shutdown hooks which are still run by `do-shutdown-hooks`.


mrjbq7:
Looks like @erg added it in 110b310c5409284af256b492e62c96d8d7fb4858, not sure what the original intention was...

erg:
What's the problem? Shutdown hooks are the opposite of startup hooks.

bjourne:
No problem. I just think the code is unused because the shutdown hooks are already handled directly by the `system:exit` word.

mrjbq7:
@erg was your thinking that ``factor_vm::stop_factor`` would be something we'd call from the C API?

erg:
Yes, I was going for symmetry. If it's really useless we could remove it.

On Sat, May 16, 2015 at 6:11 PM, John Benediktsson <notifications@github.com

mrjbq7:
I feel like the bug here is that shutdown-hooks seem to only run if the Factor word ``exit`` is called?

Perhaps we should call ``factor_vm::stop_factor()`` from ``factor_vm::primitive_exit()`` and remove the ``do-shutdown-hooks`` from ``exit``.

Thoughts @erg?

Are there cases where we want to ``(exit)`` where we don't want shutdown-hooks to run?

mrjbq7:
Maybe look into how ``die()`` (and ``factorbug()``), since it seems to exit without running shutdown hooks.

mrjbq7:
We can't call ``stop_factor()`` from ``primitive_exit`` without trapping errors in shutdown hooks then just exiting like ``system:exit`` does...

PR #1321, opened 2015-05-17
Gc fixes afl part5

bjourne:
Another attempt at fixing the gc maps. :) The tracing is inverted so that it now tracks the dead locations and not the live ones. It's very similar to how the original `compiler.cfg.stacks.uninitialized` worked but with stack height tracing added. There is also a new `##clear` instruction to make it easy to see where in the cfg the compiler inserts the clearing replaces.

### DIFF
--- a/basis/compiler/cfg/builder/blocks/blocks.factor
+++ b/basis/compiler/cfg/builder/blocks/blocks.factor
@@ -36,7 +36,7 @@ IN: compiler.cfg.builder.blocks
     [ out-d>> length ] [ in-d>> length ] bi - ;
 
 : emit-call-block ( word height -- )
-    dup adjust-d ##call, make-kill-block ;
+    adjust-d ##call, make-kill-block ;
 
 : emit-primitive ( node -- )
     [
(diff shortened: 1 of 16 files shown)


mrjbq7:
Is the only remaining cleanup the ``trace-stack-state`` vs ``trace-stack-state2``? Or is there other unused stuff as part of the patches?

Merged, thanks!

mrjbq7:
Looks like the clearing tests fail, I fixed them in c7ea41987406b9812459a51949d5be31c2f9cde3 if you want to make sure its what you expect.

``"compiler.cfg.stacks.clearing" test``

mrjbq7:
@bjourne I get this error with a ``load-all`` in DEBUG build.

```
assertion "(parent->current_gc && parent->current_gc->op == collect_growing_heap_op) || parent->data->seg->in_segment_p((cell)obj)" failed: file "vm/collector.hpp", line 21
Abort trap: 6
```

bjourne:
Hm yes I see. The `%clear` code was half-baked.

Issue #1319, opened 2015-05-14
Threaded server gc crash


bjourne:
Annoying heisenbug in `benchmark.tcp-echo1`. Here is a small repro with code adapted from Doug's vocab:

```factor
USING: accessors byte-arrays io io.encodings.binary io.servers
io.sockets io.streams.duplex kernel math memoize namespaces sequences ;
IN: benchmark.craschbench2

MEMO: test-bytes ( n -- byte-array ) iota >byte-array ;

TUPLE: tcp-echo < threaded-server ;

: <tcp-echo2> ( -- obj )
    binary \ tcp-echo new-threaded-server
    f 0 <inet4> >>insecure ;

: write-read ( bytes -- )
    [ write flush ] [ length read drop ] bi ;

M: tcp-echo handle-client*
    drop 10,000 [ 64 test-bytes write-read ] times ;

: server>address ( server -- port )
    servers>> first addr>> port>> local-server ;

: inner ( -- )
    <tcp-echo2> start-server server>address
    dup remote-address set binary <client> local-address set
    [ in>> input-stream set ]
    [ out>> output-stream set ] bi
    10,000 [ 64 read write flush ] times ;

: tcp-echo1-benchmark ( -- )
    "TEST START" print
    inner ;

MAIN: tcp-echo1-benchmark
```

If Factor is compiled in debug mode it crashes like once every third run. Maybe it has something with thread context switching to do?

mrjbq7:
Is this because of ``trace-stack-state2`` vs ``trace-stack-state``?

bjourne:
It's probably earlier than that. I'm checking out old versions and the bug happens at least since November last year.

bjourne:
It's the `VM_C_API void reset_context(factor_vm* parent, context* ctx)` function. It calls `init_context` which gc:s when there are important variables not protected by roots.

erg:
Fixed.

PR #1323, opened 2015-05-22
Jit bootstrap fixes

bjourne:
Here is a pr to fix the #1319 bug. `reset_context` can if you are unlucky gc. It allocates 48 bytes and the nursery is 2mb so the odds of it triggering a gc cycle is really low. But if it happens, the two top stack items must be preserved so they can be traced by the gc and the function previously didn't do that.


erg:
It seems to show the ui and then crash on Windows 64. Have you tested there?

erg:
I done goofed. I forgot to bootstrap with a new boot image.

Pulled. Closing.

PR #1325, opened 2015-05-25
Jit bootstrap fixes

bjourne:
Here is the pr again. The last commit is the one that fixes the bugs on 32bit platforms.


erg:
Looks like win32 is still crashing on bootstrap.

```
C:\factor32>factor.com \Users\Doug\config\mason-unix32.factor
Checking for multimedia extensions... SSE4.2 detected
Building with POPCNT support
Out of memory: VirtualAlloc

Nursery: Start=349c0000, size=100000, end=34ac0000
Aging: Start=347c0000, size=200000, end=349c0000
Tenured: Start=1c5c0000, size=18000000, end=345c0000
Cards:base=89b020, size=185400
```

bjourne:
Crap. Is it with new boot images too?

erg:
Yes, try it on win32 if you get a chance.

erg:
Ok, the new boot image isnt enough, the mason instance also has to be bootstrapped with it.

Everything works. Thanks!

PR #1334, opened 2015-06-03
Vm cleanup

bjourne:
Here is a small pr to make the cpp code cleaner. I made a macro `FACTOR_FOR_EACH` to make iteration nicer. I also added some functions to io.cpp so that if you run `factor -i=/tmp` it doesn't crash.

### DIFF
--- a/vm/image.cpp
+++ b/vm/image.cpp
@@ -18,7 +18,7 @@ void factor_vm::load_data_heap(FILE* file, image_header* h, vm_parameters* p) {
   init_data_heap(p->young_size, p->aging_size, p->tenured_size);
 
   fixnum bytes_read =
-      safe_fread((void*)data->tenured->start, 1, h->data_size, file);
+      raw_fread((void*)data->tenured->start, 1, h->data_size, file);
 
   if ((cell)bytes_read != h->data_size) {
     std::cout << "truncated image: " << bytes_read << " bytes read, ";
@@ -37,7 +37,7 @@ void factor_vm::load_code_heap(FILE* file, image_header* h, vm_parameters* p) {
 
   if (h->code_size != 0) {
     size_t bytes_read =
-        safe_fread((void*)code->allocator->start, 1, h->code_size, file);
+        raw_fread((void*)code->allocator->start, 1, h->code_size, file);
     if (bytes_read != h->code_size) {
       std::cout << "truncated image: " << bytes_read << " bytes read, ";
       std::cout << h->code_size << " bytes expected\n";
@@ -239,9 +239,8 @@ void factor_vm::load_image(vm_parameters* p) {
     free(msg);
     exit(1);
   }
-
   image_header h;
-  if (safe_fread(&h, sizeof(image_header), 1, file) != 1)
+  if (raw_fread(&h, sizeof(image_header), 1, file) != 1)
     fatal_error("Cannot read image header", 0);
 
   if (h.magic != image_magic)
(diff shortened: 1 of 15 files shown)


vm/utilities.hpp
```diff
@@ -1,5 +1,11 @@
 namespace factor {
 
+// Poor mans range-based for loops.
+#define FACTOR_FOR_EACH(iterable)                           \
+  for (typeof((iterable).begin()) iter = (iterable).begin();    \
```
mrjbq7: Does this call ``begin()`` twice?

vm/utilities.hpp
```diff
@@ -1,5 +1,11 @@
...
+       iter != (iterable).end();                                \
```
mrjbq7: And does this call ``end()`` on each loop?

vm/image.cpp
```diff
@@ -253,7 +252,7 @@ void factor_vm::load_image(vm_parameters* p) {
   load_data_heap(file, &h, p);
   load_code_heap(file, &h, p);
 
-  safe_fclose(file);
```
mrjbq7: Why change this one?  To hide ``io-error`` if close fails?

erg: ``safe_fclose`` is wrong anyway -- you're not allowed to retry on ``EINTR``
on ``flclose`` or ``close``.

On Wed, Jun 3, 2015 at 11:24 AM, John Benediktsson <notifications@github.com

vm/io.cpp
```diff
@@ -13,6 +13,39 @@ The Factor library provides platform-specific code for Unix and Windows
 with many more capabilities so these words are not usually used in
 normal operation. */
 
+int raw_fread(void* ptr, size_t size, size_t nitems, FILE* stream) {
+  size_t items_read = 0;
```
mrjbq7: Seems weird to return a ``size_t`` cast to an ``int``.

vm/io.cpp
```diff
@@ -13,6 +13,39 @@ The Factor library provides platform-specific code for Unix and Windows
...
+
+  do {
+    size_t ret = fread((void*)((int*)ptr + items_read * size), size,
+                       nitems - items_read, stream);
+    if (ret == 0) {
```
mrjbq7: The man pages say:

| If an error occurs, or the end-of-file is reached, the return value is a short object count (or zero).

So perhaps you need to check ``if (ret <= expected)``.

bjourne: Ok, I'll look into it closer. The idea was to return `-1` on error, therefore the size_t -> int change.

vm/io.cpp
mrjbq7: Maybe an ``ssize_t`` and overflow detection?  I dunno.

vm/image.cpp
bjourne: It's replaced with `raw_fclose`. Now if it fails, it will not segfault Factor.

bjourne:
Updated the pr. The `FACTOR_FOR_EACH` is fixed so now end() is only called once. I also changed the raw_fread/safe_fread combo back to returning a size_t. It returns 0 on error.

mrjbq7:
Looks pretty good to me, @erg?

erg:
Looks good.

On Thu, Jun 4, 2015 at 11:05 PM, John Benediktsson <notifications@github.com

mrjbq7:
Merged, thanks!!

PR #1348, opened 2015-06-09
Locals fixing

### LINKED ISSUE #1340: Locals off-by-one restart?
Somehow the restart for ``<bit-array>`` causes ``seq`` not to be found as a local.

```factor
:: foo ( n -- seq )
    n <bit-array> :> seq
    seq ;
```

bjourne:
I think this pr should fix #1340. For some reason the `locals` variable got out of sync with the words in the manifest after a restart. I didn't look at that so much because it was easier to just remove `locals` and use `manifest get qualified-vocabs>> last words>>` instead.

### DIFF
--- a/basis/locals/locals.factor
+++ b/basis/locals/locals.factor
@@ -6,8 +6,8 @@ locals.errors ;
 IN: locals
 
 SYNTAX: :>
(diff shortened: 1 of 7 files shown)


mrjbq7:
Merged, thanks!!

Issue #1189, opened 2014-11-07
json writer broken on some objects?


erg:
Should we allow ``>json`` to convert keys into strings if they are not already?

```factor
! Works
H{ { "1" 2 } { "3" 4 } } >json print
! {"1":2,"3":4}

! Fails in ``jsvar-encode`` because it treats keys as sequences
H{ { 1 2 } { 3 4 } } >json print
```

jonenst:
Here's a patch for writing escaped unicode characters > 0x10000
```diff
commit 96004a9ff495dfad400f3c51707959c4a37c89b0
Author: Jon Harper <jon.harper87@gmail.com>
Date:   Fri Jan 2 19:30:26 2015 +0100

    JSON: writer, allow escaped unicode > 0x10000

diff --git a/basis/json/writer/writer-tests.factor b/basis/json/writer/writer-tests.factor
index 8f46af1..4c3d563 100644
--- a/basis/json/writer/writer-tests.factor
+++ b/basis/json/writer/writer-tests.factor
@@ -67,3 +67,5 @@ TUPLE: person first-name age ;
     "\0\x01\x02\x03\x04\x05\x06\a\b\t\n\v\f\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\e\x1c\x1d\x1e\x1f"
     >json
 ] unit-test
+
+{ "\"\\ud834\\udd1e\"" } [ t json-escape-unicode? [ "𝄞" >json ] with-variable ] unit-test
diff --git a/basis/json/writer/writer.factor b/basis/json/writer/writer.factor
index ec65429..55aca17 100644
--- a/basis/json/writer/writer.factor
+++ b/basis/json/writer/writer.factor
@@ -2,7 +2,8 @@
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors ascii assocs combinators fry hashtables io
 io.streams.string json kernel locals math math.parser mirrors
-namespaces sequences strings tr words ;
+namespaces sequences strings tr words io.encodings.utf16.private
+formatting ;
 IN: json.writer
 
 SYMBOL: json-allow-fp-special?
@@ -39,6 +40,22 @@ M: t stream-json-print
 M: json-null stream-json-print
     [ drop "null" ] [ stream-write ] bi* ;
 
+: json-print-generic-escape-surrogate-pair ( stream char -- stream )
+    0x10000 - [ encode-first ] [ encode-second ] bi
[...]

PR #1354, opened 2015-06-10
Json see 1189

jonenst:
Hi,
this pr adds support for UTF16 escapes and fp-specials in json.reader and also adds an explicit error in json.writer when refusing to serialize fp-specials


erg:
Merged. Thanks!

What's left to do to json before we close #1189?

jonenst:
I think the only thing left is adding more unit tests

PR #1359, opened 2015-06-14
Adobe Brackets editor integration

kerabromsmu:


misc/brackets
```diff
@@ -0,0 +1 @@
+Subproject commit 963cbcb18da33c96b61009009f6fb11746f1ca11
```
mrjbq7: What is this file for?

basis/editors/brackets/brackets.factor
```diff
@@ -0,0 +1,33 @@
...
+SINGLETON: brackets-editor
+brackets-editor \ editor-class set-global
+
+! SYMBOL: brackets-path
+
+! HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\Brackets.exe
+
+<PRIVATE
+
+: windows-get-brackets-path ( -- path )
```
mrjbq7: Another way to do the ``os`` dispatch is:

```factor
HOOK: brackets-path os ( -- path )

M: windows brackets-path
    HKEY_LOCAL_MACHINE ...
```

Then in ``M\ brackets-editor editor-command`` you can just do:

```factor
M: brackets-editor editor-command
    [ brackets-path "brackets" or , drop , ] { } make ;
```

mrjbq7:
Awesome, thanks for the contribution!

I'd like to merge this, but need to know what the ``misc/brackets`` file is, and whether you'd like to make the ``brackets-path`` change I suggested (or I can make that change as part of adding Mac OS X support).

misc/brackets
kerabromsmu: This is adobe brackets mode for factor syntax highlight. It's a submodule - link to my project at https://github.com/kerabromsmu/brackets-factor

mrjbq7: Would you mind if we copied it directly rather than a git submodule?

kerabromsmu: I don't mind, but it probably won't be very convenient to maintain changes in both places. The syntax highlight is currently rather primitive, I hope to improve it in the future.

kerabromsmu:
Please feel free to do your suggested change. Anyway, I wasn't going to dig into Mac OS X support in the nearest future

mrjbq7:
Ok, I merged this and added support for Mac OS X.

Would you mind checking that it still works on Windows?

Regarding the ``brackets-editor`` repository you have:

1) Should I wait to merge it?

2) I see that it builds on the ``factor.js`` in CodeMirror. Should we have a ``misc/codemirror`` that has the ``factor.js`` file and a ``misc/brackets`` that incorporates it?  Wondering how to avoid duplication as improvements are made to the mode.

Thanks!

kerabromsmu:
It still works. Thanks!!

The story of factor.js is the following: as you've probably inferred, I use Brackets to edit code. A few weeks ago it didn't have a syntax highlight for Factor. So I dug into it and made a simple mode for Factor highlight for CodeMirror, which Brackets use for syntax highlight. It worked in CodeMirror, but when I made an extension for Brackets that included the same code, it didn't work. Why would I even use the same code twice? That's because the updated CodeMirror would have made its way into a new Brackets only after an indetermined time. And I didn't want to wait that long. So, it didn't work, and then I had to rewrite it to emulate the simple mode, which is even more code duplication. But that's what you pay for being impatient.

That said, I hope that when in the next version of Brackets they include both the updated CodeMirror and the support for simple mode, I can remove the duplicate code and only leave the factor.js that is now part of CodeMirror.

mrjbq7:
This was merged, if we want the syntax highlighting merged into ``misc/`` then we can do a separate pull request for that.  Thanks!

Issue #1345, opened 2015-06-08
gc-decode test failures


mrjbq7:
On Mac OS X x86-64.

```
==== resource:extra/tools/gc-decode/gc-decode-tests.factor

resource:extra/tools/gc-decode/gc-decode-tests.factor: 76

Unit Test: {
    { { } }
    [
        all-words [ normal? ] filter 50 sample [
            [ word>gc-info-expected ] [ word>gc-info ] bi
            same-gc-info?
        ] reject
    ]
}

=== Expected:
{ }
=== Got:
{ gml-determinant }
[Traceback]
```

bjourne:
Something maybe is wrong with the spill slots. `\ gml-determinant word>gc-info` says there are 13 of them. But `\ gml-determinant word>gc-info-expected` thinks there should be 14.

PR #1360, opened 2015-06-14
Fix 1345

bjourne:
Here is a branch I've worked on which should eliminate the #1345 bug and also add docs and refactor code. There were two problems, in ##phi nodes the vreg has more than one "parent" so which one was chooses as parent was random. I made it so the one with the lowest vreg nr always is choosen. The linear-scan step had the same problem so I changed the key from { start end } to { start end vreg } and because vreg nr is unique, it should make it so compilation becomes more deterministic.

### DIFF
--- a/basis/compiler/cfg/def-use/def-use-docs.factor
+++ b/basis/compiler/cfg/def-use/def-use-docs.factor
@@ -1,6 +1,10 @@
-USING: compiler.cfg compiler.cfg.instructions help.markup help.syntax
+USING: assocs compiler.cfg compiler.cfg.instructions help.markup help.syntax
 sequences ;
 IN: compiler.cfg.def-use
+
+HELP: compute-defs
+{ $values { "cfg" cfg } }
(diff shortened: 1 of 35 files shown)


mrjbq7:
Merged, thanks @bjourne !!

mrjbq7:
@bjourne any idea what this failure is?

```
USE: project-euler.044
```

results in this:

```
==== project-euler:

resource:extra/project-euler/project-euler.factor

 3: USING: definitions io io.files io.pathnames kernel math math.parser
15:     project-euler.041 project-euler.042 project-euler.043 project-euler.044
                                                                               ^
Generic word lookup-base-pointer* does not define a method for the POSTPONE: f class.
Dispatching on object: f


==== project-euler.044:

Generic word lookup-base-pointer* does not define a method for the POSTPONE: f class.
Dispatching on object: f
```

bjourne:
Strange. I must have missed that.

Issue #1329, opened 2015-05-31
load failure for compiler.cfg.gvn on mac32


erg:
```
Load failures

compiler.cfg.gvn
compiler.cfg.gvn.testing
Details:
==== compiler.cfg.gvn:

No suitable arithmetic method
left    f
right   0
generic -


==== compiler.cfg.gvn.testing:

resource:extra/compiler/cfg/gvn/testing/testing.factor

3: USING: accessors assocs compiler.cfg compiler.cfg.debugger
4: compiler.cfg.graphviz compiler.cfg.gvn
                                         ^
No suitable arithmetic method
left    f
right   0
generic -
```

mrjbq7:
I don't see this on mac64.

erg:
It doesn't even happen all the time on mac32.

bjourne:
Hasn't most of what was in `extra/compiler/cfg/gvn` been migrated/merged to `compiler/cfg/value-numbering/`? So maybe the test is triggering a failure in some old unused code that should be cleaned up?

mrjbq7:
The opposite, I think. The ``compiler.cfg.gvn`` is the new global value numbering that should replace the old one but is not enabled in master right now.

ajvondrak:
> is not enabled in master right now

Probably because it was being all [lame](https://github.com/slavapestov/factor/issues/671#issuecomment-8670229). Also, not pointing any fingers here or anything, but that @ajvondrak guy *did* kinda stop working on it... :sweat_smile:

erg:
Random failure that stopped happening, possibly fixed with the VM work.

Completely unrelated to gvn compiler pass and its author working or not working. :smiley_cat: :smile_cat: 

Closing for now.

PR #1361, opened 2015-06-15
Fix 1345 part2

bjourne:
I think I found the cause of the `project-euler.044` compilation problem. But the committs in this pr should fix it.

### DIFF
--- a/basis/compiler/cfg/linearization/linearization.factor
+++ b/basis/compiler/cfg/linearization/linearization.factor
@@ -74,5 +74,11 @@ SYMBOL: numbers
 : number-blocks ( bbs -- )
     H{ } zip-index-as numbers set ;
 
+: blocks>insns ( bbs -- insns )
+    [ instructions>> ] map concat ;
+
 : cfg>insns ( cfg -- insns )
-    linearization-order [ instructions>> ] map concat ;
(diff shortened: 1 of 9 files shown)


mrjbq7:
Merged, thanks!

mrjbq7:
I see this on linux-32 and mac-32:

```
=== benchmark.bootstrap1
vocab:bootstrap/assembler/x86.32.windows.factor

36: << "vocab:bootstrap/assembler/x86.32.factor" parse-file suffix! >> call
                                                                      ^
No suitable arithmetic method
left    f
right   0
generic -
```

bjourne:
Hm is it something random? The benchmarks works for me when I run them.

erg:
Looks like a duplicate of #1329.

Somewhere in the compiler is ``get``ing a variable that is f instead of a number?

mrjbq7:
Saw this on mac32:

```
==== math.transforms.bwt:

resource:extra/math/transforms/bwt/bwt.factor

3: USING: accessors assocs fry kernel locals sequences
4: sequences.rotated sorting ;
                    ^
No suitable arithmetic method
left    f
right   2
generic -


==== sequences.rotated:

No suitable arithmetic method
left    f
right   2
generic -
```

PR #1362, opened 2015-06-15
math.parser improvements ?

jonenst:
Hi,
This pr cleans up the syntax docs which showed "-5/-6" which is not supported anymore (probably since 2009), so I think it's safe to remove it from the docs.

Also, This pr adds support for "1+1/3." to produce the float 1.33333333333 ( we already had support for "1/3." which produces 0.3333333333). Also add in the docs examples like 1/2.5 or 1+1/2.5 for 0.4 and 1.4

I was thinking of more changes but wanted to discuss first:
- string>number throws an exception for 1/0, but returns f for other invalid strings. Should it also return f for 1/0? Or just add a note in the doc that this will throw an exception?
- Should the following be parsed as numbers ?
  - "1+1"
  - "1+1."
  - "1+1/1"
  - "1+1/0."
  - "1+1/."
  - "1+2/1" 
  - "1+2/2"
  - "1+5/2"

My first thought was to only allow the part after "+" to be >0 and <1, but for compatibility with the previous check ("ratio?"), I did "dup >integer number= not"
cheers,
Jon

### DIFF
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -83,7 +83,7 @@ M: real pprint*
 
 M: float pprint*
     dup fp-nan? [
-        \ NAN: [ fp-nan-payload >hex text ] pprint-prefix
+        \ NAN: [ double>bits >hex text ] pprint-prefix
     ] [
         number-base get {
             { 10 [ number>string text ] }
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -81,8 +81,11 @@ M: real pprint*
         [ unsupported-number-base ]
     } case ;
 
+: payload-nan? ( f -- ? )
+    dup fp-nan? [ fp-nan-payload 0x8000000000000 = not ] [ drop f ] if ;
+
 M: float pprint*
-    dup fp-nan? [
+    dup payload-nan? [
         \ NAN: [ double>bits >hex text ] pprint-prefix
     ] [
         number-base get {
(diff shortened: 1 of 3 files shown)


bjourne:
I like `1+1/3.` because it's logical with `1/3.`. I don't like the other rational syntaxes because `1+1` doesn't seem like something someone would type and expect to evaluate to 2.

jonenst:
Also
"1/1e2" string>number ! 0.01
"1+1/1e2" string>number ! f

jonenst:
Also
```1+1/1e10 ! 1.0000000001```
```1+1/1e100 ! 1.0 because 1/1e100 == 1e-100 != 0 (the check for ?add-ratio)```
```1+1/1e1000 ! f because 1/1e1000 == 0.0 (the check for ?add-ratio)```

jonenst:
And what about things like ```0xc+b/a``` or ```0o4+7/3``` or ```0b1101+1101/110``` which work (not affected by this patch) ? I think it's important that every undocumented thing should be parsed as f.

jonenst:
So I'm looking at this again. Should we support
  * 0xc+a/b ? I think not, leave hex oct and bin for integers and floats, not ratios.
  * 1+4/3 ? I think not. I'm not used to this number representation (I think it's used in the US?) but wikipedia says ( https://en.wikipedia.org/wiki/Fraction_(mathematics) ) that it "mixed numbers" are the sum of a non-zero integer and a proper fraction (a proper fraction : absolute value of the fraction is strictly less than one)
  * 1/2.5 ? I think not, only allow this as the float value of a ratio by suffixing , e.g 1/2.0
  * 1+1/2.5 ? same as previous
  * 1+1/2e5 ? same as previous

mrjbq7:
We support 4/3 right now, which is nice for example every 2.5 seconds could be ``60/2.5`` rather than a weird looking constant (maybe alternatively could use literals ``$[ 60 2.5 / ]``).

I guess I don't see why not support ``1+4/3`` or ``1/2.5``, unless we wanted our math parser to be strict simplified representation only?

mrjbq7:
Thanks @jonenst !

A day late and a dollar short but this is great !

PR #1363, opened 2015-06-16
prettyprinter, locale independant float printing

jonenst:
Hi,
this pr fixes the float printing https://github.com/slavapestov/factor/issues/1044
it changes the following:
- add a new primitive  (so need to bootstrap) format-float-C that always uses the C locale by using c++ stringstreams. The formatting is passed by parameters instead of a format string. use the new primitive in math.parser and calendar.format
- remove "fix-float" from the previous format-float because it would specifically look for the "." separator which doesn't work for some locales. the formatting vocabulary still uses format-float so is correctly locale dependant, but the output my differ a bit from before since we don't fix-float, so 2.0 can print as 2 depending on the format string.

What do you think ?

### DIFF
--- a/vm/math.cpp
+++ b/vm/math.cpp
@@ -1,6 +1,7 @@
 #include "master.hpp"
 #include <sstream>
 #include <iomanip>
+#include <stdexcept>
 
 namespace factor {
 
@@ -220,7 +221,14 @@ void factor_vm::primitive_format_float() {
   char* fill = alien_offset(ctx->pop());
   double value = untag_float_check(ctx->peek());
   std::ostringstream localized_stream;
-  localized_stream.imbue(std::locale(locale));
+  try {
+    localized_stream.imbue(std::locale(locale));
+  } catch (const runtime_error& error) {
+    byte_array* array = allot_byte_array(1);
+    array->data<char>()[0] = '\0';
+    ctx->replace(tag<byte_array>(array));
+    return;
+  }
   switch (format[0]) {
     case 'f': localized_stream << std::fixed; break;
     case 'e': localized_stream << std::scientific; break;
--- a/vm/math.cpp
+++ b/vm/math.cpp
@@ -1,4 +1,6 @@
 #include "master.hpp"
+#include <sstream>
+#include <iomanip>
 
 namespace factor {
 
@@ -211,10 +213,36 @@ void factor_vm::primitive_fixnum_to_float() {
 
 /* Allocates memory */
 void factor_vm::primitive_format_float() {
-  byte_array* array = allot_byte_array(100);
+  char* locale = alien_offset(ctx->pop());
   char* format = alien_offset(ctx->pop());
+  fixnum precision = untag_fixnum(ctx->pop());
+  fixnum width = untag_fixnum(ctx->pop());
+  char* fill = alien_offset(ctx->pop());
   double value = untag_float_check(ctx->peek());
-  SNPRINTF(array->data<char>(), 99, format, value);
+  std::ostringstream localized_stream;
+  localized_stream.imbue(std::locale(locale));
+  switch (format[0]) {
+    case 'f': localized_stream << std::fixed; break;
+    case 'e': localized_stream << std::scientific; break;
+  }
+  if (isupper(format[0])) {
+    localized_stream << std::uppercase;
+  }
+  if (fill[0] != '\0') {
+    localized_stream << std::setfill(fill[0]);
+  }
+  if (width >= 0) {
+    localized_stream << std::setw(width);
+  }
+  if (precision >= 0) {
+    localized_stream << std::setprecision(precision);
+  }
+  localized_stream << value;
+  const std::string& tmp = localized_stream.str();
+  const char* cstr = tmp.c_str();
+  int size = tmp.length()+1;
+  byte_array* array = allot_byte_array(size);
+  memcpy(array->data<char>(), cstr, size);
   ctx->replace(tag<byte_array>(array));
 }
 
--- a/basis/calendar/format/format.factor
+++ b/basis/calendar/format/format.factor
@@ -150,7 +150,7 @@ M: timestamp year. ( timestamp -- )
 ! Should be enough for anyone, allows to not do a fancy
 ! algorithm to detect infinite decimals (e.g 1/3)
 : ss.SSSSSS ( timestamp -- )
-    second>> >float "%09.6f" format-float write ;
+    second>> >float "0" 9 6 "f" "C" format-float write ;
 
 : (timestamp>rfc3339) ( timestamp -- )
     {
--- a/basis/formatting/formatting-docs.factor
+++ b/basis/formatting/formatting-docs.factor
@@ -139,7 +139,7 @@ HELP: strftime
 } ;
 
 ARTICLE: "formatting" "Formatted printing"
-"The " { $vocab-link "formatting" } " vocabulary is used for formatted printing."
+"The " { $vocab-link "formatting" } " vocabulary is used for english formatted printing."
 { $subsections
     printf
     sprintf
--- a/basis/formatting/formatting.factor
+++ b/basis/formatting/formatting.factor
@@ -30,8 +30,7 @@ IN: formatting
     [ 0 ] [ string>number ] if-empty ;
 
 : format-simple ( x digits string -- string )
-    [ [ >float ] [ number>string ] bi* "%." ] dip
-    surround format-float ;
+    [ >float "" -1 ] 2dip "" format-float ;
 
 : format-scientific ( x digits -- string ) "e" format-simple ;
 
--- a/basis/formatting/formatting.factor
+++ b/basis/formatting/formatting.factor
@@ -30,7 +30,7 @@ IN: formatting
     [ 0 ] [ string>number ] if-empty ;
 
 : format-simple ( x digits string -- string )
-    [ >float "" -1 ] 2dip "" format-float ;
+    [ >float "" -1 ] 2dip "C" format-float ;
 
 : format-scientific ( x digits -- string ) "e" format-simple ;
 
--- a/basis/stack-checker/known-words/known-words.factor
+++ b/basis/stack-checker/known-words/known-words.factor
@@ -420,7 +420,7 @@ M: object infer-call* \ call bad-macro-input ;
     { fixnum>float { fixnum } { float } }
 
     ! float
-    { (format-float) { float byte-array } { byte-array } }
+    { (format-float) { float byte-array fixnum fixnum byte-array byte-array } { byte-array } }
     { bits>float { integer } { float } }
     { float* { float float } { float } }
     { float+ { float float } { float } }
--- a/core/bootstrap/primitives.factor
+++ b/core/bootstrap/primitives.factor
@@ -476,7 +476,7 @@ tuple
     { "bits>float" "math" "primitive_bits_float" ( n -- x ) }
     { "double>bits" "math" "primitive_double_bits" ( x -- n ) }
     { "float>bits" "math" "primitive_float_bits" ( x -- n ) }
-    { "(format-float)" "math.parser.private" "primitive_format_float" ( n format -- byte-array ) }
+    { "(format-float)" "math.parser.private" "primitive_format_float" ( n fill width precision format locale -- byte-array ) }
     { "bignum*" "math.private" "primitive_bignum_multiply" ( x y -- z ) }
     { "bignum+" "math.private" "primitive_bignum_add" ( x y -- z ) }
     { "bignum-" "math.private" "primitive_bignum_subtract" ( x y -- z ) }
--- a/core/math/parser/parser.factor
+++ b/core/math/parser/parser.factor
@@ -6,7 +6,7 @@ sequences.private splitting strings strings.private ;
 IN: math.parser
 
 <PRIVATE
-PRIMITIVE: (format-float) ( n format -- byte-array )
+PRIMITIVE: (format-float) ( n fill width precision format locale -- byte-array )
 PRIVATE>
 
 : digit> ( ch -- n )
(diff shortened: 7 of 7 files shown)


bjourne:
Maybe you can extend `(format-float-C)` so that it can fully replace `(format-float)`? Because it's problematic to have two similar but slightly different float formatting primitives. Like change the stack effect of the new primitive to `( n fill width precision format locale -- byte-array )` and pass `"C"` to format with the C locale or `""` to format with the users locale. Then the remaining places where `(format-float)` is used could easily be replaced with that new primitive. The full power of snprintf-formatting isn't used anywhere so the flags you have implemented in `primitive_format_float_C` should be enough.

Otherwise I like this pr. But that's just my opinion!

vm/math.cpp
```diff
@@ -217,6 +219,40 @@ void factor_vm::primitive_format_float() {
...
+  double value = untag_float_check(ctx->peek());
+  std::ostringstream locale_C_stream;
+  locale_C_stream.imbue(std::locale("C"));
+  switch (format[0]) {
+    case 'F': locale_C_stream << std::uppercase;
+    case 'f': locale_C_stream << std::fixed; break;
+    case 'E': locale_C_stream << std::uppercase;
+    case 'e': locale_C_stream << std::scientific; break;
+    case 'G': locale_C_stream << std::uppercase;
+  }
```
bjourne: I think `if (isupper(format[0])) { locale_C_stream << std::uppercase; }` is clearer here than using the break-fallthrough trick.

jonenst:
I reworked this PR:
  * primtive format float which takes the locale
  * merge with new PRIMITIVE: changes
  * use uppercase instead of switch as suggested by bjourne
  * look for "e" or "E" in fix-float depending on case

The tests for math.parser always pass. The test for formatting pass if and only if LC_NUMERIC is a locale that uses "." as a decimal mark.

One thing that can be changed is that currently a C++ exception is thrown if a bad locale name is passed in, which results in an exception in factor:
```factor
1.0 "" -1 -1 "f" "foo" format-float
```
=> ```Object:  { "kernel-error" 5 6 f } Summary: VM``` SIGABORT and some stuff is printed on stderr:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  locale::facet::_S_create_c_locale name not valid
```

maybe try/catch in primitive_format_float and return f (or "" for backward compatibility) ? Also, I'm not sure if ostringstream can generate ios::badbit or ios::failbit or other errors during the formatting, and I'm not sure how to handle them.

Note that the previous implementation had the following behavior (because the return value of printf wasn't checked):
```factor
IN: scratchpad 1.0 "ko" format-float .
"ko.0"
```

jonenst:
Also, this brings the question of what "formatting" should do regarding locales. @mrjbq7 ? Is it really intended to use the user locale ? Things like: should ```1.1 "%f" printf```` print "1,1" ?

mrjbq7:
@jonenst I never gave locale much thought.  Having some kind of with-locale word that changes the output is probably a good idea.  Not sure about parsing and whether with-locale should change the parsing routines to handle 1,1...

jonenst:
I guess the most important thing is to have very explicit documentation on which words are affected by ```with-locale```. This way you can write code and know that a caller won't break it with a locale setting. If not, you end up having to use ```"C" with-locale``` everywhere defensively.

So maybe group all locale dependent words in "formatting" ? And it would be nice to have locale dependant parsing routines like scanf in C.

bjourne:
Wrt the exception, I don't think it matters that `primitive_format_float` throws if you pass in an invalid locale name because the word is in a private area and you aren't supposed to call it directly. 

Some tests break because `12.34 "%.2f" sprintf` is printed as `12,34` instead of `12.34`. My suggestion to pass the locale name as a parameter wasn't so good because it's always the C-locale you want to use. But you can just change the call in `format-simple` to `"C" format-float` and then all formatting would be using the C locale. So this looks really good to me.

jonenst:
@mrjbq7  Do you agree to document the "formatting" vocab as locale independant (ie english only) for now? Changing this shouldn't be a backwards compatibility problem since the previous behavior for floats was broken ("1,5.0"), and some features were already english only (%a day of week, etc). I think it's a good solution for now.

Also I implemented the exception catching in C++ and made it return an empty string instead because depending on factor's correct handling of c++ exceptions adds complexity (even if the word is private now, it will probably be used in the future for locale dependant vocabs, so let's make it robust now)

So later, another vocab can be created for locale dependant parsing/formatting (and/or "formatting" updated, but taking backwards compatibility into account..). For example, Python has the "n" format specifier for locale dependant numbers (https://docs.python.org/3/library/string.html and https://pypi.python.org/pypi/parse) and also locale.str, locale.atof, locale.atoi (https://docs.python.org/3.2/library/locale.html). Passing the locale to (format-float) will be useful for this so I think we should keep it. 

Finally, a recap note: All this started when we starting using GTK as a ui backend (GTK is the one calling setlocale to not use the default "C" locale and use the environnement variables instead). The X11 backend doesn't call setlocale. The cocoa and windows backends also problably don't. So the easy solution would have been to simply call gtk_disable_setlocale (https://developer.gnome.org/gtk3/stable/gtk3-General.html#gtk-disable-setlocale). Now that the whole codebase is locale independant, the result is the same...until we add some support for locale dependant programs. At this point, will all the ui backends automatically call setlocale (or just gtk, or none (with gtk_disable_setlocale)) (and what about console apps) ? I think the most consistent is none and the programmer has to call it.. but that's for later :)

mrjbq7:
@jonenst Sorry I didn't reply sooner -- I was traveling for a few days without a laptop or cellphone reception.  I love your attention to detail here - thanks!

Some comments:

1) Its a good idea to document ``formatting`` being English only (at least until it is ever changed to be localized).

2) It seems like the ``gtk`` backend shouldn't behavior different from the other backends.

3) I'm not sure about ``std::ostringstream`` vs ``sprintf``, but seems okay?

4) Is it a good idea to print ``2.0`` as ``2``?

5) Can ``calendar.format`` use ``formatting`` so ``format-float`` is only used in a couple places?

jonenst:
>@jonenst Sorry I didn't reply sooner -- I was traveling for a few days without a laptop or cellphone reception. I love your attention to detail here - thanks!

Thanks, no problem

> 2) It seems like the gtk backend shouldn't behavior different from the other backends.

Yeah I agree. According to https://mail.gnome.org/archives/gtk-devel-list/2001-October/msg00082.html, in gtk you can do something like
```C
    gtk_disable_setlocale();

    setlocale (LC_ALL, "");
    setlocale (LC_NUMERIC, "C");
 
    gtk_init();
```
So let's just call gtk_disable_setlocale by default and people who want localized programs can call libc's setlocale to configure the locale exactly like they want it before starting the ui ? (we'll have to add support for setlocale in libc.factor, with platform dependant LC_* constants according to http://sourceforge.net/p/factor/mailman/message/29294415/); Hmm according to this email the x11 backend does calls Xlib's setlocale, but on ubuntu 14.04 this didn't lead to numbers printed as "1,5.0" whereas with the gtk backend it did. Need to investigate...

> 3) I'm not sure about std::ostringstream vs sprintf, but seems okay?

I think that if it works ok on all platforms, we can't pass on being independant of the locale for core float pretty printing

> 4) Is it a good idea to print 2.0 as 2?

For factor code, no for homoiconicity, but for factor code we still print 2.0; for user apps, both are valid I guess. Printing "2" is glibc's printf the default behavior of "%g", but  "%#g" does does not remove trailing zeros.

> 5) Can calendar.format use formatting so format-float is only used in a couple places?

It can, but then formatting must be english. I think it's ok if we plan to go for "%n" for localized numbers like in python

jonenst:
> according to http://sourceforge.net/p/factor/mailman/message/29294415/); Hmm according to this email the x11 backend does calls Xlib's setlocale, but on ubuntu 14.04 this didn't lead to numbers printed as "1,5.0" whereas with the gtk backend it did. Need to investigate...

So this section of the X11 code comes from this email thread:
http://sourceforge.net/p/factor/mailman/message/18485632/

Setting the locale was to fix an input problem for a japanese debian user usin uim-xkk. I'll try to reproduce it.

On Ubuntu 14.04, I don't have X_LOCALE defined (see a reference here https://lists.debian.org/debian-devel/1998/09/msg00550.html), so my constants are the libc's constants, not the xlocale constants. In particular, LC_ALL is 6, not 0 as defined in basis/x11/xlib/xlib.factor. If I change LC_ALL to 6, then the "1,5.0" bug appears.

jonenst:
I tested with ibus-anthy on ubuntu 14.04:
- using the gtk backend, it is possible to input japanese using ibus-anthy. It's not ideal: the characters appear at the bottom of the window in an overlay window instead of where the cursor is.
- using the x11 backend, with or without the change to the value of LC_ALL (0 in X_LOCALE mode, 6 in libc mode), it is not possible to input japanese using ibus-anthy. The characters appear at the bottom of the window like with gtk, but when pressing enter, either nothing is input near the cursor, or unprintable characters.

jonenst:
So xopenIm uses LC_CTYPE internally. In X_LOCALE mode, 0 is LC_ALL; in libc mode, on my system, 0 is LC_CTYPE and 6 is LC_ALL; So right now factor thinks it sets LC_ALL, but in reality sets LC_CTYPE, which is what is actually used by xopenim anyway. So both 0 and 6 values "work" as an input to setlocale. Also, I tested using "C" setlocale, and the "Xutf8LookupString" method returned the empty string when entering japanase, so the setlocale call is really needed. 

Also, factor uses "XwcLookupString" to retrieve the string and decodes it using utf16, but this function is documented as using the locale enconding (I don't know what it was on my system, but not UTF-16). So the resulting string contained non alpha characters and was filtered by the code in ```valid-input?```. By changing it to Xutf8LookupString and decoding as utf-8, I was able to enter japanase with the x11 ui backend. Here's a proof of concept patch:

```diff
diff --git a/basis/x11/xim/xim.factor b/basis/x11/xim/xim.factor
index 5fd42fd..3d79244 100644
--- a/basis/x11/xim/xim.factor
+++ b/basis/x11/xim/xim.factor
@@ -41,17 +41,17 @@ SYMBOL: keybuf
 SYMBOL: keysym
 
 : prepare-lookup ( -- )
-    buf-size uint <c-array> keybuf set
+    buf-size char <c-array> keybuf set
     0 KeySym <ref> keysym set ;
 
 : finish-lookup ( len -- string keysym )
-    keybuf get swap 2 * head utf16n decode
+    keybuf get swap head utf8 decode
     keysym get *KeySym ;
 
 : lookup-string ( event xic -- string keysym )
     [
         prepare-lookup
         swap keybuf get buf-size keysym get 0 int <ref>
-        XwcLookupString
+        Xutf8LookupString
         finish-lookup
     ] with-scope ;
diff --git a/basis/x11/xlib/xlib.factor b/basis/x11/xlib/xlib.factor
index 3e8533c..c9cfaff 100644
--- a/basis/x11/xlib/xlib.factor
+++ b/basis/x11/xlib/xlib.factor
@@ -1427,7 +1427,8 @@ X-FUNCTION: void XUnsetICFocus ( XIC ic ) ;
 
 X-FUNCTION: int XwcLookupString ( XIC ic, XKeyPressedEvent* event, ulong* buffer_return, int bytes_buffer, KeySym* keysym_return, Status* status_return ) ;
 
-X-FUNCTION: int Xutf8LookupString ( XIC ic, XKeyPressedEvent* event, c-string buffer_return, int bytes_buffer, KeySym* keysym_return, Status* status_return ) ;
+X-FUNCTION: int Xutf8LookupString ( XIC ic, XKeyPressedEvent*
+event, char* buffer_return, int bytes_buffer, KeySym* keysym_return, Status* status_return ) ;
 
 ! !!! category of setlocale
 CONSTANT: LC_ALL      0
```

jonenst:
So, in retrospect, here's what I understand we should do:
  - we must call setlocale(X,"") at least with LC_CTYPE to allow to input characters in some locales (that's how the X server works). I guess it's better with LC_ALL (gtk recommends it) to respect the user choice of locale. And so we should call it on all backends.
  - we must call the libc setlocale with libc's constants, instead of xlib's (apparently it was removed entirely because it was not used anymore: http://lists.x.org/archives/xorg-devel/2013-November/thread.html#39051): remove the defintions from basis/x11/xlib/xlib.factor and add them to basis/libc/libc.factor (or basis/libc/{windows,linux,macosx} if the constants are different)
 - use my patch to use xutf8lookupstring instead of xwclookupstring to have the correct encoding instead of a locale dependant one (defined in the file "locale.alias") to fix X11 backend input.
- use C++ locale independant streams for factor's core functions to be independant of the locale

Is that ok ?


Also, for reference, here are the values for libc's setlocale on ubuntu 14.04:
```
#define __LC_CTYPE		 0
#define __LC_NUMERIC		 1
#define __LC_TIME		 2
#define __LC_COLLATE		 3
#define __LC_MONETARY		 4
#define __LC_MESSAGES		 5
#define __LC_ALL		 6
#define __LC_PAPER		 7
#define __LC_NAME		 8
#define __LC_ADDRESS		 9
#define __LC_TELEPHONE		10
#define __LC_MEASUREMENT	11
#define __LC_IDENTIFICATION	12
```

mrjbq7:
That seems like a pretty good approach.

jonenst:
I plan to work on this to finish it, still good with the approach of making the core number printing locale independant by using c++ streams ?

mrjbq7:
I can't think of any real downsides, can you?

There are a few math.parser pull requests / issues and I'm not sure what's mergeable now (sorry!). If you have some commits to review/merge maybe open a new PR?

jonenst:
The main thing I think is that it adds more dependencies to the C++ stdlib because we're now depending on c++ streams to print numbers. But I guess it's ok. Can someone test it on windows at least once ? :) I'll update this PR to make it mergeable.

About the other PRs, i'll check them and report

jonenst:
Rebased on master. So the procedure is to have it work is:
- have a working factor, move your boot-image from the factor directory
- checkout this branch
- start factor, call ```make-boot-image```
- make clean; make
- ./factor -i=bootimage

jonenst:
Reapplied your ideas from 19b1dc1d055a and made it inline

bjourne:
Isn't this pr ready to be applied? What is going on? :)

jonenst:
There's a tiny conflict with a915d3bdb4eaaa12a85073a7e9abb1fe700d689b in basis/stack-checker/known-words/known-words.factor, the merge is obvious. I'm bootstrapping to see if everything works correctly.

This PR can go as is, it will fix the "1,5.0" bug.

jonenst:
Rebased on current master

jonenst:
So.. I think this should be merged in :)

jonenst:
Is there something to do on this PR ?

bpollack:
I'm on board, to the extent it matters; this seems pretty clean and quite useful.

mrjbq7:
Merged, thanks!!

PR #1364, opened 2015-06-16
Numbers2

jonenst:
More changes to math.parser. We can cherry-pick the commits we want.
- 78d9c873516293d534cc9291b008dfcdb84eee88 removes the 0 division exception and outputs f instead for consistency. Not sure if that's good or not
- 6a95b098eb14b6fa9b09fa83017de024b6271567 allows to see the sign of NaNs by printing the full 64 bits instead of only the 52 lsb. The added msb are either "7ff" (positive nan) or "fff" (negative nan). Also before this patch, "NAN: 1" printed as "NAN: 1", but now we will always see the much longer form "NAN: 7ff0000000000001".
- 78d9c873516293d534cc9291b008dfcdb84eee88 tries to make some NaNs nicer by printing "0/0." or "-0/0." when the payload 50 lsb bits are 0. These are the NaNs returned by most standard functions or cpu instructions (for example the sse DIVSD instruction) , but the sign is totally unreliable:
  - "0/0." and "-1 fsqrt" return "-0/0" (hex 0xfff8000000000000)
  - -1 flog returns "0/0." (hex 0x7ff8000000000000)
    So I think 78d9c873516293d534cc9291b008dfcdb84eee88 can't be merged yet because "0/0." outputting "-0/0." breaks homoiconicity. Maybe we can just special case "0/0." to be "NAN: 7ff8000000000000"; Or we can ignore the sign when printing "0/0.". I think I prefer the first solution to avoid prettyprintg "NAN: 7ff8000000000000" and "NAN: fff8000000000000" the same way.

What do you think ?

### DIFF
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -82,7 +82,7 @@ M: real pprint*
 
 M: float pprint*
     dup fp-nan? [
-        \ NAN: [ fp-nan-payload >hex text ] pprint-prefix
+        \ NAN: [ double>bits >hex text ] pprint-prefix
     ] [
         number-base get {
             { 16 [ [ >hex ] "0x" pprint-prefixed-number ] }
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -82,7 +82,11 @@ M: real pprint*
 
 M: float pprint*
     dup fp-nan? [
-        \ NAN: [ double>bits >hex text ] pprint-prefix
+        dup fp-nan-payload 50 2^ 1 - bitand 0 = [
+           fp-sign "-0/0." "0/0." ? text
+        ] [
+           \ NAN: [ double>bits >hex text ] pprint-prefix
+        ] if
     ] [
         number-base get {
             { 16 [ [ >hex ] "0x" pprint-prefixed-number ] }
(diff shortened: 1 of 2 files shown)


jonenst:
Also I'll add docs/tests when we decide if we want any of these commits

bjourne:
Those are all small but good changes IMHO. Do they pass all existing unit tests?

jonenst:
Haven't done a load-all test-all.
After thinking about it, I'm not so sure about another thing: I chose to display both signaling and quiet NaNs as "[-]0/0." if the payload is 0 (ie NAN: 7ff8000000000000, NAN: 7ff4000000000000 as "0/0.",  NAN: fff8000000000000, NAN: fff4000000000000 as "0/0.") , which is maybe not so good.
I think signaling NANs are rare, so maybe we should display them as NAN: fff4000000000000

Also, I need to add a special case for 0/0. to output NAN: 7ff8000000000000 (positive quiet nan with payload 0) instead of NAN: 7ff8000000000000 (negative quiet nan with payload 0)

mrjbq7:
@jonenst I made this commit (7b0733f72f8cb264e3c82bd63fccd1aa1099d3b4) to fix a docs example -- look legit?

jonenst:
Yup legit :)
Le 1 août 2015 5:00 PM, "John Benediktsson" <notifications@github.com> a
écrit :

mrjbq7:
I merged the zero division commit, but I don't know if ``0/0.`` and ``-0/0.`` are "universal" enough to have constant values?  Also, right now ``-0/0.`` parses as ``0.0``, I think.

PR #1365, opened 2015-06-16
Tab change focus

jonenst:
Found this branch in my repo, not sure if I already submitted this. It looks good to me, it allows to change the focus with tab and shift tab, for example on :
```factor
<pile>
  "coucou" <model> <model-field> add-gadget
  "cocou" <model> <model-field> add-gadget
"coucou" open-window
```


jonenst:
Well, now that I look at it again, the code is not correctly indented, there are no docs and tests, and no way to enter a tab character now that it changes the focus. I'll redo it and submit again :)

PR #1366, opened 2015-06-16
Fix 1353

### LINKED ISSUE #1353: compiler.cfg.linear-scan.allocation register test failures
I changed the ``<word>`` hashcode algorithm to just combine the name and vocab without using a zero, but that resulted in a slightly different hashcode.  These were the test failures in ``compiler.cfg.linear-scan.allocation`` that resulted:

```
==== resource:basis/compiler/cfg/linear-scan/allocation/allocation-tests.factor

resource:basis/compiler/cfg/linear-scan/allocation/allocation-tests.factor: 26

Unit Test: {
    { RCX }
    [
        {
            {
                int-regs
                V{
                    T{ live-interval-state
                        { vreg 49 }
                        { reg R8 }
                        { start 30 }
                        { end 46 }
                        { ranges { ~live-range~ } }
                        { uses { ~vreg-use~ ~vreg-use~ } }
                        { reg-class int-regs }
                    }
                }
            }
            { float-regs V{ } }
        } active-intervals set
        unassigned-interval dup machine-registers
        assign-register reg>>
    ]
}

=== Expected:
RCX
=== Got:
R8
[Traceback]

resource:basis/compiler/cfg/linear-scan/allocation/allocation-tests.factor: 32

Unit Test: {
    { { RCX 1/0. } }
    [
        { { int-regs V{ } } { float-regs V{ } } }
        active-intervals set
        unassigned-interval machine-registers register-status
    ]
}

=== Expected:
{ RCX 1/0. }
=== Got:
{ R8 1/0. }
[Traceback]
```

bjourne:
This pr should fix #1353. It makes the order in which registers are allocated more predictable.

### DIFF
--- a/basis/compiler/cfg/linear-scan/allocation/allocation-docs.factor
+++ b/basis/compiler/cfg/linear-scan/allocation/allocation-docs.factor
@@ -1,11 +1,16 @@
 USING: assocs compiler.cfg compiler.cfg.instructions
 compiler.cfg.linear-scan.allocation compiler.cfg.linear-scan.allocation.state
-compiler.cfg.linear-scan.live-intervals help.markup help.syntax kernel sequences ;
+compiler.cfg.linear-scan.live-intervals hashtables help.markup help.syntax
+kernel sequences ;
 IN: compiler.cfg.linear-scan.allocation
 
 HELP: (allocate-registers)
 { $values { "unhandled-min-heap" "stuff" } }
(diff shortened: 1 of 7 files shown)


mrjbq7:
What's the logic for changing register order?

mrjbq7:
Merged!

bjourne:
Aesthethics. :) On 64 bit most registers are general purpose so any order is fine. But oldschool assembly programming conventions says that the preferred register use order is RAX, RBX, RCX and so on.

Issue #1312, opened 2015-05-08
Wrong behaviour of :> - Cannot execute a word before it has been compiled


nedzadarek:
```factor
USE: locals
: qux ( -- x ) 42 ;
: foo ( -- x ) qux ;
qux ! 42
foo ! 42

: bar ( -- x ) 
[let 10 :> qux qux ] ;
bar ! 10
qux ! Error: Cannot execute a word before it has been compiled
foo ! 42 - works
```
Acording to documentation of `:>`, `qux` should work.

bjourne:
Is that because you name your local variable qux and there is already a word with that name?

nedzadarek:
Yes.

mrjbq7:
Normally shadowed words require disambiguation (e.g., ``FROM: assocs => change-at ;``).

Thats not a very nice error message we generate, and probably no "fix" besides renaming unless we make it restartable, or always allow shadowing.

bjourne:
Is there a use case for shadowing like nedzadareks example?

mrjbq7:
I've run into this when I have some "common-sounding" words in a vocab and mix with locals, but only occasionally.

mrjbq7:
@bjourne fixed this, closing.

PR #1369, opened 2015-06-22
Fix 1338

### LINKED ISSUE #1338: Funky "has not been compiled" problem
Start a fresh Factor instance, and paste this into the UI listener.

```factor
:: sieve ( n -- primes )
    n 1 + t <array> :> sieve
    f 0 sieve set-nth
    f 1 sieve set-nth

    2 n [a,b] [| i |
        i sieve nth [
            2 i * n i <range> [| j |
                f j sieve set-nth
            ] each
        ] when
    ] each

    sieve [ drop ] selector [ each-index ] dip ;
```

You will get two restarts: one for ``locals``, one for ``math.ranges``.

After choosing the "right" vocabs to use, try running ``100 sieve``.

You get this error:

```
Cannot execute a word before it has been compiled
word sieve
```

bjourne:
This pr should fix #1338 and probably also #1312. I just added `manifest [ clone [ clone ] change-qualified-vocabs ] change` and changes the name to `with-lambda-scope` to communicate the intent better.

### DIFF
--- a/basis/functors/functors.factor
+++ b/basis/functors/functors.factor
@@ -141,12 +141,6 @@ DEFER: ;FUNCTOR delimiter
 
 <PRIVATE
 
-: push-functor-words ( -- )
-    functor-words use-words ;
-
-: pop-functor-words ( -- )
-    functor-words unuse-words ;
-
 : (parse-bindings) ( end -- words )
     [ dup parse-binding dup ]
     [ first2 [ make-local ] dip 2array ]
@@ -159,16 +153,14 @@ DEFER: ;FUNCTOR delimiter
     [
         building get use-words
         (parse-bindings)
-        building get unuse-words
     ] with-bindings ;
 
 : parse-functor-body ( -- form )
-    push-functor-words
+    functor-words use-words
     "WHERE" parse-bindings
     [ [ swap <def> suffix ] { } assoc>map concat ]
-    [ [ \ ;FUNCTOR parse-until >quotation ] ((parse-lambda)) ] bi*
-    [ ] append-as
-    pop-functor-words ;
+    [ [ \ ;FUNCTOR parse-until >quotation ] with-lambda-scope ] bi*
+    [ ] append-as ;
 
 : (FUNCTOR:) ( -- word def effect )
     scan-new-word [ parse-functor-body ] parse-locals-definition ;
--- a/basis/locals/parser/parser-tests.factor
+++ b/basis/locals/parser/parser-tests.factor
@@ -4,18 +4,6 @@ tools.test vocabs vocabs.parser ;
 IN: locals.parser.tests
 
 <<
-! ((parse-lambda))
-{
-    "V{ 99 :> kkk kkk }"
-} [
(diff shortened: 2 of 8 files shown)


basis/functors/functors.factor
```diff
@@ -159,16 +153,14 @@ DEFER: ;FUNCTOR delimiter
     [
         building get use-words
         (parse-bindings)
-        building get unuse-words
     ] with-bindings ;
 
 : parse-functor-body ( -- form )
-    push-functor-words
+    functor-words use-words
```
mrjbq7: This is outside the scope of the ``with-lambda-scope`` so shouldn't we still ``unuse-words``?

basis/locals/parser/parser-tests.factor
```diff
@@ -67,3 +55,25 @@ IN: locals.parser.tests
     ] with-compilation-unit
     [ locals>> [ name>> ] map ] [ keys ] bi*
 ] unit-test
+
+<<
```
mrjbq7: Does this need ``<<``?

basis/locals/parser/parser.factor
```diff
@@ -31,15 +31,16 @@ ERROR: invalid-local-name name ;
 
 SINGLETON: lambda-parser
 
-: ((parse-lambda)) ( assoc reader-quot: ( -- quot ) -- quot )
+: with-lambda-scope ( assoc reader-quot: ( -- quot ) -- quot )
     '[
         in-lambda? on
         lambda-parser quotation-parser set
-        [ use-words @ ] [ unuse-words ] bi
+        manifest [ clone [ clone ] change-qualified-vocabs ] change
```
mrjbq7: Seems like a clean fix -- another option would be to define ``M\ manifest clone`` and all four containers in it (``search-vocab-names``, ``search-vocabs``, ``qualified-vocabs``, ``auto-used``).  On the other hand, this is a simple specific fix which I like.

mrjbq7: Actually, we already define ``M\ manifest clone``, so we could just change this to:

```
manifest [ clone ] change
```

mrjbq7:
Okay, I merged this with two small fixes:

functors: fc467b72a59be1850827a3612af929dadc96c36c

locals.parser: e69d0a7b45bab455c9136912376386bfe592424e

basis/locals/parser/parser.factor
bjourne: Nice. I didn't see that.

basis/functors/functors.factor
bjourne: I don't think so because it is only called from `parse-locals-definition` which sets up the "lambda scope" for it. I saw that you added a new scope for it and that should work too.

basis/locals/parser/parser-tests.factor
bjourne: I'm fairly sure it's needed. `<<` and `>>` does some kind of magic so you need them when running tests that tests the parser.

PR #1374, opened 2015-06-27
locals.parser.tests: fixes the unit tests

bjourne:
Trivial fix to make the unit tests run here.

### DIFF
--- a/basis/locals/parser/parser-tests.factor
+++ b/basis/locals/parser/parser-tests.factor
@@ -3,6 +3,8 @@ locals.parser namespaces parser prettyprint sequences sorting
 tools.test vocabs vocabs.parser ;
 IN: locals.parser.tests
 
+<<
+
 ! (::)
 {
     "dobiedoo"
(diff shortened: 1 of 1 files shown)


mrjbq7:
I removed those because I thought they "shouldn't be necessary". And tests pass if you run them yourself but as part of load-all test-all it fails.  I suppose it's bad form to leave a broken build but was hoping to identify what caused the failure. 

Maybe we merge this and then investigate separately?

bjourne:
Oh, sorry I totally misunderstood you then. But you're right it is a little strange that those weird `<<` and `>>` are needed there.

mrjbq7:
Seems like ``manifest`` is being set to ``f`` or unset somehow.

This reproduces the test failures:

```factor
f manifest [ "locals" test ] with-variable
```

mrjbq7:
Closing for now I forgot about this and applied the fix but didn't give you credit, whoops.

PR #1380, opened 2015-07-06
Vm refactor c++14

bjourne:
Here are some c++ code refactorings. I've changed many of the structs into lambda functions which I think are much prettier and is available now since we use c++11. I also replaced the `binary_payload_start` with a `slot_count` method to make address calculations easier. And I removed `tuples.hpp` and `strings.hpp` because tiny header files are annoying. ;)


vm/code_heap.cpp
```diff
@@ -49,43 +49,25 @@ void code_heap::free(code_block* compiled) {
...
-  void operator()(code_block* block, cell size) {
-    FACTOR_ASSERT(all_blocks->find((cell)block) != all_blocks->end());
-  }
-};
-
 void code_heap::verify_all_blocks_set() {
-  all_blocks_set_verifier verifier(&all_blocks);
-  allocator->iterate(verifier);
+  auto all_blocks_set_verifier = [&](code_block* block, cell size) {
+    all_blocks.find((cell)block) != all_blocks.end();
```
mrjbq7: This used to be a ``FACTOR_ASSERT``, should it still be?

vm/collector.hpp
```diff
@@ -95,14 +95,13 @@ template <typename TargetGeneration, typename Policy> struct collector {
   }
 
   void trace_partial_objects(cell start, cell card_start, cell card_end) {
-    object* obj = (object*)start;
-    cell end = start + obj->binary_payload_start();
-    start += sizeof(cell);
+    cell *scan_start = (cell*)start + 1;
```
mrjbq7: Previously ``scan_start = start + sizeof(cell)``, now ``scan_start = start + 1``, is that intentional?

vm/slot_visitor.hpp
```diff
@@ -1,91 +1,92 @@
...
-      return sizeof(string);
-    /* everything else consists entirely of pointers */
-    case ARRAY_TYPE:
-      return array_size<array>(array_capacity((array*)this));
-    case TUPLE_TYPE: {
-      tuple_layout* layout = (tuple_layout*)fixup.translate_data(
-          untag<object>(((tuple*)this)->layout));
-      return tuple_size(layout);
+  cell t = type();
+  if (t == ARRAY_TYPE) {
```
mrjbq7: Why not one big switch statement?

vm/code_heap.cpp
mrjbq7: Oh, I see, this is protected by a ``ifdef FACTOR_DEBUG``...

vm/collector.hpp
mrjbq7: Because it is ``start++``, I see.

mrjbq7:
Merged!

erg:
clang be like:
```
vm/code_heap.cpp:68:34: warning: inequality comparison result unused [-Wunused-comparison]
    all_blocks.find((cell)block) != all_blocks.end();
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
```

mrjbq7:
Fixed.

vm/collector.hpp
jonenst: Because integer arithmetic (start + sizeof(cell)) was replaced with the equivalent pointer arithmetic ((cell*)start + 1).

PR #1383, opened 2015-07-10
Image analyzer

bjourne:
I've written a tool for introspecting Factor images. It's useful to have when trying to debug why the deploy image files seems to be growing. You can load and count how many objects there are in the data and code heaps and so on.


mrjbq7:
Ooooooooh, Christmas in July!  This looks like a neat new toy to play with, thanks!!

erg:
Awesome!

Do you think this could be merged with or subsume ``bootstrap.image``?

bjourne:
Thanks. :) Not yet, because bootstrap.image mostly writes while these vocabs reads. But it should be possible with some more work to have a set of common vocabs shared by both image writers and readers.

PR #1385, opened 2015-07-12
Save image fix

bjourne:
Small fix to the image saving. I noticed that the filesystem path of the image is being saved in it. That's a small privacy hole and redundant so I changed `primitive_save_image` so that those paths aren't saved.


mrjbq7:
Merged, thanks!!

bjourne:
:+1: think new boot images is needed too

mrjbq7:
Oh right, not sure if I did that or not but did again just in case.

PR #1390, opened 2015-07-15
Fuel colors and scaffold

bjourne:
Here is a few fixes for even more colors and better scaffolding in fuel.


mrjbq7:
Merged!

I don't use fuel so I always assume these changes work.  :-)

PR #1391, opened 2015-07-15
Remove sub-primitive wrapper words in threads.private?

mrjbq7:
Inside `threads.private` are these "simple" primitive definitions with a warning that we need to have these primitives wrapped in an inline word that calls them:

``` factor
PRIMITIVE: (set-context) ( obj context -- obj' )
PRIMITIVE: (set-context-and-delete) ( obj context -- * )
PRIMITIVE: (start-context) ( obj quot -- obj' )
PRIMITIVE: (start-context-and-delete) ( obj quot -- * )

! Wrap sub-primitives; we don't want them inlined into callers
! since their behavior depends on what frames are on the callstack

: set-context ( obj context -- obj' )
    (set-context) ; inline

: start-context ( obj quot: ( obj -- * ) -- obj' )
    (start-context) ; inline

: set-context-and-delete ( obj context -- * )
    (set-context-and-delete) ; inline

: start-context-and-delete ( obj quot: ( obj -- * ) -- * )
    (start-context-and-delete) ; inline
```

Is this still needed?  I removed the aliases and renamed the words (like `(set-context)` becomes `set-context`) and the tests seem to pass.

How would I test for the problem the warning is describing?

### DIFF
--- a/basis/bootstrap/assembler/ppc.factor
+++ b/basis/bootstrap/assembler/ppc.factor
@@ -793,7 +793,7 @@ IN: bootstrap.ppc
     3 jit-switch-context
     jit-push-param ;
 
-[ jit-set-context ] \ (set-context) define-sub-primitive
+[ jit-set-context ] \ set-context define-sub-primitive
 
 : jit-pop-quot-and-param ( -- )
     3 ds-reg 0 jit-load-cell
@@ -815,7 +815,7 @@ IN: bootstrap.ppc
     jit-push-param
     jit-jump-quot ;
 
-[ jit-start-context ] \ (start-context) define-sub-primitive
+[ jit-start-context ] \ start-context define-sub-primitive
 
 : jit-delete-current-context ( -- )
     jit-load-context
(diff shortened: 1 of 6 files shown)


erg:
The non-optimizing compiler can't inline the sub-primitives if they are in wrappers like this. We should just clarify the comment.

mrjbq7:
I'm a little confused, you mean in the existing code (before my patch), the non-optimizing compiler can inline the sub-primitives, but if we apply the patch then it can't?

mrjbq7:
Can we fix the non-optimizing compiler?

If we can't fix it, then will this work the same?

```factor
ALIAS: set-context (set-context)
ALIAS: set-context-and-delete (set-context-and-delete)
ALIAS: start-context (start-context)
ALIAS: start-context-and-delete (start-context-and-delete)
```

erg:
The non-optimizing compiler ignores ``inline`` on regular words but knows how to inline sub-primitives.

mrjbq7:
Why would it work inlining a sub-primitive in one word definition (the alias), but not in another (a different word that calls it directly)?

mrjbq7:
Closing for now.

Issue #1389, opened 2015-07-14
UI theme improvement


nicolas-p:
I am proposing a small UI theme redesign which I think is a nice improvement.

This is a screenshot of the current UI:

![current factor ui](https://cloud.githubusercontent.com/assets/11431580/8683274/07bdf6e8-2a72-11e5-965f-edc7b501ad59.png)

This is a mock-up I made:

![redesigned factor ui](https://cloud.githubusercontent.com/assets/11431580/8683286/14c6605a-2a72-11e5-9605-55f113c2ba96.png)

List of changes:
- top buttons unified in a toolbar (colour F2F2F2 with a 1-px border of colour B3B3B3)
- "Search" label placed inside the search box and replaced by the input when the user types
- headers extend the full width of the window
- some text placement corrections
- Apple-style scroll bars: hidden when not needed, simple black bars otherwise
- status bar height increased by 5 px
- tip of the day removed (or at least there should be a cross to close it; in this case the blue background should also extend the full width of the window)

What do you think?

mrjbq7:
That looks nice, we implement our own UI in OpenGL so building "apple-style" scroll-bars might be a little harder than changing the theme but I'd have to look into it more.

I particularly like your button theme, much more "modern".

The headers extended full width of the window looks really great.

I like seeing the tip-of-the-day on new listener windows, do you find it annoying?

Have you had a chance to implement any of these, or is this just a mockup at this point?

Thanks!

nicolas-p:
I'm glad you like it!

Concerning scrolI-bars, I have just obtained something close to what I had in mind just by modifying the TIFF images and restarting Factor. I end up with a lot of transparent images.

I shall try to implement the other ideas in the next few days.

I've never liked tips of the day in any software. But here they're clearly not as intrusive as the pop-up windows we see often.

mrjbq7:
I generally agree on tips of the day, but figured most people are unaware of all the features Factor provides so thats why I'm a little biased the other way. :-)

nicolas-p:
There's progress. It is not a mock-up this time, this is a real screenshot:

![capture d ecran 2015-07-16 a 22 03 07](https://cloud.githubusercontent.com/assets/11431580/8733882/81035242-2c06-11e5-839c-20eca7f268f3.png)


Implemented:
- scroll-bars
- unified toolbar (except border)
- search label removed
- documentation header extending the full width and always visible at the top when scrolling
[...]

PR #1396, opened 2015-07-18
UI improvements

nicolas-p:
Implementation of #1389


mrjbq7:
I'm going through these, I think things like making the scrollbar disappear completely (but still take up UI room) isn't ideal.

It would be nice to have the new toolbar buttons as separate button graphics so we can have the old buttons in non-toolbar contexts.

I had to update USING lists (necessary if you don't have ``auto-use`` on):

```diff
diff --git a/basis/ui/gadgets/menus/menus.factor b/basis/ui/gadgets/menus/menus.
factor
index 6d329a5..6b90152 100644
--- a/basis/ui/gadgets/menus/menus.factor
+++ b/basis/ui/gadgets/menus/menus.factor
@@ -2,10 +2,10 @@
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors colors.constants kernel locals math.rectangles
 math.vectors namespaces opengl sequences sorting ui.commands
-ui.gadgets ui.gadgets.buttons ui.gadgets.corners
-ui.gadgets.frames ui.gadgets.glass ui.gadgets.packs
-ui.gadgets.worlds ui.gestures ui.operations ui.pens
-ui.pens.solid ui.render ;
+ui.gadgets ui.gadgets.borders ui.gadgets.buttons
+ui.gadgets.corners ui.gadgets.frames ui.gadgets.glass
+ui.gadgets.packs ui.gadgets.worlds ui.gestures ui.operations
+ui.pens ui.pens.solid ui.render ;
 IN: ui.gadgets.menus
 
 : show-menu ( owner menu -- )
diff --git a/basis/ui/pens/solid/solid.factor b/basis/ui/pens/solid/solid.factor
index 3e3f1a0..24e314d 100644
--- a/basis/ui/pens/solid/solid.factor
+++ b/basis/ui/pens/solid/solid.factor
@@ -1,6 +1,7 @@
 ! Copyright (C) 2009 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: kernel accessors opengl math colors ui.pens ui.pens.caching ;
+USING: kernel accessors opengl math colors sequences ui.pens
+ui.pens.caching ;
 IN: ui.pens.solid
 
 TUPLE: solid < caching-pen color interior-vertices boundary-vertices ;
diff --git a/basis/ui/tools/browser/browser.factor b/basis/ui/tools/browser/brow
ser.factor
index 0ff58ea..9dd7a07 100644
--- a/basis/ui/tools/browser/browser.factor
+++ b/basis/ui/tools/browser/browser.factor
@@ -1,15 +1,15 @@
 ! Copyright (C) 2006, 2009 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors arrays classes combinators
+USING: accessors arrays classes colors.constants combinators
 combinators.short-circuit compiler.units debugger fry help
 help.apropos help.crossref help.home help.stylesheet help.topics
 kernel models sequences sets ui ui.commands ui.gadgets
-ui.gadgets.borders ui.gadgets.buttons ui.gadgets.editors
-ui.gadgets.glass ui.gadgets.labels ui.gadgets.panes
-ui.gadgets.scrollers ui.gadgets.status-bar ui.gadgets.tracks
-ui.gadgets.viewports ui.gadgets.worlds ui.gestures
-ui.tools.browser.history ui.tools.browser.popups ui.tools.common
-vocabs ;
+ui.gadgets.borders ui.gadgets.buttons ui.gadgets.buttons.private
+ui.gadgets.editors ui.gadgets.glass ui.gadgets.labels
+ui.gadgets.panes ui.gadgets.scrollers ui.gadgets.status-bar
+ui.gadgets.tracks ui.gadgets.viewports ui.gadgets.worlds
+ui.gestures ui.pens.solid ui.tools.browser.history
+ui.tools.browser.popups ui.tools.common vocabs ;
 IN: ui.tools.browser
 
 TUPLE: browser-gadget < tool history scroller search-field popup ;
diff --git a/basis/ui/tools/listener/listener.factor b/basis/ui/tools/listener/listener.factor
index 3e1ff52..646c32e 100644
--- a/basis/ui/tools/listener/listener.factor
+++ b/basis/ui/tools/listener/listener.factor
@@ -7,14 +7,15 @@ documents.elements fry hashtables help help.markup help.tips io
 io.styles kernel lexer listener locals make math models
 models.arrow models.delay namespaces parser prettyprint
 quotations sequences source-files.errors strings system threads
-tools.errors.model ui ui.commands ui.gadgets ui.gadgets.buttons
-ui.gadgets.editors ui.gadgets.glass ui.gadgets.labeled
-ui.gadgets.panes ui.gadgets.scrollers ui.gadgets.status-bar
-ui.gadgets.tracks ui.gestures ui.operations ui.pens.solid
-ui.tools.browser ui.tools.common ui.tools.debugger
-ui.tools.error-list ui.tools.listener.completion
-ui.tools.listener.history ui.tools.listener.popups vocabs
-vocabs.loader vocabs.parser vocabs.refresh words ;
+tools.errors.model ui ui.commands ui.gadgets ui.gadgets.borders
+ui.gadgets.buttons ui.gadgets.editors ui.gadgets.glass
+ui.gadgets.labeled ui.gadgets.panes ui.gadgets.scrollers
+ui.gadgets.status-bar ui.gadgets.tracks ui.gestures
+ui.operations ui.pens.solid ui.tools.browser ui.tools.common
+ui.tools.debugger ui.tools.error-list
+ui.tools.listener.completion ui.tools.listener.history
+ui.tools.listener.popups vocabs vocabs.loader vocabs.parser
+vocabs.refresh words ;
 IN: ui.tools.listener
 
 ! If waiting is t, we're waiting for user input, and invoking
```

The walker window (``1 2 +<Ctrl-W>``) has a toolbar on the bottom of the screen.  Perhaps ``solid-underlined`` should just draw underlines on both top and bottom?  Or maybe, configure which border that it "underlines", top or bottom.

I'm not sure about having ``"          Search"`` as the search field label, I guess you wanted some padding from the buttons?  I can look into having the text appear on the field.

I kinda like the next/prev links being in a dependable location (perhaps instead of printing them in a table, we can do something on the same line, maybe like: ``<< prev / next >>``. Speaking of, the fonts are a little small in the next/prev, and hierarchy links at the top.

Generally very favorable to this, though!  Looks more modern and we're getting ready for a 0.98 release so it's good timing.

mrjbq7:
(Let me know if you want to make any of these patches, or I can take this PR and run with it -- I don't mind working on the last bits I mentioned above).

mrjbq7:
Also, thanks for the hard work of doing this!!

mrjbq7:
```factor
: top-vertices-only ( vertices -- top-vertices )
    { 0 1 2 3 2 3 0 1 0 1 } swap nths ;
```

nicolas-p:
I came to some of the same conclusions shortly after I sent the pull request. I ran "demos" and I saw it was a mistake to completely remove the old buttons.

As you said, having the scrollbars still taking up room isn't good. It looks especially bad in the bottom of the Listener. The scrollbars should now appear on top of the content area, not in their own space. But I saw in the "boids" demo that scrollbars were used as sliders, we need to preserve compatibility with this.

Having transparent images is also not really "clean" I guess.

For the walker window, we could have a "top-toolbar" and a "bottom-toolbar". One would draw only the bottom border, and the other only the top border. If we draw both border on all toolbars it won't be beautiful I think. We will end up with 2-px lines.

I was also thinking of having the prev/next buttons on the same line, inside some nice arrow-shaped buttons. They could even be always visible at the bottom of the screen.

I would like to do the prev/next buttons. But the rest might be too difficult for me as I don't know the system well enough. So you can have a go at:
- scrollbars on top of the content, without breaking sliders
- search label inside search box
- separate top-toolbar / bottom-toolbar / normal buttons

I can also try to redesign the normal buttons. And I would like to improve a bit the right-click menu. Isn't it possible to display a platform-native menu?

Apparently the user interface hadn't had any visible change since 2009. Wow!

mrjbq7:
What color do you suggest using for the "labeled" gadgets such as the groups in the walker (``<Ctrl-W>``)?  If we're going to re-theme with a lighter gray toolbar, need a complementary color to use in the other graphics as well as our old buttons.

nicolas-p:
I suggest grey85 (217 217 217).

gggggg

I can make the new background images for labels and buttons with this colour. 

Also, in the Walker, I would put the text "Thread: Listener (suspended)" in a header similar to the one in the Browser.

nicolas-p:
What about this?

walker 3

(It is a mock-up.)

nicolas-p:
And an idea for the previous/next buttons in the Browser:

browser 1

mrjbq7:
I like both of those mockups, especially the walker looks all shiny and the colors are a great idea.

An alternative for the prev/next would be small links in the header:

```< The listener / Runtime code reloading >```

**Comment by nicolas-p** [comment]
I think I have found the right way to implement all 1-px borders.

It will be difficult to implement the walker like above if we use ```bottom-vertices-only``` and ```top-vertices-only``` because we will have to make ```left-vertices-only``` as well and manage every time which element is going to have the border.

We can get rid of that and use piles and shelves that have a grey75 background and a 1-px gap between elements. This is a bit like what I had done the other day except in this case the piles and shelves would actually contain the... content.

I'll try to implement this in the Listener and Browser now, so I have the right foundations to do the Walker and Traceback another day.

**Comment by mrjbq7** [comment]
You might find this useful for your walker mockup:

```factor
"hello" <label>
    [ t >>bold? ] change-font
{ 10 10 } <border>
    { COLOR: white COLOR: grey85 } <gradient> >>interior gadget.
```
You can paint gradients of colors pretty easily, and the gradient can actually be a list of more than two colors, although that might be too much.

nicolas-p:
I'm making good progress on the Walker and Traceback.

I little question: How do I create a new image file from the modified code in my "basis" directory?

mrjbq7:
You have a ``factor.image`` which is a saved version of the code you've changed, so you can just ``refresh-all`` and then ``save`` and it will be there when you restart.

I noticed you might be using ``auto-use`` which at some point we'll have to make sure your ``USING:`` lists are all proper to resolve names without having to throw a restart.

If you're asking about boot images, thats not relevant here because the UI is in "basis" and the boot images are for bootstrapping the system using the code in "core".

mrjbq7:
Don't forget the inspector, error-list, and debugger tools.

mrjbq7:
In your UI listener, is ``IN: scratchpad`` have a white background?

mrjbq7:
Also, the focus is broken when you do "asdfasdf" enter, it should focus the keyboard on the selected line, so you can press up and down without looking for the mouse.

nicolas-p:
Yes, I was aware of all these issues.

For the error pop-up and the menu, I would like a drop shadow. I'll have to make eight TIFF images probably.

Too many elements have a white background at the moment (the toolbar in the error pop-up for instance).

erg:
Is it too much to ask to figure out how to use SVG in the UI?

mrjbq7:
Maybe a first version that just draws a border, doing drop shadow maybe a little fancy.

@erg i don't think we have anything that properly supports SVG rendering natively.

nicolas-p:
I introduced this bug a few days ago:
capture d ecran 2015-07-22 a 20 27 23
I drew a grey background over the entire listener window, and then a white background over what I would call the listener area. The method ```prompt.``` is supposed to paint the text with a red background but this interferes with the white background and I can't find a solution.

mrjbq7:
I can take a look in the next couple days, traveling a little bit right now.

you can play with prompt in the UI listener:

```factor
output-stream get "this is a prompt" prompt. nl
```

It looks a little like a label doesn't draw its own background for some reason, you can see that by removing the ``white-interior`` in ``add-listener-area``:

```diff
 : add-listener-area ( listener -- listener )
     dup output>> margins <scroller> >>scroller
-    dup scroller>> white-interior 1 track-add ;
+    dup scroller>> 1 track-add ;
```

Instead of white, it gets gray background.

I'm not quite sure what's going on but it seems that it's due to some combination of the new layouts, because this works in a window:

```
USING: ui.gadgets.panes.private
H{ 
    { foreground COLOR: black }
    { background COLOR: red }
} "this is a prompt" <styled-label> "test" open-window
```

But not from listener:

```
H{ 
    { foreground COLOR: black }
    { background COLOR: red }
} "this is a prompt" <styled-label> gadget.
```

nicolas-p:
I had to post this screenshot (not a mock-up) :

capture d ecran 2015-07-22 a 21 59 04

mrjbq7:
Haha yeah looks great!!

erg:
This is awesome!

Are you doing it with themes in mind or just hardcoding it to the new look?

On Wed, Jul 22, 2015 at 1:54 PM, John Benediktsson <notifications@github.com

nicolas-p:
It's hardcoded. Themes would be too difficult to do, and not necessary in my opinion.
(Well, a dark theme would not be a bad idea actually , especially as this is a programming environment.)

erg:
People have asked in the past, just wondering.

Ask for too much but be happy with what you get :)

mrjbq7:
I have half way coded theme support but I think we should release the new UI first. We can do light dark theme after.

mrjbq7:
@erg the windows gradient bug is a blocker for this.

nicolas-p:
I have just regrouped all colour definitions in a single dictionary ```ui.gadgets.colors```, which opens the possibility of a dark theme in the future.

mrjbq7:
Do you think the green bar at the top of the walker should be green if thread is ``+running+`` and yellow if ``+suspended+`` or light red if ``+stopped+``?

nicolas-p:
That's a brilliant idea!
I thought it would be easy to implement but my code isn't working.

```factor
CONSTANT: thread-status-stopped-background
T{ rgba { red 0.9568 } { green 0.8509 } { blue 0.8509 } { alpha 1.0 } } inline
CONSTANT: thread-status-suspended-background
T{ rgba { red 0.9568 } { green 0.9450 } { blue 0.8509 } { alpha 1.0 } } inline
CONSTANT: thread-status-running-background
T{ rgba { red 0.9295 } { green 0.9569 } { blue 0.8510 } { alpha 1.0 } } inline

: add-thread-status ( track -- track )
    dup status>> dup self <thread-status> margins
    swap value>> {
      { +stopped+ [ thread-status-stopped-background ] }
      { +suspended+ [ thread-status-suspended-background ] }
      { +running+ [ thread-status-running-background ] }
    } case <solid> >>interior
    f track-add ;
```

mrjbq7:
You need to set the color when things change, not just in the beginning when you add-thread-status.

Also this patch fixes your render bugs with ``IN: scratchpad``, basically the gadget background of labels was being ignored in favor of the parent's background.  I want to fix this properly later tonight or tomorrow, but you can try this locally to get rid of that problem so you can focus on other stuff:

```diff
diff --git a/basis/ui/gadgets/labels/labels.factor b/basis/ui/gadgets/labels/labels.factor
index c95ac71..b946af6 100644
--- a/basis/ui/gadgets/labels/labels.factor
+++ b/basis/ui/gadgets/labels/labels.factor
@@ -1,6 +1,6 @@
 ! Copyright (C) 2005, 2009 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors arrays colors.constants combinators fonts
+USING: accessors arrays colors.constants combinators fonts fry
 kernel make math.functions models namespaces sequences splitting
 strings ui.baseline-alignment ui.gadgets ui.gadgets.tracks
 ui.pens.solid ui.render ui.text ;
@@ -65,12 +65,11 @@ M: label cap-height*
     label-metrics cap-height>> round ;
 
 M: label draw-gadget*
-    >label<
-    [
-        background get [ font-with-background ] when*
+    [ >label< ] [ gadget-background ] bi
+    '[
+        _ background get or [ font-with-background ] when*
         foreground get [ font-with-foreground ] when*
-    ] dip
-    draw-text ;
+    ] dip draw-text ;
 
 M: label gadget-text* string>> % ;
```

mrjbq7:
This is maybe a terrible way to do it but fixes your color thread state thing:

```diff
diff --git a/basis/ui/tools/walker/walker.factor b/basis/ui/tools/walker/walker.factor
index ce74706..d672416 100644
--- a/basis/ui/tools/walker/walker.factor
+++ b/basis/ui/tools/walker/walker.factor
@@ -52,14 +52,27 @@ M: walker-gadget focusable-child*
         drop
     ] "" make ;
 
+: update-thread-status-color ( status -- )
+    [ walker-gadget? ] find-window children>> first
+    children>> second swap
+    {
+        { +stopped+ [ COLOR: red ] }
+        { +suspended+ [ COLOR: yellow ] }
+        { +running+ [ COLOR: green ] }
+        { f [ COLOR: white ] }
+    } case <solid> >>interior drop ;
+
 : <thread-status> ( model thread -- gadget )
-    '[ _ walker-state-string ] <arrow> <label-control> ;
+    '[
+        [ _ walker-state-string ]
+        [ update-thread-status-color ] bi
+    ] <arrow> <label-control> ;
 
```

mrjbq7:
Here's a cleaner way to do the thread-status, it makes a custom ``model-changed`` that calculates the new text and the new color:

```factor
: thread-status-text ( status thread -- string )
    [
        "Thread: " %
        dup name>> %
        " (" %
        swap {
            { +stopped+ "Stopped" }
            { +suspended+ "Suspended" }
            { +running+ "Running" }
        } at %
        ")" %
        drop
    ] "" make ;

: thread-status-color ( status -- color )
    {
        { +stopped+ COLOR: red }
        { +suspended+ COLOR: yellow }
        { +running+ COLOR: green }
    } at ;

TUPLE: thread-status < label thread ;

M: thread-status model-changed
    [ value>> ] dip {
        [ [ thread>> thread-status-text ] [ string<< ] bi ]
        [ [ thread-status-color <solid> ] [ parent>> interior<< ] bi* ]
    } 2cleave ;

: <thread-status> ( model thread -- gadget )
    "" thread-status new-label
        swap >>thread
        swap >>model ;
```

mrjbq7:
I committed a fix for the label issue in 5fbe6aa9e6eb68b525e0ef56df1d6b2bcc2de3b9, you should rebase this branch on master when you get a chance -- note that @erg has been doing some refactoring so you might need to bootstrap after rebasing.

mrjbq7:
Alternatively, you can keep going on this branch and know that the label issue is fixed and we can rebase later.

By the way, have you noticed some minor rendering bugs with the toolbar buttons?  If you click press them some like ``End`` in the listener are all dark, but ``Clear Stack`` doesn't paint the dark gray all the way down:

Image

mrjbq7:
(Also, I edited the thread-status example above to be a little cleaner)

mrjbq7:
Also, the ``<border-button>`` doesn't look right in the boids demo:

![Image]
(http://i.imgur.com/uSJAPHB.png)

nicolas-p:
Thank you, it works.

At the moment, it is difficult to see the difference between running and suspended because the colours I've chosen are close. I tried to steal GitHub's red and green and to make a matching yellow, but the result wasn't beautiful. I need to find the right colours.

I made changes in toolbars yesterday, that is probably when I introduced the bug you saw. Unfortunately I haven't got a Retina display so I can't see the problem. Tell me if this fixes it:
```factor
IN: ui.gadgets.toolbar
: <toolbar> ( target -- toolbar )
    horizontal <track>
        1 >>fill
        1/2 >>align
        { 5 5 } >>gap
        swap
        [ [ "toolbar" ] dip class-of get-command-at commands>> ]
        [ '[ [ _ ] 2dip <toolbar-button> f track-add ] ]
        bi assoc-each ;
```
But I'm certain this is going to create other problems (in the Browser and Error List, particularly).

```<border-button>```'s don't look right because I had first replaced FactorTan with white, and some time later I wanted to revert the change but I wrote FactorLightTan. I was aware of the problem but I didn't bother to correct it because I am thinking of designing new buttons.

By the way, I see on your screenshots that the bitmaps don't look good on your Retina display.

**Comment by nicolas-p** [comment]
I think the last commit fixes the toolbar problem. Tell me if it works. As I said, on my screen I can't see the difference.

**Comment by nicolas-p** [comment]
I found the right colours for the Walker. I managed to keep the same green that I liked.

To emphasise the difference, I suggest this:

capture d ecran 2015-07-26 a 15 30 31

**Comment by mrjbq7** [comment]
You should move the ``HELP: <toolbar>`` section from ``ui/gadgets/buttons/buttons-docs.factor`` to ``ui/gadgets/toolbars/toolbars-docs.factor``. 

Both ``ui.gadgets.slots`` and ``ui.tools.deploy`` need a ``USE: ui.gadgets.toolbars``.

**Comment by mrjbq7** [comment]
Also, the issue with the graphics looking pixelated on retina is because we don't currently load ``2x`` graphics that would be twice the resolution on a retina display... there's also some issues with Windows on retina display as well because the text is pixelated.

Out of curiosity, which platform are you developing on?

**Comment by mrjbq7** [comment]
Also, note for later: the changes to the ``$navigation-table`` break the generation of next/prev links on the [docs.factorcode.org](docs.factorcode.org) website.

I haven't thought about it much, but I think we may want ``print-topic`` to do what it used to do (so ``apropos`` has a ``$title`` and the ``help.html`` isn't busted), and use ``print-content`` to print just the body of the help article. 

Or perhaps another way of having next/prev links on the webpages and titles on the ``apropos``.

**Comment by mrjbq7** [comment]
Also, highlighting of selected text is busted for the ``IN: scratchpad``:

![image]
(http://i.imgur.com/soiSReH.png)

**Comment by mrjbq7** [comment]
This is almost ready to merge, I can feel it getting close!!

**Comment by mrjbq7** [comment]
Your bold / colored thread status labels are really nice.

**Comment by nicolas-p** [comment]
I had also thought about the online documentation. I was crossing my fingers, hoping it would still work.

I have found a super easy way to fix our scrollbar problem (they still take up space even if they're not needed and it's not pretty). I made a 1-px-by-1-px transparent image that replaces all my previous transparent images. Scrollbars were still taking space. I then also replaced *horizontal-scroller-right.tiff* with a transparent pixel and it worked. Toolbar handle pictures would have to be slightly adapted because they don't appear in the right place. The thing is that we loose the arrows. I personally don't need them but I don't know if you're prepared to sacrifice them.

capture d ecran 2015-07-27 a 09 16 26
(I only modified the horizontal one.)

**Comment by nicolas-p** [comment]
Actually, this is probably a bad way of doing it. I see that scrollers are implemented with sliders, and I don't want to break that. There must be a bit of code somewhere in ```ui.gadgets.scrollers``` which makes the scrollbar appear and disappear when necessary. We should just modify it to also display or remove the scrollbar background as well.

**Comment by mrjbq7** [comment]
Don't worry, I can fix the online documentation later when this is ready to merge.

**Comment by mrjbq7** [comment]
I wonder if you should move your definitions from ``ui.gadgets.colors`` to ``ui.gadgets.themes`` instead so we keep all the "skinning" together?

**Comment by nicolas-p** [comment]
Couldn't I even move ```ui.gadgets.colors``` and the bitmaps to ```ui.gadgets.theme.light```?
Then ```ui.gadgets.theme``` would just redirect towards ```ui.gadgets.theme.light```.

Some time later, we can implement a theme switching mechanism in ```ui.gadgets.theme``` and add ```ui.gadgets.theme.dark```. (To do that, we would have to create a ```content-font-color``` constant and to use it everywhere text is displayed.)

**Comment by mrjbq7** [comment]
Hah, yes doing a light vocab would be good too!

**Comment by nicolas-p** [comment]
I tried the following change:

```factor
IN: ui.gadgets.sliders.private
M: slider pref-dim*
    [ dup slider-enabled? 
      [ dup interior>> ] 
      [ { 0 0 } ] 
      if pen-pref-dim 
    ] [ drop { 100 100 } ] [ orientation>> ] tri
    set-axis ;
```
Now when I open a browser window large enough to hold the content, the scrollbars appear but once I resize the window, they don't take any space, even if I make the window too small. When I open a browser window too small for the content, the scrollbars take space, even if I make the window larger.

So this is a first step in the right direction, but there is probably something else to change near ```update-scroller```.

mrjbq7:
You might need to call ``relayout`` from ``M\ scroller layout*`` to cause it to recalculate ``pref-dim``?

nicolas-p:
Most of the things I've tried end up in an infinite loop apparently. How would you write the new ```layout*``` method?

mrjbq7:
Hmm, I see the infinite loop, looks like thats the wrong place to insert a ``relayout``.  I also tried in ``model-changed`` which is where we do it for ``labels`` but that doesn't seem to work.  Will have to look at it later.

Also, I think your ``M\ slider pref-dim*`` should be a little different?

```factor
M: slider pref-dim*
    [
        dup slider-enabled? [
            dup interior>> pen-pref-dim
        ] [
            drop { 0 0 }
        ] if
    ] [ drop { 100 100 } ] [ orientation>> ] tri
    set-axis ;
```

Note: the ``drop { 0 0 }`` and moving the ``pen-pref-dim`` inside the true block of the if.

nicolas-p:
Is it possible to define all colour constants in ```ui.gadgets.theme.light``` but to make them appear in ```ui.gadgets.theme``` to be used by other dictionaries?

Have you had time to look at our scrollbar problem since yesterday?

mrjbq7:
In ``ui.gadgets.theme.light`` you can have the constants (not the ``IN:`` does not say "light"):

```factor
IN: ui.gadgets.theme
CONSTANT: text-color COLOR: black
...
```

And then in ``ui.gadgets.theme``, you can have this somewhere (maybe at the bottom), we use ``<<`` to make a nested compilation unit that requires the "light" vocabulary before finishing loading of the "theme" so the constants are available immediately by "theme" users:

```factor
<< "ui.gadgets.theme.light" require >>
```

We use this pattern for our ``libc`` constants:

https://github.com/slavapestov/factor/blob/master/basis/libc/libc.factor#L125

Basically, when it loads the "light" vocabulary, it defines the constants in the "theme" vocabulary.  @erg and I have talked about other (cleaner) ways to do this, but this will work fine for now.

mrjbq7:
I looked at the scroller problem, but then got busy and didn't finish it.  I'll try and look at it soon.

nicolas-p:
I did a lot of code duplication in `help` and in `ui.tools.browser` when I implemented the new prev/next links. I'm trying to refactor it using dynamic variables but there is something I don't understand.

I don't need to show you all the code but the important parts are :
```factor
SYMBOL: direction
: prev ( -- ) -1 direction set ;
: next ( -- ) 1 direction set ;
```
```factor
: add-help-footer ( track -- track )
    horizontal <track> with-lines
    prev (add-help-footer)
    next (add-help-footer)
    f track-add ;
```
With this, I get two next links.

If I write:
```
    next (add-help-footer)
    prev (add-help-footer)
```
I get two previous links.

It looks like my `prev` and `next` words are not executed at the right time, thus both `(add-help-footer)` produce the same link. What did I do wrong?

mrjbq7:
If I had to guess without seeing more of the code, it's something to do with where you do ``direction get``... see this for example:

```factor
SYMBOL: direction
: prev ( -- ) -1 direction set ;
: next ( -- ) 1 direction set ;

: (add-help-header) ( -- quot )
    [ direction get ] ;
```

Getting something like this:

```factor
IN: scratchpad prev (add-help-header)
               next (add-help-header) [ call ] bi@ . .
1
1

IN: scratchpad next (add-help-header)
               prev (add-help-header) [ call ] bi@ . .
-1
-1
```

A way to fix it might be to do something like this where you get the symbol, and then curry its value into whatever needs it?

```factor
: (add-help-header) ( -- quot )
    direction get '[ _ ] ;
```

nicolas-p:
You guessed right. `direction get` was called in two places, and they were both inside quotations inside quotations inside quotations. So I gave up dynamic variables and instead went back to something I had done earlier with lexical variables and that worked well.

I think we still have two things to do before this branch can be merged:
- improve scrollbars (make them completely disappear and reappear when necessary)
- implement the partly bold and partly red/orange/green status text in the Walker

They both seem too difficult for me at the moment.

mrjbq7:
The problem with the scrollbars, is I'm not sure a good way to get them to appear without displacing or relayout of the parent widget.

Maybe we should merge most of your work and then have some secondary cleanup?  It looks so much nicer, I keep switching to this branch to do my development :-)

mrjbq7:
(We also have the Windows bug #1397 with gradient backgrounds which I spent some time on now that I have a Windows 10 VM setup).

mrjbq7:
I get this error in the branch:

```
resource:basis/ui/tools/browser/browser.factor: 45

Asset: <help-footer>

Stack effect declaration is wrong
inferred ( x x -- x x )
declared ( browser-gadget direction -- gadget )
```

nicolas-p:
I can't find the bug. `<help-footer>` looks good to me.

mrjbq7:
This fixes it for me:

```diff
diff --git a/basis/ui/tools/browser/browser.factor b/basis/ui/tools/browser/browser.factor
index 4631308..d6e61a7 100644
--- a/basis/ui/tools/browser/browser.factor
+++ b/basis/ui/tools/browser/browser.factor
@@ -43,7 +43,7 @@ M: browser-gadget set-history-value
     { 1 0 } >>fill f track-add ;
 
 : <help-footer> ( browser-gadget direction -- gadget )
-    [ model>> ] dip [ '[ _ _ $navigation ] try ] <pane-control>
+    [ model>> ] dip '[ [ _ $navigation ] try ] <pane-control>
     { 0 0 } <border> { 1/2 1/2 } >>align
     toolbar-background <solid> >>interior ;
```

nicolas-p:
Yes it works, but I'll admit I don't understand why.

mrjbq7:
Seems like ``<pane-control>`` expects ``( model quot -- )`` but this line is actually ``( model direction quot -- )``:

```[ model>> ] dip [ '[ _ _ $navigation ] try ] <pane-control>```

If you change it to curry the direction, then it's the right stack effect:

```[ model>> ] dip '[ [ _ $navigation ] try ] <pane-control>```

**Comment by mrjbq7** [comment]
@nicolas-p what do you think of making a big UI step forward by merging this with some small changes:

* not using gradients behind text because of Windows bug
* having always visible scrollbars

Then we can get back to those two bugs later when I have time to figure out how to fix it?

I really like all these improvements and don't want to let this bitrot!

**Comment by mrjbq7** [comment]
Also, don't worry about rebasing this branch on master (I've been doing that all along and can make sure to merge it properly).

**Comment by mrjbq7** [comment]
The merge conflict right now is just adding ``classes.error`` to the USING: list of ``basis/help/help.factor``.

**Comment by mrjbq7** [comment]
Also, this implements your colored walker text:

```factor
: thread-status-text ( status -- string )
    {
        { +stopped+ "Stopped" }
        { +suspended+ "Suspended" }
        { +running+ "Running" }
    } at "(" ")" surround ;

: thread-status-foreground ( status -- color )
    {
      { +stopped+   [ COLOR: red ] }
      { +suspended+ [ COLOR: orange ] }
      { +running+   [ COLOR: green ] }
      { f           [ text-color ] }
    } case ;

: thread-status-background ( status -- color )
    {
      { +stopped+   [ thread-status-stopped-background ] }
      { +suspended+ [ thread-status-suspended-background ] }
      { +running+   [ thread-status-running-background ] }
      { f           [ content-background ] }
    } case ;

TUPLE: thread-status < label ;

M: thread-status model-changed
    [ value>> ] dip {
        [ [ thread-status-text ] [ string<< ] bi* ]
        [ [ thread-status-foreground ] [ font>> foreground<< ] bi* ]
        [ [ thread-status-background <solid> ] [ parent>> parent>> interior<< ] bi* ]
    } 2cleave ;

: <thread-status> ( model -- gadget )
    "" thread-status new-label
        swap >>model ;

: add-thread-status ( track -- track )
    horizontal <track> { 5 5 } >>gap
        "Thread:" <label>
            dup font>> t >>bold? drop
            f track-add
        self name>> <label> f track-add
        over status>> <thread-status>
            dup font>> t >>bold? drop
            f track-add
    margins f track-add ;
```

nicolas-p:
I would say it is ready to be merged.

I used a condition on the OS to display a flat colour only on Windows and keep the gradient on other platforms.

The toolbars can stay the way they are for the moment: when they are not needed, they take some white space. It's not too disturbing, and I think it's still better than the always-visible toolbar background that was used previously.

mrjbq7:
Merged!!!!

Thanks so much for the hard work! Let's let everyone try it in ``master`` and see what people think!!

mrjbq7:
@nicolas-p I implemented your idea to have the "Search" field say "Search" if you haven't typed anything.

If you're curious how it works, you can look at f24444e99019a68bfa8fa546a3ea8b5218e0f4cb.

nicolas-p:
I wanted to see what the latest nightly build looks like on Windows (to check that it doesn't have a gradient). There is an error message when I want to open the browser:

probleme factor

I managed to get a normal browser by clicking on Inspector Help, but even that doesn't work all the time.

The traceback says it has something to do with the baseline alignment that I introduced in toolbars (it is mainly useful for the error list toolbar).

nicolas-p:
Also, one late thought: I think the bitmaps I created (new buttons, and transparent toolbar background) are rather large (in term of kilobytes) compared to the originals.

There is a commit from seven years ago (bb46c64f71b52381e56640811851a388856419d7) that implements LZW-compressed images. Maybe someone can re-compress my bitmaps.

mrjbq7:
@nicolas-p can you try latest nightly build? there was a lot of changes but git master and latest nightly build work for me on windows 10.

Thanks!  Good idea about images, will have to look into that.

mrjbq7:
Looks like these are already ``Compression: LZW``?

```
$ identify -verbose horizontal-scroller-left.tiff 
Image: horizontal-scroller-left.tiff
  Format: TIFF (Tagged Image File Format)
  Mime type: image/tiff
  Class: DirectClass
  Geometry: 8x16+0+0
  Resolution: 49.15x49.15
  Print size: 0.162767x0.325534
  Units: PixelsPerCentimeter
  Type: Bilevel
  Base type: Bilevel
  Endianess: LSB
  Colorspace: Gray
  Depth: 8/1-bit
  Channel depth:
    gray: 1-bit
    alpha: 1-bit
  Channel statistics:
    Pixels: 128
    Gray:
      min: 255 (1)
      max: 255 (1)
      mean: 255 (1)
      standard deviation: 0 (0)
      kurtosis: 0
      skewness: 0
      entropy: nan
    Alpha:
      min: 0 (0)
      max: 0 (0)
      mean: 0 (0)
      standard deviation: 0 (0)
      kurtosis: 0
      skewness: 0
      entropy: nan
  Alpha: graya(255,0)   #FFFFFF00
  Colors: 1
  Histogram:
       128: (255,255,255,  0) #FFFFFF00 graya(255,0)
  Rendering intent: Perceptual
  Gamma: 0.454545
  Chromaticity:
    red primary: (0.64,0.33)
    green primary: (0.3,0.6)
    blue primary: (0.15,0.06)
    white point: (0.3127,0.329)
  Background color: graya(255,1)
  Border color: graya(223,1)
  Matte color: graya(189,1)
  Transparent color: graya(0,0)
  Interlace: None
  Intensity: Undefined
  Compose: Over
  Page geometry: 8x16+0+0
  Dispose: Undefined
  Iterations: 0
  Compression: LZW
  Orientation: TopLeft
  Properties:
    date:create: 2015-08-16T08:05:44-07:00
    date:modify: 2015-08-16T08:05:44-07:00
    signature: 9d6077540c9ba55f393f4f9b5d513cdee5376daae6046394d7d10569c3105c87
    tiff:alpha: unassociated
    tiff:document: /Applications/Factor/basis/ui/gadgets/theme/horizontal-scroller-left.tiff
    tiff:endian: lsb
    tiff:photometric: RGB
    tiff:rows-per-strip: 64
  Artifacts:
    filename: horizontal-scroller-left.tiff
    verbose: true
  Tainted: False
  Filesize: 396B
  Number pixels: 128
  Pixels per second: 128PB
  User time: 0.000u
  Elapsed time: 0:01.000
  Version: ImageMagick 6.9.1-10 Q16 x86_64 2015-07-26 http://www.imagemagick.org
```

nicolas-p:
I looked at the bug on Windows with the new nightly build. It looked better at first but once I tried typing into the search box, the same error message popped up. (It is the 64-bit version on Windows 7.)

erg:
@nicolas-p I pushed a fix for the popup. I'm enjoying the new ui, great job!

nicolas-p:
The Browser problem is solved. Thank you.

jonenst:
I was reading through this, and was wondering what the walker status +stopped+ was ? I only see +running+ and +suspended+ during normal usage ?

PR #1399, opened 2015-07-21
Compiler regressions fix

bjourne:
This fixes a small regression in the compiler which probably was introduced by one of my earlier compiler pr:s. For some reason stack frames were generated even for leaf words which made the generated assembly code a little longer. But it should now be fixed.


mrjbq7:
Merged, thanks!!

PR #1400, opened 2015-07-21
cpu.x86: use push and pop when emitting %prologue and %epilogue

bjourne:
A fun trick they use in llvm is to replace some add/sub rsp instructions with push and pops. The latter instructions take one byte each while the former takes four. We can use that too. It reduces the code size by 0.87% :) So, not so much but it was an easy optimization.


mrjbq7:
Merged (awhile ago), thanks!!

PR #1406, opened 2015-07-26
Bignumslashf

jonenst:
Hi,
this PR fixes arithmetic errors in /f for results in range ]0x1.0p-1022,0x0.4p-1022], for example ```1 1023 2^ /f```
It also "fixes" the algorithm that would lose a lot of performance by mistake
After I understood the bignum/f algorithm, I add comment and simplified things like a loop that could only execute once.

I found these errors after working on https://github.com/slavapestov/factor/issues/1231


mrjbq7:
Merged, thanks!!

PR #1408, opened 2015-07-26
Dos numberstring

jonenst:
There are several things in this PR that I found while working on the main feature. I'm not sure about the last one for 1+1/3. but a test depended on it so I put it here. What do you think ?

### DIFF
--- a/core/math/parser/parser-tests.factor
+++ b/core/math/parser/parser-tests.factor
@@ -99,6 +99,8 @@ unit-test
 
 { 0x1.999999999999ap-3 } [ "0.2" string>number ] unit-test
 { 0x1.3333333333333p0  } [ "1.2" string>number ] unit-test
+{ 0o1.146314631463146314p0 } [ "1.2" string>number ] unit-test
+{ 0b1.0011001100110011001100110011001100110011001100110011p0 } [ "1.2" string>number ] unit-test
 { 0x1.5555555555555p0  } [ "1.333,333,333,333,333,333" string>number ] unit-test
 { 0x1.aaaaaaaaaaaabp0  } [ "1.666,666,666,666,666,666" string>number ] unit-test
 
@@ -240,6 +242,40 @@ unit-test
 { 0.0 } [ "1.0p-1075" hex> ] unit-test
 { 1/0. } [ "1.0p1024" hex> ] unit-test
 { -1/0. } [ "-1.0p1024" hex> ] unit-test
+
+{ 1.0 } [ "1.0p0" bin> ] unit-test
+{ 1.5 } [ "1.1p0" bin> ] unit-test
+{ 1.875 } [ "1.111p0" bin> ] unit-test
+{ 1.90625 } [ "1.11101p0" bin> ] unit-test
+{ 1.03125 } [ "1.00001p0" bin> ] unit-test
+{ 15.5 } [ "1111.1p0" bin> ] unit-test
+{ 15.53125 } [ "1111.10001p0" bin> ] unit-test
+{ -15.5 } [ "-1111.1p0" bin> ] unit-test
+{ 15.5 } [ "1111.1p0" bin> ] unit-test
+{ -15.5 } [ "-1111.1p0" bin> ] unit-test
+{ 62.0 } [ "1111.1p2" bin> ] unit-test
+{ 3.875 } [ "1111.1p-2" bin> ] unit-test
+{ $[ 1 bits>double ] } [ "1.0p-1074" bin> ] unit-test
+{ 0.0 } [ "1.0p-1075" bin> ] unit-test
+{ 1/0. } [ "1.0p1024" bin> ] unit-test
+{ -1/0. } [ "-1.0p1024" bin> ] unit-test
+
+{ 1.0 } [ "1.0p0" oct> ] unit-test
+{ 1.5 } [ "1.4p0" oct> ] unit-test
+{ 1.875 } [ "1.7p0" oct> ] unit-test
+{ 1.90625 } [ "1.72p0" oct> ] unit-test
+{ 1.03125 } [ "1.02p0" oct> ] unit-test
+{ 15.5 } [ "17.4p0" oct> ] unit-test
+{ 15.53125 } [ "17.42p0" oct> ] unit-test
+{ -15.5 } [ "-17.4p0" oct> ] unit-test
+{ 15.5 } [ "17.4p0" oct> ] unit-test
+{ -15.5 } [ "-17.4p0" oct> ] unit-test
+{ 62.0 } [ "17.4p2" oct> ] unit-test
+{ 3.875 } [ "17.4p-2" oct> ] unit-test
+{ $[ 1 bits>double ] } [ "1.0p-1074" oct> ] unit-test
+{ 0.0 } [ "1.0p-1075" oct> ] unit-test
+{ 1/0. } [ "1.0p1024" oct> ] unit-test
+{ -1/0. } [ "-1.0p1024" oct> ] unit-test
 
 { 0 } [ "0" string>number ] unit-test
 { 0 } [ "00" string>number ] unit-test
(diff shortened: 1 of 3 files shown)


jonenst:
OOps forgot to commit something, and will remove that last 1+1/3. commit, wait a sec

jonenst:
Ok. I pushed -f . Should be good now. I included some other commits because I think they can go in, if not we can just skip them

mrjbq7:
Does this bootstrap?

I get this:

```
resource:basis/math/functions/functions.factor

233: : most-negative-finite-float ( -- x )
234:     -0x1.ffff,ffff,ffff,fp1023 >integer ; inline
                                   ^
T{ no-math-method f 1 2 / }

The following restarts are available:

:1      Load resource:basis/math/functions/functions.factor again
:2      Load resource:basis/math/ratios/ratios.factor again
Factor 0.98 x86.64 (1682, heads/master-0-gb9c0255, Tue May 12 09:30:20 2015)
```

jonenst:
I hadn't tested bootstrapping.. it's because of
8c476a9 math.parser: don't lose precision in make-float-bin-exponent

I'm not sure what is happening ? While recompiling math.ratios and math.functions, the parser can't do arithmetic like 1 2 / that creates a ratio ? Not sure if we should work around in math.functions and define the constant somewhere else, or ensure that the parser doesn't use ratios, or something else ?

mrjbq7:
Hmm, I'm not sure why ratios would not be available - I'll try and look at this later today.

jonenst:
Anyway this is not new, the following patch breaks bootstrap even before the changes of this PR
```diff
diff --git a/basis/math/functions/functions.factor b/basis/math/functions/functi
index 6b1382f..bea47c8 100644
--- a/basis/math/functions/functions.factor
+++ b/basis/math/functions/functions.factor
@@ -230,6 +230,7 @@ M: complex log >polar [ flog ] dip rect> ; inline
 
 <PRIVATE
 
+: useless ( -- x ) 1.1e10;
 : most-negative-finite-float ( -- x )
     -0x1.ffff,ffff,ffff,fp1023 >integer ; inline

```
So vocabs loaded by basis/bootstrap/math/math.factor during the ```"math.ratios" require``` phase were already not allowed to used dec floats with positive exponents (because the math.parser implementation uses ratios for these. My change is to also use ratios for hex floats with positive exponents). Note that floats with negative exponents work because ratios are not used...

jonenst:
I don't know the bootstrap system enough, but does this mean we should move math.ratios to core ?

mrjbq7:
Hmm, I see.  ``math.ratios`` depends on ``math.functions``.  And ``math.parser`` depends on ``math.ratios`` in some cases...

jonenst:
This workaround fixes bootstrap
```diff
diff --git a/basis/math/functions/functions.factor b/basis/math/functions/functions.factor
index 6b1382f..d338129 100644
--- a/basis/math/functions/functions.factor
+++ b/basis/math/functions/functions.factor
@@ -230,11 +230,9 @@ M: complex log >polar [ flog ] dip rect> ; inline
 
 <PRIVATE
 
-: most-negative-finite-float ( -- x )
-    -0x1.ffff,ffff,ffff,fp1023 >integer ; inline
+CONSTANT: most-negative-finite-float -17976931348623157081452742373170435679807056752584499659891747680315726078002853876058955863276
 
-: most-positive-finite-float ( -- x )
-    0x1.ffff,ffff,ffff,fp1023 >integer ; inline
+CONSTANT: most-positive-finite-float  17976931348623157081452742373170435679807056752584499659891747680315726078002853876058955863276
 
 CONSTANT: log-2   0x1.62e42fefa39efp-1
 CONSTANT: log10-2 0x1.34413509f79ffp-2
```

mrjbq7:
What if we changed it to not do ``recip`` and instead do ``1.0 swap /``, does that still have precision problems?

jonenst:
```1.0 swap /``` still has precision problems. Even more tests fail with it (like 1.0e2 then becomes 99.9999999999999)

**Comment by mrjbq7** [comment]
I think if we move ``gcd`` to ``math``, then we can remove the requirement on ``math.functions`` and probably move ``math.ratios`` into ``core`` so that ``math.parser`` can depend on it?

@erg what do you think?  These would be moved:

```factor
FROM: math => gcd fast-gcd lcm ;
```

mrjbq7:
And if we move ``rect>`` and ``>rect``, we can move ``math.complex`` which is I think a good idea.

erg:
The number tower was meant to be opt-in but nobody seems to not include
ratios or complex numbers. I guess it's fine to move those into core.

On Thu, Jul 30, 2015 at 9:04 AM, John Benediktsson <notifications@github.com

mrjbq7:
Well in this case math.ratios would be needed to parse floats by math.parser, so thats a good reason for "core".

mrjbq7:
So, I moved ``math.ratios`` to core so we can use them in ``math.parser``, and I get this test failure on this branch?

```
==== resource:core/math/parser/parser-tests.factor

resource:core/math/parser/parser-tests.factor: 453

Unit Test: {
    { 1.25 }
    [
        "1+1/4" 3000 [ 48 ] "" replicate-as append
        "e-3000" append string>number
    ]
}

=== Expected:
1.25
=== Got:
f
```

Want to try rebasing on master and looking at that?

mrjbq7:
Also, it appears that ``benchmark.parse-float`` is faster, but the ``benchmark.parse-fixnum`` and ``benchmark.parse-bignum`` get slower?

jonenst:
It needs https://github.com/jonenst/factor/commit/f582ab1a0d1a22a4588b0c8b33348e77164fae6f to pass this test. I had included it in the first version of this PR, but it's controversial and I didn't remember why I had included it so I removed it in the second version.. Sorry about that.

So maybe remove or comment out this test for now and discuss what we want in https://github.com/slavapestov/factor/pull/1362 ?

Basically, the main problem with https://github.com/jonenst/factor/commit/f582ab1a0d1a22a4588b0c8b33348e77164fae6f is that it has the following behavior
```factor
1+1/1e10 ! 1.0000000001
1+1/1e100 ! 1.0
1+1/1e1000 ! f
```
which is not intuitive (the last one is f because we check that the right part of ```+``` is not number= to an integer to refuse input like ```1+0``` or ```1+0.0```, but by the time we do this check, we can't tell if the input was ```1+0.0``` or ```1+1/1e1000```)

jonenst:
> Also, it appears that benchmark.parse-float is faster, but the benchmark.parse-fixnum and benchmark.parse-bignum get slower?

Well when parsing non floats number, we are incrementing the 'magnitude' field at each digit for nothing because we only find out that it's a float at the end when parsing the dot or the exponent. But I didn't expect that to be a big performance hit... How much slower is it ?

mrjbq7:
Looks like a bit of a regression for integers, and not the speedup I thought.

Old:

```factor
IN: scratchpad gc [ parse-fixnum-benchmark ] time
Running time: 0.427108349 seconds

IN: scratchpad gc [ parse-bignum-benchmark ] time
Running time: 0.578912179 seconds

IN: scratchpad gc [ parse-ratio-benchmark ] time
Running time: 0.224219623 seconds

IN: scratchpad gc [ parse-float-benchmark ] time
Running time: 0.209668632 seconds
```

New:

```factor
IN: scratchpad gc [ parse-fixnum-benchmark ] time
Running time: 0.805471984 seconds

IN: scratchpad gc [ parse-bignum-benchmark ] time
Running time: 0.62233446 seconds

IN: scratchpad gc [ parse-ratio-benchmark ] time
Running time: 0.272957543 seconds

IN: scratchpad gc [ parse-float-benchmark ] time
Running time: 0.235760426 seconds
```

jonenst:
Looking at the diff, I accidently removed "inline" from (add-digit), can
you try putting it back ?

Jon

On Thu, Jul 30, 2015 at 8:14 PM, John Benediktsson <notifications@github.com

mrjbq7:
Okay, will do!

erg:
If you move ratios and complex to core, then you have larger deploy images. You can't strip out math.

mrjbq7:
math already had ratio and complex tuple. I added gcd to it, though.

erg:
Are you saying that core/ already had ratio and complex tuples defined in core?

mrjbq7:
Yeah in core/math/math.factor.

mrjbq7:
Okay, merged!  Committed b3323c15bd914d2dc3b518703ac6b9c5b8c55b76 which fixed some performance issues.

Issue #1308, opened 2015-04-22
Possible bug in SSE code generation


bjourne:
```factor
USING: arrays kernel math math.vectors math.vectors.simd.cords sequences ;
IN: test9

! In this test, we should try to have a SIMD register live when gc is
! performed.
: test-simd-gc ( a b -- c )
    v+ dup first 10 > [ second ] [ third ] if 1array ; inline

: do-test ( -- a )
    double-4{ 2 20 30 40 } double-4{ 2 4 3 2 }
    test-simd-gc ;

IN: scratchpad do-test .
{ 5.349853483873686e-317 }
```

The result should be `{ 33.0 }` (cause 30+3=3). I think it reads from an uninitialized spill slot which causes the weird looking value. Maybe something with normal 64bit registers being 8 byte long and SSE registers being 16 bytes.

mrjbq7:
Hmm, sounds like a job for @slavapestov !

mrjbq7:
@bjourne did you ever look into this more?

bjourne:
Not yet! :+1:

bjourne:
Works now!

PR #1411, opened 2015-07-28
Compiler fixes again #1308

bjourne:
This pr should fix the #1308 bug. A lot of the commits are just docs I wrote or some small refactorings. The bug appears to have been that it made a vreg the leader of another as long as they had the same register class. But they also need to have the same representation size.

### DIFF
--- a/basis/bootstrap/image/image-docs.factor
+++ b/basis/bootstrap/image/image-docs.factor
@@ -1,4 +1,5 @@
-USING: help.markup help.syntax io io.files io.pathnames strings ;
+USING: bootstrap.image.private help.markup help.syntax io io.files
+io.pathnames quotations strings words ;
 IN: bootstrap.image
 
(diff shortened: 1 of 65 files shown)


mrjbq7:
Wow, awesome!!!

Merged!!!

mrjbq7:
@bjourne saw this test failure on macosx/x86.64

```
==== resource:basis/compiler/cfg/linear-scan/allocation/state/state-tests.factor

resource:basis/compiler/cfg/linear-scan/allocation/state/state-tests.factor: 112

Unit Test: {
    { t }
    [
        3 { } insns>cfg [ align-spill-area ] keep
        spill-area-align>> cell =
    ]
}

=== Expected:
t
=== Got:
f
[Traceback]
```

bjourne:
Oh I see. I fix I fix

PR #1413, opened 2015-07-29
compiler.cfg.utilities: make it so block>cfg initializes spill-area-a…

bjourne:
This fixes the test failure!


mrjbq7:
Merged, thanks!

PR #1422, opened 2015-08-06
Fuel syntax updates etc

bjourne:
Here is another bunch of fuel fixes, mostly in the syntax highlighting area.


misc/fuel/factor-mode.el
```diff
@@ -317,16 +306,15 @@ these lines in your .emacs:
 (defconst factor-symbols-lines-regex
   (concat (syntax-begin '("SYMBOLS")) ws+ symbols-to-semicolon))
 
-(defconst factor-int-constant-def-regex
-  (factor-second-word-regex '("ALIEN" "CHAR" "NAN")))
+;; (defconst factor-int-constant-def-regex
```
mrjbq7: Probably can remove this at some point, since you moved it down below.

Also, should the ``syntax-and-1-symbol`` be ``factor-`` prefixed like the other stuff? I'm not sure how namespaces work in Emacs.

mrjbq7:
Merged!!

misc/fuel/factor-mode.el
bjourne: Yes, mostly. But it's so tedious writing `factor-` prefixes everywhere so I thought this could be some exceptions.

PR #1423, opened 2015-08-08
Serialize

jonenst:
Hi,
this "fixes" the serialize vocab which would emit numbers in range [2^1008; 2^1024[ with 131 bytes, but they can be emitted in 129 bytes for free (and the comments pretended it dit). It's also backwards compatible in the sense that old serialized values still deserialize as the correct answer.

### DIFF
--- a/basis/serialize/serialize-tests.factor
+++ b/basis/serialize/serialize-tests.factor
@@ -5,7 +5,7 @@ USING: tools.test kernel serialize io io.streams.byte-array
 alien arrays byte-arrays bit-arrays specialized-arrays
 sequences math prettyprint parser classes math.constants
 io.encodings.binary random assocs serialize.private alien.c-types
-combinators.short-circuit ;
+combinators.short-circuit literals ;
 SPECIALIZED-ARRAY: double
 IN: serialize.tests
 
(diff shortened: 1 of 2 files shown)


mrjbq7:
Awesome, merged!!

mrjbq7:
Question for you: should this be ``0x7f <=`` ?

https://github.com/slavapestov/factor/blob/master/basis/serialize/serialize.factor#L41

jonenst:
No, ```0x7e <=``` is correct.
If you change it to 0x7f <=, then 0x7f (127) would be serialized as B{ 0xff } (because 0x7f | 0x80 = 0xff), which conflicts with numbers> 2^1024 that are serialized as B{ 0xff .... }

I experimented with serializing 127 in 1byte (as B{ 0xff }) instead of 2 by changing the marker to 0x80 and starting the last form at 2^1016 instead of 2^1024, but the cost of that is that all numbers in range [2^1016, 2^1024[ go from 129 bytes to 131 bytes. And it's not backwards compatible. So I figured it was not worth it.

mrjbq7:
Okay, I see.

I think this comment is more correct?

```diff
-! 127<=x<2^1024 => B{ length(x) x[0] x[1] ... }; 1<length(x)<129 fits in 1 byte
+! 127<x<2^1024 => B{ length(x) x[0] x[1] ... }; 1<length(x)<129 fits in 1 byte
```

jonenst:
Hm no, I think that's good. 127 is serialized as B{ length value }: B{ 1 127 }
0xfe is 126 | 0x80. Maybe change ```0xfe <=``` to ```0xff <``` to make it more clear that it's related to the comment that was " <= 126" which I changed to "< 127" ?

mrjbq7:
Ahh right, I see my mistake -- thanks!

PR #1425, opened 2015-08-08
authors, use utf-8 to fix Björn's name

jonenst:
I noticed @bjourne 's name had an unprintable character on http://docs.factorcode.org/content/article-vocab-authors.html. Factor is pretty good for working with non ascii characters so we need to fix it ! :)

It was written in latin1 in ./extra/pcre/authors.txt, this commit changes it back to utf-8.


mrjbq7:
Awesome, good fix! Merged!

PR #1426, opened 2015-08-08
curses.listener: fix, need stream-write1 after changes to stream-bl

jonenst:
a simple fix for curses.listener which wouldn't start at all because it didn't implement stream-write1. It is now needed since 708224644775f8777b273118b810db5463d5d47c


mrjbq7:
Merged, thanks!!

PR #1441, opened 2015-08-15
Fix some callstack overflows

bjourne:
Sorry for the bad branch name! 

The branch fixes two callstack overflow problems. 1) If you are almost out of callstack, then do a major gc, then you run out of callstack, which SIGSEGV and factor crashes. So we unlock the guard page for the duration of the gc cycle which gives it 4096 extra bytes. 2) If you are out of callstack and call a primitive, it fubbles up the callstack_top pointer. So I added code to handle that case. There are still some annoying overflow bugs left.. Many of the commits are also code simplifications or random cleanups.


mrjbq7:
Merged, nice fixes and cleanup!

PR #1454, opened 2015-08-22
Vm cleanups 2015 aug part2

bjourne:
Here is few vm cleanups I've worked on. I removed the `primitive_datastack` primitives because the code can use `primitive_datastack_for` and similar instead.


mrjbq7:
Good cleanups, merged!! 

Thanks!

PR #1465, opened 2015-08-30
Formatting

jonenst:
Small docs fixes for "formatting" and an enhancement for peg.ebnf to allow to escape double quotes so you can make rules that contain both single and double quotes.


mrjbq7:
Merged, thanks!

I spent some time trying to make it so you didn't have to ``flatten``, but am not sure a clean way besides a messy way of making a ``<parse-result>`` that effectively does a ``push-all`` instead of ``push``.

PR #1466, opened 2015-09-02
Windows fixes part3

bjourne:
Here is another VM refactoring branch. It removes the `RT_EXCEPTION_HANDLER` relocation type which was only used on Windows and replaces it with a special object. Then it changes the ffi to not look for decorated symbols on win32. It makes the code for loading ffi functions a bit simpler. Guess I can try committing this myself? :) The boot images will need to be updated.


mrjbq7:
Looks good to me (haven't tested it but if it works locally for you let the build bots have a go!)

Do you know how to upload new boot images?

```
./factor -run=bootstrap.image.upload
```

I will send you password separately.

mrjbq7:
Yay!!

bjourne:
Hm it crashes. I'll try uploading that boot image.

mrjbq7:
Looks like some bootstrap.image test failures.  Also, for some reason formatting tests fail on Windows.  There's a difference between the number of exponent characters and I can't remember right now where that was fixed up, so maybe not caused by you?

erg:
Vs 2013 has the buggy formatting. I restarted with vs2015.

bjourne:
I see the bootstrap.image failure. I'll try and push a fix.

Issue #1257, opened 2014-12-20
win32 binary not working on Windows XP SP3.


mrjbq7:
Reported on the mailing list:

http://www.mail-archive.com/factor-talk@lists.sourceforge.net/msg07036.html

PR #1469, opened 2015-09-10
math.parser: support >base for byte-array

AlexIljin:
Hey! I've added byte-array support to math.parser >base, so now I can do things like:
"license.txt" sha-256 checksum-file >hex .

Please, let me know if it was a bad idea, otherwise please pull.

Also, I was only able to test this on my older Factor installation v0.96, because that's the latest stable release that works in Windows.

### DIFF
--- a/core/math/parser/parser-tests.factor
+++ b/core/math/parser/parser-tests.factor
@@ -224,6 +224,7 @@ unit-test
 { "0.0" } [ 0.0 >hex ] unit-test
 { "1.0p-1074" } [ 1 bits>double >hex ] unit-test
 { "-0.0" } [ -0.0 >hex ] unit-test
+[ "deadbeef" ] [ B{ 222 173 190 239 } >hex ] unit-test
 
 { "1.0p0" } [ 1.0 >bin ] unit-test
 { "1.1p2" } [ 6.0 >bin ] unit-test
--- a/core/math/parser/parser.factor
+++ b/core/math/parser/parser.factor
@@ -486,6 +486,9 @@ M: ratio >base
         [ drop "+" glue ]
     } case ;
 
+M: byte-array >base
+    [ >base ] curry { } map-as concat ;
+
 <PRIVATE
 
 : fix-float ( str -- newstr )


mrjbq7:
The ``hex-string`` word in ``checksums`` vocabulary does that:

```factor
IN: scratchpad B{ 222 173 190 239 } hex-string .
"deadbeef"
```

(I'm not saying your patch is bad, just that the feature exists for you to use.)

mrjbq7:
Doesn't 0.97 work in Windows? Or are you using Windows XP?

AlexIljin:
I'm using XP, and I've reported the issue about 0.97 in the mailing list some time ago.

Thanks for the pointer to hex-string! I'll use that if you reject the pull, which would be no problem.
It's so hard to come up with new stuff in Factor, you have covered all the bases, and I mean that as a compliment. Code reuse in Factor is fantastic.

mrjbq7:
We love all contributions!  I have to talk to @erg about ``>hex`` and see it makes sense.  I might be wrong, but the recent nightly build might work on Windows XP, could you check?

mrjbq7:
Also, note your patch is probably wrong for this case:

```factor
IN: scratchpad B{ 1 2 3 4 } hex-string .
"01020304"
```

versus your patch which doesn't pad with zeros:

```factor
IN: scratchpad B{ 1 2 3 4 } [ >hex ] { } map-as concat .
"1234"
```

mrjbq7:
(This is the bug that maybe fixed Windows XP for you #1257)

jonenst:
I don't think it's a good idea to have >base work on byte arrays. ```[ >base ] keep base>``` should be as close as possible to the identity, and for byte arrays it would change the type

It's true that hex-string is hard to find in checksums.

Maybe rename hex-string to bytes>hex and add hex>bytes ? Not sure to what vocab, maybe basis/byte-arrays/hex, but it's not used by default ...
I found 3 occurences of hex>bytes in the whole git repo:
./basis/byte-arrays/hex/hex.factor:    2 <groups> [ hex> ] B{ } map-as
./basis/windows/ole32/ole32.factor:    [ 3 tail concat 2 group [ hex> ] B{ } map-as ] bi
./basis/colors/hex/hex.factor:    2 group [ hex> 255 /f ] map first3 1.0 <rgba> ;

although the last is reallty: hex>bytes [ 255 /f ] { } map-as
and the correct implementation should use <groups> not group

AlexIljin:
John, you are correct about the zero padding, my bad.

AlexIljin:
About the nightly build, Factor 0.98 x86.32 (1712, heads/master-75e50ec5e0, Sun Sep  6 08:25:43 2015) is running fine on my XP system. That's great! Looking forward to the next stable release!

mrjbq7:
Awesome!!

mrjbq7:
It's has a bunch of UI improvements made by @nicolas-p, a ton of VM and compiler work by @bjourne and a bunch of other stuff by the rest of us!

mrjbq7:
Here's implementation of ``bytes>hex`` and ``hex>bytes`` that don't use ``grouping`` so they can be used from ``core``.  I suggest putting them in ``math.parser``... @erg?

```diff
diff --git a/core/math/parser/parser.factor b/core/math/parser/parser.factor
index 4664c8b..601f587 100644
--- a/core/math/parser/parser.factor
+++ b/core/math/parser/parser.factor
@@ -1,8 +1,8 @@
 ! Copyright (C) 2009 Joe Groff, 2013 John Benediktsson
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors byte-arrays combinators kernel kernel.private
-layouts make math math.private namespaces sbufs sequences
-sequences.private splitting strings strings.private ;
+layouts make math math.private sbufs sequences sequences.private
+strings strings.private ;
 IN: math.parser
 
 <PRIVATE
@@ -577,3 +577,23 @@ M: float >base
     } cond ;
 
 : # ( n -- ) number>string % ; inline
+
+: hex>bytes ( hex -- bytes )
+    dup length 2/ <byte-array> [
+        [
+            [ digit> ] 2dip over even? [
+                [ 16 * ] [ 2/ ] [ set-nth ] tri*
+            ] [
+                [ 2/ ] [ [ + ] change-nth ] bi*
+            ] if
+        ] curry each-index
+    ] keep ;
+
+: bytes>hex ( bytes -- hex )
+    dup length 2 * CHAR: 0 <string> [
+        [
+            [ 16 /mod [ >digit ] bi@ ]
+            [ 2 * dup 1 + ]
+            [ [ set-nth ] curry bi-curry@ bi* ] tri*
+        ] curry each-index
+    ] keep ;
```

mrjbq7:
(And then we remove ``hex-string`` and replace all usages with ``bytes>hex``)

erg:
Looks fine.

jonenst:
You can close this.

PR #1472, opened 2015-09-22
Fix 1370

### LINKED ISSUE #1370: get-ifaddrs compiles incorrectly
I was trying to list my network interfaces with this code on Mac 64:

```factor
USING: accessors alien.c-types alien.destructors alien.syntax
classes.struct kernel libc sequences unix.types ;
IN: ifaddrs

STRUCT: ifaddrs
    { ifa_next ifaddrs* }
    { ifa_name c-string }
    { ifa_flags uint }
    { ifa_addr sockaddr* }
    { ifa_netmask sockaddr* }
    { ifa_dstaddr sockaddr* }
    { ifa_data void* } ;


FUNCTION: int getifaddrs ( ifaddrs** ifap ) ;

FUNCTION: void freeifaddrs ( ifaddrs* ifp ) ;

DESTRUCTOR: freeifaddrs

: get-ifaddrs ( -- ifaddrs )
    ifaddrs <struct>
    dup getifaddrs io-error
    [ ifa_next>> ] follow
    [ ifa_name>> ] map B ;
```

What's weird is that this works with ``B`` (and using the Listener), but if you remove the breakpoint it seems to infinite loop.

Yes, I know it's not freeing memory properly yet, but...

Any ideas about the infinite loop / freezing factor?

@bjourne @erg @slavapestov

bjourne:
This pr fixes #1370. It's a little long because it includes unrelated changes too. But most of that is new docs and unit tests. I thought it would be polite to first make a pr since it's a lot of commits. But if you like it maybe I can go ahead and commit it?


mrjbq7:
Looks good! Split-uses is a little strange that it hides n=4 but hopefully that's for a reason.

PR #1479, opened 2015-09-29
Doc fixes for upstream

AlexIljin:
These are only the documentation changes, but I also took the liberty to rename some variables in some stack effect declarations, purely for better readability and consistency.

### DIFF
--- a/basis/io/files/info/info-docs.factor
+++ b/basis/io/files/info/info-docs.factor
@@ -2,7 +2,7 @@ USING: arrays help.markup help.syntax kernel io.files ;
 IN: io.files.info
 
 HELP: file-info
-{ $values { "path" "a pathname string" } { "info" file-info } }
+{ $values { "path" "a pathname string" } { "info" file-info-tuple } }
 { $description "Queries the file system for metadata. If " { $snippet "path" } " refers to a symbolic link, it is followed. See the article " { $link "file-types" } " for a list of metadata symbols." }
 { $errors "Throws an error if the file does not exist." } ;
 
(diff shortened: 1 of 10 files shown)


mrjbq7:
I'm not sure about the vector stack effect change since ``u`` and ``v`` are frequently used for vectors.

See, for example:

https://en.wikipedia.org/wiki/Vector_notation#Operations

AlexIljin:
OK, but the word is v+n, not u+n. See what I mean?

mrjbq7:
Well, I think the desire was to have the same first arg whether it had 1-arity or 2-arity, so if you follow that:

```
: vadd ( v w -- x )
: v+n ( v n -- x )
```

But generally, most math seems to use ``u + v`` or ``u * v`` for operations... ?

mrjbq7:
Cherrypicked the first three commits.  @erg can weigh in on the math.vectors change?

AlexIljin:
Yes, and I only changed things for consistency so that in vneg, vnot, vsqrt etc. you actually see the "v" parameter. Same for v^n, v/n, and so on. Purely a cosmetic fix.

AlexIljin:
u+v and such are left intact.

mrjbq7:
ahh, gotcha.  @erg?

mrjbq7:
P.S. thanks for the contributions!!!

AlexIljin:
Happy to help!

AlexIljin:
It seems that @erg has left the building : ))

erg:
Ha. I'm here, I forgot to reply!

Last patch seems fine!

mrjbq7:
Okay, pulled the last commit, thanks @AlexIljin!

PR #1491, opened 2015-10-22
Fix clang version pattern

arkrost:
'Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)' was parsed incorrectly.

### DIFF
--- a/build-support/factor.sh
+++ b/build-support/factor.sh
@@ -113,7 +113,7 @@ set_md5sum() {
 }
 
 semver_into() {
-	CLANG_RE_OLD="^([0-9]*)\.([0-9]*)-(.*)?$" # 3.3-5
+	CLANG_RE_OLD="^([0-9]*)\.([0-9]*)-?(.*)?$" # 3.3-5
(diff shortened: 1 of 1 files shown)


mrjbq7:
So it used to parse ``3.3-5`` and now would parse ``3.3-``?

arkrost:
```
CLANG_RE_OLD="^([0-9]*)\.([0-9]*)-(.*)?$" # 3.3-5
```
In fact it used to parse '3.3-5 tratata'.  There must be major version, minor version and some special part that *must* start with dash.

```
RE_SEMVER="^([0-9]*)\.([0-9]*)\.([0-9]*)-?(.*)?$" # 3.3.3-5
```
On the next line second pattern doesn't require special part to start with dash.

My clang compiler version is 'Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)'. And ```semver_into``` had to parse '6.0 (clang-600.0.57) (based on LLVM 3.5svn)'. Special part starts with whitespace, nut not with dash. So it is not parsed and I can't use ```build-support/factor.sh update``` because of ```unsupported version number, please report a bug```.

erg:
Merged. Sorry about the bug, thanks for the fix!

PR #1497, opened 2015-10-29
Auto load urls.secure when needed

### LINKED ISSUE #1493: Load urls.secure when needed
It's kind of annoying to have to ``"urls.secure" require`` when you have a vocab that depends on an ``https`` URL.  Better would be for the URLs vocabulary to load ``urls.secure`` if someone creates a URL that requires SSL.

rationalrevolt:
Fixes #1493 

Changes:
1. Added new word `ensure-vocab-loaded` in `vocabs`
2. Modified word `url-addr` to ensure `urls.secure` is loaded when word is invoked on secure urls

### DIFF
--- a/basis/urls/urls.factor
+++ b/basis/urls/urls.factor
@@ -4,7 +4,7 @@
 USING: accessors arrays assocs combinators fry hashtables
 io.pathnames io.sockets kernel lexer make math.parser
 namespaces peg.ebnf present sequences splitting strings
-strings.parser urls.encoding vocabs.loader ;
+strings.parser urls.encoding vocabs vocabs.loader ;
 
 IN: urls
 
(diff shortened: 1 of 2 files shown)


mrjbq7:
This doesn't seem to make the ``hacker-news`` vocabulary work without the ``"urls.secure" require`` line.

Thoughts?

mrjbq7:
Maybe it needs to be in ``>url``?  I haven't looked to see what words the ``http.client`` uses...

rationalrevolt:
I think you might need to do the following before you test:
"vocabs" reload
"urls" reload

It seems like the default image already has these two vocabularies loaded and factor isn't trying to reload them from the updated vocabulary files when you launch the listener.

mrjbq7:
Weird, I thought I had.

Merged, thanks!!

mrjbq7:
Also, the build test failures are just because ``vocabs.loader`` change requires a new boot image.  That's been uploaded so the next time the build bots try again it will upload nightly binaries properly.

PR #1500, opened 2015-11-03
Show persistent border on right clicking presentation gadgets

### LINKED ISSUE #1499: Right click doesn't keep object highlighted
If you right click on a link or object in the listener, the pop-up menu appears, but the object doesn't remain highlighted so it's not super obvious what the thing was that you clicked on.

rationalrevolt:
Fixes #1499

This commit introduces a persistent border around `presentation` gadgets by using a `border-pen` that always has a border when the `presentation-menu` becomes visible. The hook `hide-glass-hook` is utilized to reset the border back to its usual value when the menu is dismissed or a menu selection is pressed, so that the original roll-over border is set back on the `presentation` gadget.

To test, ensure to run `"ui.gadgets.presentations" reload` to force the a `reload` on the `presentations` vocabulary.

### DIFF
--- a/basis/ui/gadgets/presentations/presentations.factor
+++ b/basis/ui/gadgets/presentations/presentations.factor
@@ -1,12 +1,15 @@
 ! Copyright (C) 2005, 2009 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors kernel namespaces ui.commands ui.gadgets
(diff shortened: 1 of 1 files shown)


mrjbq7:
Nice!  I am traveling now but will test this tomorrow.

mrjbq7:
Thanks!!

mrjbq7:
Merged, thanks @rationalrevolt !!

PR #1521, opened 2016-01-09
Ffi docs

jonenst:
Hi,
small docs improvements for the ffi.

I wasn't so sure about c-string slots from structs, it seems you can only write to them with a byte-array, and reading theseveral times returns strings that are not eq?, so I didn't elaborate on that. If someone is sure of what should happen, maybe it's worth mentionning it.

Also, if there are other places where c-string conversions happen, they should be added..


mrjbq7:
Merged, thanks!

PR #1524, opened 2016-02-12
Add suppport for invoking VSC on a Mac

bpollack:
I don't have a Windows box handy to double-check my Windows edits, but I believe I did everything sanely there.


mrjbq7:
Merged, thanks!!

PR #1525, opened 2016-02-15
Make Visual Studio Code integration go directly to lines

bpollack:
I'm not sure whether this didn't originally exist when @erg wrote this or not, but the `-g` flag seems to work just fine on Windows, Linux, and OS X.


mrjbq7:
Merged, thanks!

PR #1526, opened 2016-02-17
Fix incorrect alien.syntax documentation for FUNCTION:

bpollack:
The library docs indicate that you need a semicolon at the end of
a `FUNCTION:` definition, but that's actually a syntax error. Likewise,
while the parentheses and commas may at one point have been optional,
they emphatically aren't anymore. This patch brings the documentation in
line with the code.

### DIFF
--- a/basis/alien/syntax/syntax-docs.factor
+++ b/basis/alien/syntax/syntax-docs.factor
@@ -26,7 +26,7 @@ HELP: LIBRARY:
 { $notes "Logical library names are defined with the " { $link add-library } " word." } ;
 
 HELP: FUNCTION:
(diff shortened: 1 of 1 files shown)


mrjbq7:
Ah, thanks!  Merged!

We are debating whether to put the semi-colon back, but in some ways it looks nicer the new consistent way with required parentheses.

bpollack:
What we really ought to do is write bindings to `libclang` so we can just automatically dump this crap from C headers. And I am not volunteering for that.

mrjbq7:
Agreed.  Most good FFI layers are doing something like that these days.

erg:
Merged.

PR #1528, opened 2016-02-19
Add support for LibreSSL's libtls library

bpollack:
These are just the low-level FFI bindings; high-level bindings (and maybe an
alternative implementation of io.sockets.secure) will come in a later commit.


erg:
Thanks, merged!

Issue #1504, opened 2015-11-12
SSL certificate errors


mrjbq7:
I think these should throw an error somewhere, or warning, or something:

```
URL" https://wrong-host.badssl.com" http-get
URL" https://self-signed.badssl.com" http-get
```

bpollack:
I drilled into this today and am about 95% convinced this is a PEBKAC issue between us and OpenSSL. Specifically, I am pretty sure that `SSL_get_verify_result` (called from `check-verify-result`) does not do what we think it does. I'd be curious to whether we're okay on Windows, since I think that uses SCrypt.

bpollack:
Narrowed down to a generic issue with how we're doing secure sockets. Simpler test case:

```
USING: io.encodings.utf8 io.sockets io.sockets.secure ;
: req-get-body ( host -- body ) "GET / HTTP/1.1\r\nHost: " "\r\n\r\n" surround ;
: https-conn ( host -- secure ) 443 <inet> <secure> ;
: test-https ( host -- resp ) dup req-get-body swap https-conn utf8 [ write flush 1024 read ] with-client ;
```

You can then trivially verify this works with e.g. `"www.google.com" test-https` and `"expired.badssl.com" test-https`.

Very convinced now this is due to us misusing OpenSSL. I'll poke more later.

bpollack:
…we never load the root certificates (note that `load-certificate-chain` is never invoked by the stdlib). So, because OpenSSL is awesome, this always guarantee certificate authentication succeeds. Even for expired certs. Because why not.

mrjbq7:
Great investigations!

Ping @bjourne who was working on OpenSSL too.

bpollack:
Honestly, my preferred fix for this is to throw out OpenSSL entirely and replace it with [libtls](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man3/tls_accept_fds.3?query=tls_init&sec=3) (from [libressl](http://www.libressl.org/)). Not sure how @bjourne feels about that, but if everyone isn't in the "I'd totally just instantly reject that PR" camp, I can at least poke at it. That'd give us a single, easy-to-redistribute SSL engine (including on Windows) that will be a lot harder for us (or any other Factor devs) to mess up.

mrjbq7:
We'd need to package that with our builds, though?  Or add it is as a requirement like libopenssl?

bpollack:
[...]

PR #1530, opened 2016-02-22
Update OpenSSL calls to work with LibreSSL/newer OpenSSL

bpollack:
`RSA_generate_key` has been deprecated for some time, and is not available in either LibreSSL or newer OpenSSL that are not compiled with deprecated call support. This commit switches to the still-supported `RSA_generate_key_ex`, which is supported by both libraries.

Note that this is still insufficient for Factor to work under LibreSSL, though we now get a call further than before.

Starts heading towards resolving #1504.

### DIFF
--- a/basis/io/sockets/secure/openssl/openssl.factor
+++ b/basis/io/sockets/secure/openssl/openssl.factor
@@ -1,14 +1,14 @@
 ! Copyright (C) 2007, 2008, Slava Pestov, Elie CHAFTARI.
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors alien alien.c-types alien.data alien.strings
-assocs byte-arrays classes.struct combinators destructors fry io
-io.backend io.buffers io.encodings.8-bit.latin1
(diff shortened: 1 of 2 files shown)


basis/io/sockets/secure/openssl/openssl.factor
```diff
@@ -113,9 +125,13 @@ M: rsa dispose* handle>> RSA_free ;
 
 : generate-eph-rsa-key ( ctx -- )
     [ handle>> ]
-    [
-        config>> ephemeral-key-bits>> RSA_F4 f f RSA_generate_key
-        dup ssl-error <rsa> &dispose handle>>
+    [| ctx |
+        RSA_new :> rsa-struct
```
bpollack: I'm not happy with this word, but I wasn't terribly clear how to shorten it meaningfully/get rid of the use of `[|`, either, so I picked some unorthodox formatting so I could at least be clear what was operating on what. Totally happy to alter this as you want.

PR #1531, opened 2016-02-22
Proxy

jonenst:
Hi,
some bug fixes and changes before the main commit which adds proxy support. The api is of course open to discussion :)

### DIFF
--- a/basis/http/client/client-docs.factor
+++ b/basis/http/client/client-docs.factor
@@ -281,6 +281,7 @@ $nl
     "http.client.encoding"
     "http.client.errors"
 }
+"For authentication, only Basic Access Authentication is implemented, using the username/password from the target url. Alternatively, the " { $link set-basic-auth } " word can be called on the " { $link request } " object."
 { $see-also "urls" } ;
 
 ABOUT: "http.client"
--- a/basis/http/client/client-docs.factor
+++ b/basis/http/client/client-docs.factor
@@ -281,7 +281,14 @@ $nl
     "http.client.encoding"
     "http.client.errors"
 }
-"For authentication, only Basic Access Authentication is implemented, using the username/password from the target url. Alternatively, the " { $link set-basic-auth } " word can be called on the " { $link request } " object."
+"For authentication, only Basic Access Authentication is implemented, using the username/password from the target or proxy url. Alternatively, the " { $link set-basic-auth } " or " { $link set-proxy-basic-auth } " words can be called on the " { $link request } " object."
+$nl
+"The http client can use an HTTP proxy transparently, by using the " { $link "http.proxy-variables" } ". Additionally, the proxy variables can be ignored by setting the " { $slot "proxy-url" } " slot of each " { $link request } " manually:"
+{ $list
+    { "Setting " { $slot "proxy-url" } " to " { $link f } " prevents http.client from using a proxy." }
+    { "Setting the slots of the default empty url in " { $slot "proxy-url" } " overrides the corresponding values from the proxy variables." }
+}
+
 { $see-also "urls" } ;
 
 ABOUT: "http.client"
--- a/basis/http/client/client-tests.factor
+++ b/basis/http/client/client-tests.factor
@@ -13,6 +13,7 @@ IN: http.client.tests
 {
     T{ request
         { url T{ url { protocol "http" } { host "www.apple.com" } { port 80 } { path "/index.html" } } }
+        { proxy-url T{ url } }
         { method "GET" }
         { version "1.1" }
         { cookies V{ } }
@@ -27,6 +28,7 @@ IN: http.client.tests
 {
     T{ request
         { url T{ url { protocol "https" } { host "www.amazon.com" } { port 443 } { path "/index.html" } } }
+        { proxy-url T{ url } }
         { method "GET" }
         { version "1.1" }
         { cookies V{ } }
@@ -57,4 +59,146 @@ IN: http.client.tests
         "server: Factor http.server"
     } [ "\n" join ] [ "\r\n" join ] bi
     [ [ read-response ] with-string-reader ] same?
+] unit-test
+
+{ "www.google.com:8080" } [
+    URL" http://foo:bar@www.google.com:8080/foo?bar=baz#quux" authority-uri
+] unit-test
+
+{ "/index.html?bar=baz" } [
+    "http://user:pass@www.apple.com/index.html?bar=baz#foo"
+    <get-request>
+        f >>proxy-url
+    request-uri
+] unit-test
+
+{ "/index.html?bar=baz" } [
+    "https://user:pass@www.apple.com/index.html?bar=baz#foo"
+    <get-request>
+        f >>proxy-url
+    request-uri
+] unit-test
+
+{ "http://www.apple.com/index.html?bar=baz" } [
+    "http://user:pass@www.apple.com/index.html?bar=baz#foo"
+    <get-request>
+        "http://localhost:3128" >>proxy-url
+    request-uri
+] unit-test
+
+{ "www.apple.com:80" } [
+    "http://user:pass@www.apple.com/index.html?bar=baz#foo"
+    "CONNECT" <client-request>
+        f >>proxy-url
+    request-uri
+] unit-test
+
+{ "www.apple.com:443" } [
+    "https://www.apple.com/index.html"
+    "CONNECT" <client-request>
+        f >>proxy-url
+     request-uri
+] unit-test
+
+{ f } [
+    "" "no_proxy" [
+        "www.google.fr" <get-request> no-proxy?
+    ] with-variable
+] unit-test
+
+{ f } [
+    "," "no_proxy" [
+        "www.google.fr" <get-request> no-proxy?
+    ] with-variable
+] unit-test
+
+{ f } [
+    "foo,,bar" "no_proxy" [
+        "www.google.fr" <get-request> no-proxy?
+    ] with-variable
+] unit-test
+
+{ t } [
+    "foo,www.google.fr,bar" "no_proxy" [
+        "www.google.fr" <get-request> no-proxy?
+    ] with-variable
+] unit-test
+
+! TODO support 192.168.0.16/4 ?
+CONSTANT: classic-proxy-settings H{
+    { "http.proxy" "http://proxy.private:3128" }
+    { "https.proxy" "http://proxysec.private:3128" }
+    { "no_proxy" "localhost,127.0.0.1,.allprivate,.a.subprivate,b.subprivate" }
+}
+
+{ f } [
+    classic-proxy-settings [
+       "localhost" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ f } [
+    classic-proxy-settings [
+       "127.0.0.1" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ URL" http://proxy.private:3128" } [
+    classic-proxy-settings [
+       "27.0.0.1" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ f } [
+    classic-proxy-settings [
+       "foo.allprivate" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ f } [
+    classic-proxy-settings [
+       "bar.a.subprivate" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ URL" http://proxy.private:3128" } [
+    classic-proxy-settings [
+       "a.subprivate" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ f } [
+    classic-proxy-settings [
+       "bar.b.subprivate" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ f } [
+    classic-proxy-settings [
+       "b.subprivate" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ URL" http://proxy.private:3128" } [
+    classic-proxy-settings [
+       "bara.subprivate" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ URL" http://proxy.private:3128" } [
+    classic-proxy-settings [
+       "google.com" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ URL" http://proxysec.private:3128" } [
+    classic-proxy-settings [
+       "https://google.com" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
+] unit-test
+
+{ URL" http://proxy.private:3128" } [
+    classic-proxy-settings [
+       "allprivate.google.com" "GET" <client-request> ?default-proxy proxy-url>>
+    ] with-variables
 ] unit-test
--- a/basis/http/client/client.factor
+++ b/basis/http/client/client.factor
@@ -91,7 +91,7 @@ SYMBOL: redirects
     redirects get request get redirects>> < [
         request get clone
         response "location" header redirect-url
-        response code>> 307 = [ "GET" >>method ] unless
+        response code>> 307 = [ "GET" >>method f >>post-data ] unless
         quot (with-http-request)
     ] [ too-many-redirects ] if ; inline recursive
 
--- a/basis/http/client/client.factor
+++ b/basis/http/client/client.factor
@@ -16,7 +16,7 @@ ERROR: too-many-redirects ;
 : write-request-line ( request -- request )
     dup
     [ method>> write bl ]
-    [ url>> relative-url present write bl ]
+    [ url>> relative-url f >>anchor present write bl ]
     [ "HTTP/" write version>> write crlf ]
     tri ;
 
--- a/basis/http/client/client.factor
+++ b/basis/http/client/client.factor
@@ -4,19 +4,43 @@ USING: accessors ascii assocs calendar combinators.short-circuit
 destructors fry hashtables http http.client.post-data
 http.parsers io io.crlf io.encodings io.encodings.ascii
 io.encodings.binary io.encodings.iana io.encodings.string
-io.files io.pathnames io.sockets io.timeouts kernel locals math
-math.order math.parser mime.types namespaces present sequences
-splitting urls vocabs.loader ;
+io.files io.pathnames io.sockets io.sockets.secure io.timeouts
+kernel locals math math.order math.parser mime.types namespaces
+present sequences splitting urls vocabs.loader combinators
+environment ;
 IN: http.client
 
 ERROR: too-many-redirects ;
 
+: success? ( code -- ? ) 200 299 between? ;
+
+ERROR: download-failed response ;
+
+: check-response ( response -- response )
+    dup code>> success? [ download-failed ] unless ;
+
 <PRIVATE
+
+: authority-uri ( url -- str )
+    [ host>> ] [ port>> number>string ] bi ":" glue ;
+
+: absolute-uri ( url -- str )
+    clone f >>username f >>password f >>anchor present ;
+
+: abs-path-uri ( url -- str )
+    relative-url f >>anchor present ;
+
+: request-uri ( request -- str )
+    {
+        { [ dup proxy-url>> ] [ url>> absolute-uri ] }
+        { [ dup method>> "CONNECT" = ] [ url>> authority-uri ] }
+        [ url>> abs-path-uri ]
+    } cond ;
 
 : write-request-line ( request -- request )
     dup
     [ method>> write bl ]
-    [ url>> relative-url f >>anchor present write bl ]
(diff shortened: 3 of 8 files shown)


basis/http/client/client-tests.factor
```diff
@@ -27,6 +28,7 @@ IN: http.client.tests
 {
     T{ request
         { url T{ url { protocol "https" } { host "www.amazon.com" } { port 443 } { path "/index.html" } } }
+        { proxy-url T{ url } }
```
mrjbq7: Why is the ``proxy-url`` an empty url tuple?

jonenst: No string reasons...
- to be like the "url" slot (but url is mandatory and this is optional)
- So that f can mean "no proxy" (but the empty URL could mean no proxy)
- so that the behavior "requests use all values from the variables by default, but fill in if it's partial" translatés directly to call derive-url on this slot (but the

jonenst: Oops.
(But the code can be written differently. It might be more complex but bot that much i guess)

jonenst:
Also, the "http" unit tests become dependant on the proxy variables: they fail if the proxy doesn't work (but they still work if the proxy is correctly configured, with no_proxy etc). Not sure if they should be independent or not.

jonenst:
Is there something I can improve in this PR ?

basis/http/http.factor
```diff
@@ -152,6 +159,7 @@ redirects ;
         <url>
             H{ } clone >>query
         >>url
+        <url> >>proxy-url
```
mrjbq7: I guess I think it's weird for this to be an empty URL?

basis/io/sockets/secure/secure.factor
```diff
@@ -114,3 +114,5 @@ HOOK: accept-secure-handshake secure-socket-backend ( -- )
     { [ os unix? ] [ "io.sockets.secure.unix" require ] }
     { [ os windows? ] [ "io.sockets.secure.windows" require ] }
 } cond
+
+M: object opt-send-secure-handshake send-secure-handshake ;
```
mrjbq7: Why is this ``opt-`` prefixed?

basis/http/client/client.factor
```diff
@@ -111,15 +142,73 @@ SYMBOL: redirects
...
+    url>> host>> "." split
+    get-no-proxy-list [
+        "," split [ "." split no-proxy-match? ] with any?
+    ] [ drop f ] if* ;
+
+: check-proxy ( request proxy -- request' )
+    dup [ host>> ] [ f ] if*
+    [ drop f ] unless [ clone ] dip >>proxy-url ;
+
+: get-default-proxy ( request -- default-proxy )
```
mrjbq7: If you're willing to have a little less performance because of ``>upper`` (and use ``http_proxy`` as the Factor variable instead of ``http.proxy``), could re-write this as something like:

```factor
: get-default-proxy ( request -- default-proxy )
    url>> protocol>> "_proxy" append {
        [ get ]
        [ os-env ]
        [ >upper os-env ]
   } 1|| ;
```

basis/http/client/client.factor
```diff
@@ -111,15 +142,73 @@ SYMBOL: redirects
...
+        <request> swap [ url>> >>url ] [ ?copy-proxy-basic-auth ] bi
+        f >>proxy-url "CONNECT" >>method write-request
+        read-response check-response drop opt-send-secure-handshake
+    ] [ drop ] if ;
+
+! Note: ipv4 addresses are interpreted as subdomains but "work"
+: no-proxy-match? ( host-path no-proxy-path -- ? )
+    dup first empty? [ [ rest ] bi@ ] when tail? ;
+
+: get-no-proxy-list ( -- list )
```
mrjbq7: The way ``no-proxy`` works is weird and not similar to other tools, I think.

basis/http/client/client.factor
```diff
@@ -6,17 +6,40 @@ http.parsers io io.crlf io.encodings io.encodings.ascii
...
+
+ERROR: download-failed response ;
+
+: check-response ( response -- response )
+    dup code>> success? [ download-failed ] unless ;
+
 <PRIVATE
 
+: authority-uri ( request -- str )
+    [ host>> ] [ port>> number>string ] bi ":" glue ;
```
mrjbq7: What is ``port`` is ``f`` ?

basis/http/client/client.factor
```diff
@@ -6,17 +6,40 @@ http.parsers io io.crlf io.encodings io.encodings.ascii
...
+
+: absolute-uri ( request -- str )
+    clone f >>username f >>password f >>anchor present ;
+
+: abs-path-uri ( request -- str )
+    relative-url f >>anchor present ;
```
mrjbq7: Why does this set ``f >>anchor``...?

basis/http/client/client.factor
```diff
@@ -91,7 +122,7 @@ SYMBOL: redirects
     redirects get request get redirects>> < [
         request get clone
         response "location" header redirect-url
-        response code>> 307 = [ "GET" >>method ] unless
+        response code>> 307 = [ "GET" >>method f >>post-data ] unless
```
mrjbq7: Why does this set ``f >>post-data``?

basis/http/client/client.factor
```diff
@@ -111,15 +142,73 @@ SYMBOL: redirects
...
+    [ proxy-url>> ] [ url>> ] bi [ or ] keep ;
+
 : <request-socket> ( -- stream )
-    request get url>> url-addr ascii <client> drop
+    request get request-socket-endpoints [ url-addr ] bi@
+    remote-address set ascii <client> local-address set
     1 minutes over set-timeout ;
 
+: https-tunnel? ( request -- ? )
+    [ proxy-url>> ] [ url>> protocol>> "https" = ] bi and ;
```
mrjbq7: So an empty ``proxy-url`` will return true?

basis/http/client/client.factor
jonenst: see https://github.com/jonenst/factor/commit/c2b3967ed5ca9f853acb71b076aabbaa4233e858

basis/io/sockets/secure/secure.factor
jonenst: It's supposed to mean that it's an optional compilation dependency. I was considering using the same name :  "send-secure-handshake" but I think it's confusing. But I don't like the opt- prefix either, do you have a better idea ?
(or a different solution would be to stop doing optional compilation dependencies to io.sockets.secure from http.client)

basis/http/client/client.factor
jonenst: It will throw an exception on f number>string... That's not supposed to happen because <client-request> calls ensure-port.
Also, I just noticed the stack effect name is wrong, this needs a url..

basis/http/client/client.factor
jonenst: This should be a separate commit, I forgot to split it (like the post-data commit). It fixes a bug because the fragment (=anchor) shouldn't be sent to the server. It's a client thing only. For example, curl does this:
```
$ nc -l 3129
$ curl "http://localhost:3129/foo#bar"

GET /foo HTTP/1.1
User-Agent: curl/7.35.0
Host: localhost:3129
Accept: */*
```

basis/http/client/client.factor
jonenst: Yeah, I'm not sure what's best. What I did was:
- f means without proxy; a url means to try with proxy. So by default it has an empty url, so by default it tries to use a proxy.
- the proxy url is then used with derive-url from the proxy variables. So if you setup the variable and leave the default empty url, you'll get a copy from the variables. But if you fill in some slots of the default url, you'll override these values from the variables.

And at the end, if the resulting proxy-url doesn't have a host, there is no proxy. So without variables and with the default empty url, it doesn't use a proxy.

basis/http/http.factor
jonenst: The <request> word also creates an empty url with an empty query and throws that away in <client-request> (ie all words except calling http-request on a manually created request). So I think it's fairly consistent with the rest of the implementation.

jonenst:
I've split the anchor commit and tried to improve the docs a bit.

basis/http/client/client.factor
jonenst: What is weird ? There was a bug when no_proxy was the empty string or contained 2 consecutives "," (it would always match, ie disable the proxy). I've pushed bceadde which fixes that and add tests

After that's fixed, it's the same as curl:
```sh
# curl
$ http_proxy=http://localhost:3129 no_proxy="" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy=http://localhost:3129 no_proxy="com" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy=http://localhost:3129 no_proxy="om" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy=http://localhost:3129 no_proxy=".com" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy=http://localhost:3129 no_proxy="www.google.com" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy=http://localhost:3129 no_proxy="www.ggle.com" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy=http://localhost:3129 no_proxy="google.com" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy=http://localhost:3129 no_proxy="oogle.com" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy=http://localhost:3129 no_proxy="www.google.com,www.google.fr" curl www.google.com &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy=http://localhost:3129 no_proxy="www.google.com,www.google.fr" curl www.google.fr &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy=http://localhost:3129 no_proxy="www.google.com" curl www.google.fr &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy=http://localhost:3129 no_proxy="216.58.211.99" curl 216.58.211.99 &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY

# factor
$ http_proxy="http://localhost:3129" no_proxy="" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy="http://localhost:3129" no_proxy="com" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy="http://localhost:3129" no_proxy="om" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy="http://localhost:3129" no_proxy=".com" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy="http://localhost:3129" no_proxy="www.google.com" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy="http://localhost:3129" no_proxy="www.ggle.com" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy="http://localhost:3129" no_proxy="google.com" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy="http://localhost:3129" no_proxy="oogle.com" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy="http://localhost:3129" no_proxy="www.google.com,www.google.fr" ./factor -e='USING: http.client kernel ; "http://www.google.com" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy="http://localhost:3129" no_proxy="www.google.com,www.google.fr" ./factor -e='USING: http.client kernel ; "http://www.google.fr" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY
$ http_proxy="http://localhost:3129" no_proxy="www.google.com" ./factor -e='USING: http.client kernel ; "http://www.google.fr" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
PROXY
$ http_proxy="http://localhost:3129" no_proxy="216.58.211.99" ./factor -e='USING: http.client kernel ; "216.58.211.99" http-get 2drop' &> /dev/null ; if [[ $? == "0" ]]  ;then echo NOPROXY; else echo PROXY; fi
NOPROXY

```

basis/http/client/client.factor
jonenst: Yeah that's a nicer implementation. Performance is not an issue I think, so this is all about the name of the variables. I liked "http.proxy" better but that's just one opinion, so we can change it :)

basis/http/client/client.factor
jonenst: I forgot to mention that check-proxy sets proxy-url to f if the url doesn't have a host, so in ```https-tunnel?``` it's never an empty url: either f, or a url with a host.

- If there is no port but there is a protocol:
  - for "http://localhost" (ie the http protocol and no port), then port 80 is used (instead of the more standard 8080 or 3128, I'm not sure what do to in this case..).
  - On linux, for T{ url { protocol "http-alt" } { host "localhost" } }, port 8080 is used
    - Note: url parsing doesn't work on "http-alt://localhost/" >url or "http-alt://localhost" >url, they interpret it as the whole path..
  - Note: On linux, there is no entry for port 3128 in /etc/services, so you can't do something like  T{ url { protocol "squid" } { host "localhost" } }
- If there is no protocol but there is a port:
  - if the user set a string (for example by using the environnement variables where you can't use factor's TUPLE litteral syntax), "localhost:3128" >url parses as T{ url { path "localhost:3128" } } so there is no host and it doesnt use the proxy (so that's not very user friendly..but because of >url?). In addition to checking for a host in proxy-url, maybe we should ensure that the path is empty ?
  - if the user set a url, like URL" //localhost:3128/" or  T{ url { host "localhost" } { port 3128 } }, it "works". That's because we always do HTTP proxy, we don't look at the protocol from the proxy-url (except for guessing a port if missing). If we plan to support more proxy types (ie SOCKS proxy), then we can start doing something based on the protocol of proxy-url..
- if there is no protocol and no port, it throws an obscure exception deep inside the call stack.  ?

Okay so that was long... In the end, what matters is what errors should we send as a friendly message, what errors should we throw as the resulting exception (or even surprising behavior)

basis/http/client/client.factor
jonenst: Note that at first I was think that "www.google.com" should not match "google.com", only ".google.com" but other tools make it match.
So the difference between ".google.com" and "google.com" is only "google.com" which doesn't match ".google.com"

jonenst:
I answered to the comments but the answers are hidden behind the "outdated diff" fold. Let me know what you think

basis/http/client/client-docs.factor
```diff
@@ -281,6 +281,14 @@ $nl
     "http.client.encoding"
     "http.client.errors"
 }
+"For authentication, only Basic Access Authentication is implemented, using the username/password from the target or proxy url. Alternatively, the " { $link set-basic-auth } " or " { $link set-proxy-basic-auth } " words can be called on the " { $link request } " object."
+$nl
+"The http client can use an HTTP proxy transparently, by using the " { $link "http.proxy-variables" } ". Additionally, the proxy variables can be ignored by setting the " { $slot "proxy-url" } " slot of each " { $link request } " manually:"
+{ $list
+    { "Setting " { $slot "proxy-url" } " to " { $link f } " prevents http.client from using a proxy." }
```
mrjbq7: So ``f`` means override the ``"no.proxy"`` variable?

jonenst: Yes, with f the proxy will never be used

basis/http/client/client.factor
mrjbq7: Should ``proxy-url`` be a string?  Since we only want the proxy to use host and port?

jonenst: It uses username and password. Later versions might also use the protocol..
So using urls is nice?
Le 8 mars 2016 16:57, "John Benediktsson" <notifications@github.com> a
écrit :

jonenst:
Also, I saw recent ip parser integration, this can be used for proper support of IP addresses in the no_proxy list. Will look into it

jonenst:
Should ip-parser understand things like "127.0.0.0/8" ? "2001:db8::/48" ? Or do we want that in another vocab ?

mrjbq7:
My goal with ip-parser was to have it support what ping supports.

```
$ ping 127.0.0.0/8
ping: cannot resolve 127.0.0.0/8: Unknown host
```

bpollack:
Plus, `127.0.0.0/8` isn't an address; it's a netmask. Having words for that might be great, but my two cents is that I don't think `ip-parser` should be handling that.

jonenst:
Rebased on current master, where io.sockets.secure is no longer an optional dependency. This removes the commit about "opt-send-secure-handshake".

basis/http/http.factor
mrjbq7: I think we should default ``proxy-url`` to ``f`` (mostly since it's never used, and also because I think setting an "invalid" proxy should be a ``invalid-proxy`` error when you try and make a request)?

basis/http/client/client.factor
```diff
@@ -111,15 +143,74 @@ SYMBOL: redirects
...
+: no-proxy-match? ( host-path no-proxy-path -- ? )
+    dup first empty? [ [ rest ] bi@ ] when
+    [ drop f ] [ tail? ] if-empty ;
+
+: get-no-proxy-list ( -- list )
+    "no_proxy" get
+    [ "no_proxy" os-env ] unless*
+    [ "NO_PROXY" os-env ] unless* ;
+
+: no-proxy? ( request -- ? )
```
mrjbq7: Should rewrite this so the split only happens with a ``no-proxy-list``.

Something like this (untested) maybe:

```factor
: no-proxy? ( request -- ? )
    get-no-proxy-list [
       [ url>> host>> "." split ] dip "," split [ "." split no-proxy-match? ] with any?
    ] [ f ] if* ;
```

basis/http/http.factor
jonenst: You want f in ```proxy-url``` but an empty url with an empty query in ```url``` ? Or both to f ?

Also, this was a way to have a proxy-url fill in missing parts, so that for example you could set the credentials globally and the host/port for each request (or any combination). And it works the same for entirely setting the proxy globally. Is this feature useful ?

jonenst: Currently, if after filling the missing parts from the variable, the proxy doesn't have a host, then it means no proxy. It could be that only the empty url means no proxy, and that any non empty url without a host is an error (and with a host it uses this as the proxy).

basis/http/client/client.factor
mrjbq7: So ``proxy-url`` overrides variables in the ``PROXY_URL``?  Is that useful?

mrjbq7: I guess I would have thought you use either ``proxy-url>>`` (if set) or ``PROXY_URL`` (if set) or no proxy.

mrjbq7:
Okay I merged this because it's awesome, but I think we should consider making it simpler with ``proxy-url`` by using the request ``proxy-url`` (if set) or ``PROXY_URL`` environment variable (if set) or not using a proxy.

jonenst:
I added some commits to throw errors when the proxy is invalid, and to workaround "localhost:3128" >url parsing as T{ url { path "localhost:3128 } } (same workaround used on the target: ```request-url```) 
https://github.com/jonenst/factor/tree/proxy

Do you want another PR ?

And yeah, maybe the api could be simpler.

mrjbq7:
Merged the patches!

PR #1532, opened 2016-02-23
openssl: only set RSA keys if required to do so

bpollack:
Modern OpenSSL and LibreSSL both do not require SSL_CTX_set_tmp_rsa to be called unless SSL_CTX_need_tmp_rsa returns true, and LibreSSL and OpenSSL compiled with deprecation warnings both will fail if this happens. This commit resolves that.

With this change, it becomes possible to use LibreSSL in place of OpenSSL with Factor.


erg:
Merged. Now we can comment out the RSA entirely in another commit someday!

PR #1537, opened 2016-03-02
docs, smpt-gmail: fix for people not using 2FA + formatting

jonenst:
Docs fix for smtp-gmail http://docs.factorcode.org/content/article-smtp-gmail.html
tested with 2FA on and off


mrjbq7:
Merged, thanks!!

PR #1538, opened 2016-03-02
Add SNI support to Factor

bpollack:
Untested on Windows, but I can guarantee it'll break there because I changed stack effects. That said, I'm unclear if it ever worked, so maybe we don't care.

I hate OpenSSL and it needs to die in a fire.

### DIFF
--- a/basis/io/sockets/secure/secure-docs.factor
+++ b/basis/io/sockets/secure/secure-docs.factor
@@ -1,4 +1,4 @@
-USING: io help.markup help.syntax calendar quotations io.sockets ;
+USING: io help.markup help.syntax calendar quotations strings io.sockets ;
 IN: io.sockets.secure
 
 HELP: secure-socket-timeout
@@ -74,7 +74,7 @@ HELP: secure
 { $class-description "The class of secure socket addresses." } ;
 
 HELP: <secure>
-{ $values { "addrspec" "an address specifier" } { "secure" secure } }
+{ $values { "addrspec" "an address specifier" } { "hostname" { $maybe string } } { "secure" secure } }
 { $description "Creates a new secure socket address, which can then be passed to " { $link <client> } " or " { $link <server> } "." } ;
 
 ARTICLE: "ssl-addresses" "Secure socket addresses"
--- a/basis/io/sockets/secure/secure.factor
+++ b/basis/io/sockets/secure/secure.factor
@@ -44,7 +44,9 @@ HOOK: <secure-context> secure-socket-backend ( config -- context )
         with-disposal
     ] with-scope ; inline
 
-TUPLE: secure { addrspec read-only } ;
+TUPLE: secure
+    { addrspec read-only }
+    { hostname read-only } ;
 
 C: <secure> secure
 
@@ -53,7 +55,8 @@ M: secure present addrspec>> present " (secure)" append ;
 CONSULT: inet secure addrspec>> ;
 
 M: secure resolve-host ( secure -- seq )
-    addrspec>> resolve-host [ <secure> ] map ;
+    [ addrspec>> resolve-host ] [ hostname>> ] bi
+    [ <secure> ] curry map ;
 
 HOOK: check-certificate secure-socket-backend ( host handle -- )
 
--- a/basis/io/sockets/secure/unix/unix.factor
+++ b/basis/io/sockets/secure/unix/unix.factor
@@ -14,10 +14,10 @@ M: ssl-handle handle-fd file>> handle-fd ;
 
 M: unix socket-handle fd>> ;
 
-M: secure ((client)) ( addrspec -- handle )
-    addrspec>> ((client)) <ssl-socket> ;
+M: secure ((client)) ( secure -- handle )
+    [ addrspec>> ((client)) ] [ hostname>> ] bi <ssl-socket> ;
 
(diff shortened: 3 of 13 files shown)


bpollack:
Hold on this, it's super-close, but I'm doing a quick cleanup.

bpollack:
Okay, theoretically good now.

bpollack:
@erg This now supports `http-client` directly.

basis/io/sockets/secure/secure-docs.factor
```diff
@@ -74,7 +74,7 @@ HELP: secure
 { $class-description "The class of secure socket addresses." } ;
 
 HELP: <secure>
-{ $values { "addrspec" "an address specifier" } { "secure" secure } }
+{ $values { "addrspec" "an address specifier" } { "host" "a string or " { $link f } } { "secure" secure } }
```
mrjbq7: We have help markup ``{ $maybe string }`` for this case:

```factor
IN: scratchpad { $maybe string } print-element
a string or f
```

basis/io/sockets/secure/secure.factor
```diff
@@ -44,16 +44,20 @@ HOOK: <secure-context> secure-socket-backend ( config -- context )
         with-disposal
     ] with-scope ; inline
 
-TUPLE: secure { addrspec read-only } ;
+TUPLE: secure
+    { addrspec read-only }
+    { hostname read-only } ;
 
-C: <secure> secure
+: <secure> ( addrspec host -- secure )
```
mrjbq7: Why can't this just be ``C: <secure> secure`` ?

basis/io/sockets/secure/secure.factor
```diff
@@ -44,16 +44,20 @@ HOOK: <secure-context> secure-socket-backend ( config -- context )
...
+    secure boa ;
 
 M: secure present addrspec>> present " (secure)" append ;
 
 CONSULT: inet secure addrspec>> ;
 
 M: secure resolve-host ( secure -- seq )
-    addrspec>> resolve-host [ <secure> ] map ;
+    [ hostname>> ] [ addrspec>> ] bi
```
mrjbq7: Better without the swaps?
```factor
[ addrspec>> resolve-host ] [ hostname>> ] bi
[ <secure> ] curry map
```

basis/io/sockets/secure/unix/unix.factor
```diff
@@ -14,10 +14,10 @@ M: ssl-handle handle-fd file>> handle-fd ;
 
 M: unix socket-handle fd>> ;
 
-M: secure ((client)) ( addrspec -- handle )
-    addrspec>> ((client)) <ssl-socket> ;
+M: secure ((client)) ( secure -- handle )
+    [ hostname>> ] [ addrspec>> ] bi ((client)) <ssl-socket> ;
```
mrjbq7: It's weird that some words are ``( hostname addrspec -- )`` and others are ``( addrspec hostname -- )``...?

mrjbq7:
Some minor comments, if you want to look it over and then tell me when ready to merge.

basis/io/sockets/secure/secure-docs.factor
bpollack: Fixed in a second.

basis/io/sockets/secure/secure.factor
bpollack: It's a leftover from when I (temporarily) introduced a `<secure-host>` variant to aid in transitioning over to requiring the hostname everywhere. Fixed.

basis/io/sockets/secure/secure.factor
bpollack: Good point, fixed.

basis/io/sockets/secure/unix/unix.factor
bpollack: Good catch. I realized halfway through that `( addrspec hostname -- ... )` is almost always the better stack ordering, but clearly didn't patch everywhere. Fixed.

bpollack:
@mrjbq7 Thanks for the suggestions! I believe I have now implemented all of them, and we should be good to merge. If you want to test BTW, sni.velox.ch provides a quick test (compare that v. alice.sni.velox.ch or bob.sni.velox.ch).

mrjbq7:
Great! Should we add a test case using that server?

bpollack:
Sure. You want me to do that at a fairly raw level directly in `io.sockets.secure`, or put the actual test in `urls.secure` (which probably isn't where it belongs, but is a much more convenient place to stick the testing code).

mrjbq7:
Either is fine, simpler is better :-)

jonenst:
Do we have other tests that require an internet access and depend on public services ? Maybe add some kind of mechanism like platforms.txt to disable them ?

bpollack:
@jonenst Well, the ones in `http.client` already do, and that's also the most convenient place to stick the SNI test, so that's what I did. @mrjbq7, if that's acceptable, I think that'll do it.

erg:
We have a test that logs into a gmail account and moves folders around and
deletes them.

It would be nice to split up the monolithic tests concept into a super fast
test suite vs tests that require auth/network vs tests that take forever
(deploy).

mrjbq7:
Merged, thanks!

PR #1539, opened 2016-03-03
Allow specifying vocabulary roots on the command line

### LINKED ISSUE #1498: Allow new roots to be specified on the command-line
We have the ``.factor-roots`` file and the ``FACTOR_ROOTS`` environment variable, but there is a use-case that would be nice to have roots specified on the command-line, something like:

```
factor -roots="~/foo:~/baz" blah.factor
```

See, for example the discussion here:

https://github.com/kanaka/mal/pull/98#issuecomment-152253257

bpollack:
Fixes #1498

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ Common arguments:
         -run=ui.tools    run Factor development UI
     -e=<code>        evaluate <code>
     -no-user-init    suppress loading of .factor-rc
+    -roots=<paths>   a list of path-delimited extra vocab roots
 
 Enter
     "command-line" help
--- a/basis/command-line/command-line-tests.factor
+++ b/basis/command-line/command-line-tests.factor
@@ -30,3 +30,7 @@ IN: command-line
     { "factor" "-foo" "a" "b" "c" } parse-command-line
     executable get script get command-line get
 ] unit-test
+
+{ "a:b:c" } [ { "factor" "-roots=a:b:c" } parse-command-line
+    "roots" get-global
+] unit-test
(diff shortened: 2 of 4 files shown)


mrjbq7:
Probably should use a different split character on Windows:

```
os windows? ";" ":" ?
```

mrjbq7:
Also, if we merge this, it should be documented in ``basis/command-line/startup/startup.factor`` in ``cli-usage``.

mrjbq7:
And in ``README.md`` maybe since that copies the ``cli-usage``.

erg:
Needs unit tests, docs, and maybe we should make it ``-root=foo -root=bar`` instead of using a separator so it's cross-platform, as john pointed out?

mrjbq7:
Can't do ``-root=foo -root=bar`` because command-line doesn't work that way, it just sets the key/values so the last overwrites the first.

bpollack:
I'll add unit tests, docs, and make it use `;` on Windows and `:` on everything else, but @mrjbq7 just pointed out why I went with the splitting in the first place. Obviously that's still doable, but that version requires a VM change.

erg:
We control the implementation of command-line so we could make it work however we want!

But yeah that's more work.

bpollack:
Oh, I thought that the command-line parsing was in the VM for some reason. In that case, yeah, it's more doable. Strong preferences? Do we really want to special-case `-root`?

mrjbq7:
I'd prefer not to special case, although you could "generalize" it by making any conflicting keys append in some standard way, but that seems a little hokey.

So, I don't mind the way you're doing it.

bpollack:
Now with docs and unit tests. Anything else blocking a merge?

erg:
Merged! Thanks!!!!

PR #1540, opened 2016-03-03
Do not attempt to set an SNI hostname on a server

bpollack:
@erg, this fixes the issue you were hitting. Sorry about that.


mrjbq7:
Merged!

PR #1541, opened 2016-03-04
Use proper SI prefixes in memory flag docs

bpollack:


erg:
Thanks!

bjourne:
Is KB the 1000 or 1024 one though? The stack sizes are actually specified in multiples of 1024 and 1048576 for megabyte counts.

bpollack:
@bjourne Sigh. You're correct. I initially thought that the actual fix was to convert to kibibytes/mebibytes (i.e. 1024), but my read of `vm/factor.cpp:21-22` and `vm/factor.cpp:69-71` was that we were using these values verbatim.

…which are, *there*, but then we're shifting by 20 bits at `vm/factor.cpp:134-137`, which I missed.

Fix to the fix coming. Thanks for catching this.

mrjbq7:
:-)

PR #1542, opened 2016-03-05
Make <inet4> and <ipv4> use ip-parser:parse-ipv4

### LINKED ISSUE #1209: Use ip-parser in <ipv4>
It might be nice to support IPV4 addresses specified in ``ip-parser:parse-ipv4`` format:

```
IN: scratchpad USE: ip-parser 
IN: scratchpad "2130706433" parse-ipv4 .
"127.0.0.1"

IN: scratchpad USE: io.sockets
IN: scratchpad "2130706433" <ipv4>
Invalid IPv4 address
string "2130706433"
reason T{ malformed-ipv4 f ~array~ }

Type :help for debugging help.
```

Probably, if it exists, we should provide similar support for ``<ipv6>``.

bpollack:
As far as I can tell, we're already handling the full IPv6 spec properly (i.e., the alt forms from ipv4 do not have ipv6 equivalents), so I'm not clear that it means anything to do the same for IPv6 addresses. That said, I'm really happy to do that if I'm missing something.

Fixes #1209


mrjbq7:
Merged, thanks!!

PR #1543, opened 2016-03-05
Import a few extra JSON tests from Jansson

### LINKED ISSUE #1189: json writer broken on some objects?

bpollack:
We already had almost all of these represented, but I added a couple extra corner cases that we were missing.

Fixes #1189


mrjbq7:
Merged, thanks!!!

Are you interested in commit rights to the factor repo?

bpollack:
@mrjbq7 Sure!

PR #1552, opened 2016-03-09
Make VSC work on OS X again

bpollack:
`-r` isn't an argument to `open`, and `-n` seems exactly the opposite of what we want in most circumstances.


mrjbq7:
Merged, thanks!!

PR #1553, opened 2016-03-11
Switch Chloe, and lots of our other HTML assets, from XHTML to HTML5

bpollack:
This converts Chloe fully to generate HTML5, not XHTML. The changes are fairly minor; mostly it's adding `<!DOCTYPE html>` to a bunch of the built-in apps, and killing the existing XML prolog generation.

### DIFF
--- a/basis/html/templates/chloe/chloe-tests.factor
+++ b/basis/html/templates/chloe/chloe-tests.factor
@@ -10,7 +10,7 @@ IN: html.templates.chloe.tests
 
 : run-template ( quot -- string )
     with-string-writer [ "\r\n\t" member? ] reject
-    "?>" split1 nip ; inline
+    [ CHAR: \s = ] trim ; inline
 
 : test-template ( name -- template )
     "vocab:html/templates/chloe/test/"
--- a/basis/html/templates/chloe/compiler/compiler.factor
+++ b/basis/html/templates/chloe/compiler/compiler.factor
@@ -176,9 +176,7 @@ ERROR: unknown-chloe-tag tag ;
 
 : compile-prologue ( xml -- )
     [
-        [ prolog>> [ write-xml ] [code-with] ]
-        [ before>> compile-chunk ]
-        bi
+        before>> compile-chunk
     ] compile-quot
     [ if-not-nested ] [code] ;
 
--- a/basis/html/html.factor
+++ b/basis/html/html.factor
@@ -1,13 +1,17 @@
 ! Copyright (C) 2004, 2009 Chris Double, Daniel Ehrenberg,
 ! Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: kernel xml.data xml.writer xml.syntax urls.encoding ;
+USING: accessors kernel xml.data xml.writer xml.syntax 
+urls.encoding ;
 IN: html
+
+TUPLE: empty-prolog < prolog ;
+M: empty-prolog write-xml drop ;
+: <empty-prolog> ( -- prolog ) \ empty-prolog new ;
 
 : simple-page ( title head body -- xml )
     <XML
-        <?xml version="1.0"?>
-        <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
+        <!DOCTYPE html>
         <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
             <head>
                 <title><-></title>
@@ -15,7 +19,7 @@ IN: html
             </head>
             <body><-></body>
         </html>
-    XML> ;
+    XML> <empty-prolog> >>prolog ;
 
 : render-error ( message -- xml )
     [XML <span class="error"><-></span> XML] ;
--- a/basis/xml/writer/writer.factor
+++ b/basis/xml/writer/writer.factor
@@ -142,9 +142,10 @@ M: public-id write-xml
 
 M: doctype-decl write-xml
     ?indent "<!DOCTYPE " write
-    [ name>> write bl ]
-    [ external-id>> [ write-xml bl ] when* ]
-    [ internal-subset>> write-internal-subset ">" write ] tri ;
+    [ name>> write ]
+    [ external-id>> [ bl write-xml ] when* ]
+    [ internal-subset>> [ bl write-internal-subset ] when* ] tri
+    ">" write ;
 
 M: directive write-xml
     "<!" write text>> write CHAR: > write1 nl ;
--- a/extra/webapps/calculator/calculator.xml
+++ b/extra/webapps/calculator/calculator.xml
@@ -1,7 +1,6 @@
 <?xml version='1.0' ?>
-
+<!DOCTYPE html>
 <t:chloe xmlns:t="http://factorcode.org/chloe/1.0">
-
 <html>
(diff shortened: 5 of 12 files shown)


basis/html/templates/chloe/compiler/compiler.factor
```diff
@@ -74,17 +79,38 @@ SYMBOL: string-context?
 
 ERROR: tag-not-allowed-here ;
 
+CONSTANT: self-closing-tags {
```
mrjbq7: Maybe make this a hash-set? HS{ }

basis/html/templates/chloe/compiler/compiler.factor
```diff
@@ -74,17 +79,38 @@ SYMBOL: string-context?
...
+
 : check-tag ( -- )
     string-context? get [ tag-not-allowed-here ] when ;
 
-: compile-tag ( tag -- )
-    check-tag
-    {
-        [ main>> tag-stack get push ]
+: (compile-tag) ( tag -- )
+    dup name>string >lower self-closing-tags in?
```
mrjbq7: Since you use "in?", otherwise a "member?" check?

basis/html/templates/chloe/compiler/compiler.factor
```diff
@@ -74,17 +79,38 @@ SYMBOL: string-context?
...
+    [ compile-self-closing-tag ]
```
mrjbq7: Stylistically, we had been thinking that if an if block expands, then expand both:

```
eh? [
    short-stuff
] [
    lots of stuff
    that spans multiple lines
] if
```

basis/html/templates/chloe/compiler/compiler.factor
bpollack: Seemed like a bad complexity trade-off to go that route; the list is only consulted at template compile time, which is relatively rare, so I wasn't convinced that the time and memory to generate the hash set was going to be made up by savings on compile time later. That said, I don't feel strongly about this. Just let me know.

basis/html/templates/chloe/compiler/compiler.factor
```diff
@@ -149,9 +175,8 @@ ERROR: unknown-chloe-tag tag ;
 
 : compile-prologue ( xml -- )
     [
-        [ prolog>> [ write-xml ] [code-with] ]
-        [ before>> compile-chunk ]
-        bi
+        "<!DOCTYPE html>\n" [write]
```
mrjbq7: So no more xhtml? I suppose that's because it's all deprecated in favor of the "future is now"?

basis/html/templates/chloe/compiler/compiler.factor
bpollack: Ah, got it. Will unify this around `member?` for the moment, pending your thoughts on flipping to a hash set.

basis/html/templates/chloe/compiler/compiler.factor
bpollack: XHTML wasn't ever terribly well supported, and actually has a negative impact in some browsers (e.g. it'll push IE into one of its quirks modes), so I *think* that fully drinking the HTML5 Koolaid is okay at this point. Certainly XHTML is no longer being developed and HTML5 is.

If there are reasons I'm not aware of why we want to keep *Chloe* XHTML generation alive, let me know; the other option I debated was revving Chloe to 1.1 and only doing the `DOCTYPE` swizzle if that's set. It's not much more code; I just didn't see any value.

basis/html/templates/chloe/compiler/compiler.factor
bpollack: Bah, thanks. Did this as a break from some Python and it shows. Fixing.

basis/html/templates/chloe/compiler/compiler.factor
mrjbq7: Ahh, you know what, I forgot that ``in?`` just does a membership check on sequences:

```factor
M: sequence in? member? ; inline
```

So it's probably okay :-)

Or switch it, up to you.

basis/html/templates/chloe/compiler/compiler.factor
mrjbq7: I love the future! Long live the future!

basis/html/templates/chloe/compiler/compiler.factor
bpollack: It actually seems weird to have both words, honestly. I knew about `in?` calling `member?`, but figured that you were aiming not to imply that I was using a literal set. But I'm not clear if we even ever care.

mrjbq7: I think the original idea was two-fold.  1) sets were created after sequences, so duplication.  2) implied complexity of member? is O(n), but in? could be faster.

I agree, it would better to unify some of the container words and implement cursors and other improvements, and hope to do some of that as part of the larger work for 0.99, but have wanted to get a good 0.98 release first.

bpollack:
@mrjbq7 Okay, cut over from `in?` to `member?` and redid the formatting to what I believe is the right Factor style. If things look good, I'll land this.

basis/html/templates/chloe/chloe-tests.factor
```diff
@@ -10,7 +10,7 @@ IN: html.templates.chloe.tests
 
 : run-template ( quot -- string )
     with-string-writer [ "\r\n\t" member? ] reject
-    "?>" split1 nip ; inline
+    "<!DOCTYPE html>" split1 nip ; inline
```
mrjbq7: A few of our templates seem to have DOCTYPES that are a little different:

```
$ find . -iname "*.xml" | xargs grep DOCTYPE
./extra/webapps/help/search.xml:<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
./extra/webapps/mason/download-package.xml:<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
./extra/webapps/mason/download-release.xml:<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
./extra/websites/concatenative/page.xml:<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
```

They either have to get fixed, or change the splitting logic...?

mrjbq7: Oh, this is only for the tests.....

mrjbq7:
Looks good except if you could see if the DOCTYPE thing needs to be fixed in a few of the existing templates...

basis/html/templates/chloe/chloe-tests.factor
bpollack: This actually makes me want to rethink the second patch. I think that I should actually just drop the prolog entirely, leave it up to the template to include `<!DOCTYPE html>` if they want to, and then modify these templates to explicitly have `<!DOCTYPE html>` in them.

Otherwise, there's a really weird situation wherein these apps, which are voluntarily not using `boilerplate`, *are* in fact getting a boilerplate—the `<!DOCTYPE>`—which they may legitimately not want.

I'll rev this to have that as a solution and we can check in to make sure that looks okay.

bpollack:
@mrjbq7 Pretty heavy (though simple) redo of the second part of this PR.

bpollack:
Landed.

Issue #1572, opened 2016-03-19
Split inputs and outputs in word help


mrjbq7:
This is a good suggestion by @nicolas-p:

https://github.com/nicolas-p/skov/issues/1#issuecomment-198567107

mrjbq7:
It probably requires different help syntax, or possibly can get it by setting the ``current-topic`` when you call ``print-topic`` and then using the stack effect of the current topic (if it's a word) to split the ``$values`` into separate groups for inputs and outputs.

nicolas-p:
I will post the images here as well.

The current display:

![capture d ecran 2016-03-18 a 23 19 12](https://cloud.githubusercontent.com/assets/11431580/13900354/b56fffec-ee04-11e5-99e1-94543c05b783.png)

My proposition:

![inputs-outputs](https://cloud.githubusercontent.com/assets/11431580/13900351/9998727c-ee04-11e5-9aaf-6bddbd25c82c.png)

mrjbq7:
This is one way to do it, by mutating the help content when it is set on a word:

```diff
diff --git a/basis/help/help.factor b/basis/help/help.factor
index 2139890..453457f 100644
--- a/basis/help/help.factor
+++ b/basis/help/help.factor
@@ -179,7 +179,21 @@ help-hook [ [ print-topic ] ] initialize
 : remove-word-help ( word -- )
     f "help" set-word-prop ;
 
+: inputs-and-outputs ( content word -- content word )
+    over [ ?first \ $values = ] find drop [
+        [ cut unclip rest ] curry dip
+        [
+            stack-effect [ in>> ] [ out>> ] bi
+            [ [ dup pair? [ first ] when effect>string ] map ] bi@
+            [ [ [ ?first ] dip member? ] curry filter ] bi-curry@
+            \ $inputs \ $outputs
+            [ [ prefix ] curry compose ] bi-curry@ bi* bi
+            2array glue
+        ] keep
+    ] when* ;
+
 : set-word-help ( content word -- )
+    inputs-and-outputs
     [ remove-word-help ] keep
     [ swap "help" set-word-prop ] keep
     xref-article ;

diff --git a/basis/help/markup/markup.factor b/basis/help/markup/markup.factor
index 1b3324b..27add1c 100644
--- a/basis/help/markup/markup.factor
+++ b/basis/help/markup/markup.factor
@@ -396,6 +396,12 @@ M: f ($instance) ($link) ;
     unclip \ $snippet swap present 2array
[...]

PR #1576, opened 2016-03-20
fix confusing typo in "?" docs

catb0t:
I was reading the docs about `?` to refresh my memory, and it took me a second to understand what was meant by `one two input objects`:

onetwo


mrjbq7:
Another might be to do this:

``{ "true/false" { { $snippet "true" } " or " { $snippet "false" } } }``

Is that more clear?

catb0t:
Actually, yes, that's likely more clear. It'll become even more clear when https://github.com/factor/factor/issues/1572 is implemented.

mrjbq7:
Fixed in 4d2baf5d50fbea0e62acf38e950724b561e0e2ad.

PR #1609, opened 2016-04-06
directory-file-paths and directory-tree-file-paths

nicolas-p:
Two functions that give the complete paths of all files in a directory instead of giving just their names.


mrjbq7:
Maybe we should change the words that don't switch into ``current-directory`` to do what you said rather than having a separate word for it.

nicolas-p:
You're right, this would probably be better.

mrjbq7:
Apparently in ``io.directories.search``, we have ``qualified-directory-entries`` and ``qualified-directory-files`` that produce ``absolute-path`` versions of ``directory-entries`` and ``directory-files``. 

I kind of think we should merge them.  @erg?

erg:
Yeah, let's merge it somehow. Don't need two words that do the same thing!

bjourne:
I've tried to fix this problem a little in ba4bb45d4d34c48f08eea9dd5e9df9ba3eac0dd6 by moving the qualified words to `io.directories`. Then for the last problem, that a variant of `directory-tree-files` that produces absolute paths is needed, I think that the word can just be removed and `recursive-directory-files` used instead. Afaict, the directory tree walking implemented in `io.directories.search` is strictly superior and has more features than the one in `io.directories.hierarchy`.

mrjbq7:
Hmm, I agree we could do some merging/deleting between ``io.directories.search`` and ``io.directories.hierarchy``, but with the following comments:

1) We need a good place for ``copy-tree``, ``delete-tree``, etc.  It would be nice to put in ``io.directories``, but we would have a dependency problem on ``io.directories.search``, which maybe could be also moved to ``io.directories``.

2) The ``directory-tree-files`` is relative to the path being search, not qualified, but ``recursive-directory-files`` is qualified.  That shouldn't matter that much, but does in the case of users like ``vocabs.metadata.resources`` and ``codebook``, because that behavior will be a little different for them.

bjourne:
I guess we can make qualified and unqualified variants of the word? Then we would end up with: `recursive-directory-files`, `recursive-directory-entries`, `recursive-qualified-directory-files` and `recursive-qualified-directory-entries`

erg:
A unix find clone would be awesome. With full support for all the command line options...

PR #1616, opened 2016-04-17
Atom package: update paths

ForNeVeR:
I've noticed that Atom package installation paths are a bit off and decided to fix them.

Also I think that `language-factor` name will be more conventional than `languages-factor` (because all Atom language packages tend to be named as `language-*`), so I've decided to change the name it too.

Please note that the package name currently is just a recommendation, so nothing will go wrong if some guys have it installed under the old name.

I have a further interest to publish that Atom package to https://atom.io/packages and support it in an external repository, so it'd become more accessible to the Atom users (_the_ standard way of installing and updating the package is to download it from https://atom.io/packages central package repository); what would you guys say?

### DIFF
--- a/misc/atom/README.md
+++ b/misc/atom/README.md
@@ -4,11 +4,11 @@ This is a syntax highlighting package for Factor.
 
 You can install this by copying this folder to:
 
-  cp -r /path/to/factor/mis/atom ~/.atom/languages-factor
+  cp -r /path/to/factor/misc/atom ~/.atom/packages/language-factor
 
 Or symlinking it:
 
-  ln -sf /path/to/factor/misc/atom ~/.atom/languages-factor
+  ln -sf /path/to/factor/misc/atom ~/.atom/packages/language-factor
 
 This was initially generated from the Textmate bundle:
 


mrjbq7:
Merged, thanks!

Regarding the ``atom.io`` publishing, what do you require to do this?  If we had this as a separate git repository under ``github.com/factor/atom-factor`` or something (what's the general naming convention for these?) would that work for you?

catb0t:
For instance, the official [language-python](https://github.com/atom/language-python) package is `atom/language-python`, and the official Scala one is `jroesch/language-scala`, so ideally it ill be `factor/language-factor`.

mrjbq7:
Yeah that's kinda a tough name since it says nothing about it being related to atom or not being factor itself. Little bit confusing.

ForNeVeR:
Guys, I think that the best name for the repository would be `factor/atom-language-factor`. It is not necessary for it to match the actual Atom package name, so I think it will be okay.

mrjbq7:
I pushed a copy of the ``misc/atom`` directory to a separate git repository for you:

https://github.com/factor/atom-language-factor

For now I'll keep them in sync, but at some point it might be good to split our ``misc`` directory into a few separate repositories for each editor.

ForNeVeR:
Thank you! I'll make some changes to that repository, publish them to atom.io and make a corresponding PR.

PR #1621, opened 2016-04-30
Please consider this request

AlexIljin:
Two documentation updates, and an update to the color-table sample.

### DIFF
--- a/basis/colors/colors-docs.factor
+++ b/basis/colors/colors-docs.factor
@@ -14,7 +14,7 @@ HELP: >rgba
 ARTICLE: "colors.protocol" "Color protocol"
 "Abstract superclass for colors:"
 { $subsections color }
-"All color objects must are required to implement a method on the " { $link >rgba } " generic word."
+"All color objects are required to implement a method on the " { $link >rgba } " generic word."
 $nl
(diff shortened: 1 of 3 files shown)


mrjbq7:
The color table already colors the text by the color, is the extra column important?

AlexIljin:
Actually, that was extremely practical. I had to find color names for some physical objects (puzzle pieces), so I held them to the monitor and tried to find the best match in the table. It was very difficult to judge color from the pixel-thin text lines, I needed a solid block of the color. Otherwise the white background ruins the perception. It turned out, there was no way to set the background color in the table cell, so that was my solution.

Feel tree to omit that patch, I can just have it for myself.

AlexIljin:
BTW, I would avoid coloring the text altogether, it makes some lines absolutely unreadable.

mrjbq7:
Looks like the ``ui.gadgets.table`` interface assigns a single color to a row, which is unfortunate.

AlexIljin:
Yes. If there was a way to assign the background color via the renderer object, then the color-table example would still be the best place to showcase the function (with the extra column of different background colors, that is). I'm just not that good in Factor to offer such a change as a patch, I can only formulate it as a suggestion for you to consider.

mrjbq7:
Merged!

Issue #1470, opened 2015-09-16
Buggy file-readable on Windows?


mrjbq7:
From the mailing list:

https://www.mail-archive.com/factor-talk@lists.sourceforge.net/msg07337.html

>  I've got another question about the file-readable? word. The documentation says that it "returns whether the file specified by path is readable by the current process." Here's my experiment:
>
>  "C:\\boot.ini" file-readable? .
> -> t
>  "C:\\boot.ini" sha-256 checksum-file hex-string .
> -> Exception "Access is denied."
>
>  Being a non-root user in Windows, I have no permission to read C:\boot.ini, but file-readable? returns true.

AlexIljin:
file-readable? checks if it can create the file-info tuple and returns true on success. And indeed, the metadata about boot.ini in the tuple is correct. But that doesn't mean that the file contents can be accessed by the current user. It seems like either the documentation is incorrect, or the implementation needs to be updated.
Tested on WinXP SP3, NTFS, under non-admin user account.

erg:
First of all, ``"C:\boot.ini" file-readable? .`` is escaping the ``b`` as a backspace.

It should be:  ``"C:\\boot.ini" file-readable? .`` or ``"C:/boot.ini" file-readable? .``

Next, ``boot.ini`` is not present on Windows 10, only pre-Windows 7.

We could check the ``attributes>>`` slot of the ``file-info`` instead of if it allows ``file-info`` or not. Even then, ``dir /a:h c:`` shows ``pagefile.sys`` and ``swapfile.sys`` but I can't checksum them because roughly "another program has opened these files".

erg:
Should we try to actually open the file and read a byte instead?

The R property shown by ``attrib`` is all but useless.
```
c:\>attrib
A            C:\ANG0
A  SHR       C:\bootmgr
A  SH        C:\BOOTNXT
A  SHR       C:\BOOTSECT.BAK
A            C:\cmd.zip
A  SH        C:\pagefile.sys
A  SH        C:\swapfile.sys
```

AlexIljin:
R means read-only, which is a file attribute, not a permission statement. The attribute is a soft deterrent to prevent accidental file modification.

[...]

PR #1631, opened 2016-05-24
Fix file readable

AlexIljin:
This fixes issue #1470.

### DIFF
--- a/basis/io/files/info/windows/windows.factor
+++ b/basis/io/files/info/windows/windows.factor
@@ -76,15 +76,20 @@ TUPLE: windows-file-info < file-info-tuple attributes ;
         [ GetFileInformationByHandle win32-error=0/f ] keep
     ] keep CloseHandle win32-error=0/f ;
 
-: get-file-information-stat ( path -- file-info )
-    dup
+: valid-handle? ( handle -- boolean )
+    INVALID_HANDLE_VALUE = not ; inline
+
+: open-read-handle ( path -- handle/f )
     GENERIC_READ FILE_SHARE_READ f
     OPEN_EXISTING FILE_FLAG_BACKUP_SEMANTICS f
-    CreateFileW dup INVALID_HANDLE_VALUE = [
-        drop find-first-file-stat WIN32_FIND_DATA>file-info
-    ] [
+    CreateFileW [ valid-handle? ] keep f ? ;
+
+: get-file-information-stat ( path -- file-info )
+    dup open-read-handle dup [
         nip
         get-file-information BY_HANDLE_FILE_INFORMATION>file-info
+    ] [
+        drop find-first-file-stat WIN32_FIND_DATA>file-info
     ] if ;
 
 M: windows file-info ( path -- info )
--- a/basis/io/files/info/windows/windows.factor
+++ b/basis/io/files/info/windows/windows.factor
@@ -80,8 +80,8 @@ TUPLE: windows-file-info < file-info-tuple attributes ;
     INVALID_HANDLE_VALUE = not ; inline
 
 : open-read-handle ( path -- handle/f )
-    GENERIC_READ FILE_SHARE_READ f
-    OPEN_EXISTING FILE_FLAG_BACKUP_SEMANTICS f
+    GENERIC_READ share-mode f
+    OPEN_EXISTING 0 CreateFile-flags f
     CreateFileW [ valid-handle? ] keep f ? ;
 
 : get-file-information-stat ( path -- file-info )
--- a/basis/io/files/info/windows/windows.factor
+++ b/basis/io/files/info/windows/windows.factor
@@ -80,6 +80,7 @@ TUPLE: windows-file-info < file-info-tuple attributes ;
     INVALID_HANDLE_VALUE = not ; inline
 
 : open-read-handle ( path -- handle/f )
+    ! Parameters of CreateFileW here should match those in open-read.
     GENERIC_READ share-mode f
     OPEN_EXISTING 0 CreateFile-flags f
     CreateFileW [ valid-handle? ] keep f ? ;
--- a/basis/io/files/info/windows/windows.factor
+++ b/basis/io/files/info/windows/windows.factor
@@ -244,6 +244,9 @@ M: windows file-systems ( -- array )
 : set-file-write-time ( path timestamp -- )
     [ f f ] dip set-file-times ;
 
-M: windows file-readable? file-info >boolean ;
+M: windows file-readable?
+    normalize-path open-read-handle
+    dup [ CloseHandle win32-error=0/f ] when* >boolean ;
+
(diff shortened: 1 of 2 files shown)


mrjbq7:
This is kinda cool, but is trying to read the file the only way to test for readability on Windows?

mrjbq7:
Do we need instead something like Python's ``os.access``?

AlexIljin:
I'm not sure if it's the only way, but I think it's the simplest. At least when I did my research on this topic (months ago) nothing else turned up.
I have some Python version installed at work, so I can check how it is implemented there, tomorrow.
I'll post here when I have new info.

erg:
I am trying this on files in ``c:\``. Here's what I got:

```factor
! before your patch
"c:\\" qualified-directory-files [ file-readable? ] map-zip [ second ] reject .
{ }

! with your patch
"c:\\" qualified-directory-files [ file-readable? ] map-zip [ second ] reject .
{
    { "c:/Config.Msi" f }
    { "c:/Documents and Settings" f }
    { "c:/HaxLogs.txt" f }
    { "c:/pagefile.sys" f }
    { "c:/Recovery" f }
    { "c:/swapfile.sys" f }
    { "c:/System Volume Information" f }
}
```

However, ``c:/HaxLogs.txt`` is actually readable even though your patch says it's not:

```factor
USE: tools.hexdump
"c:\\HaxLogs.txt" binary file-contents hexdump.
Length: 93, 5dh
00000000h: 2d 2d 2d 2d 2d 2d 2d 2d 20 48 41 58 4d 20 72 65 -------- HAXM re
00000010h: 6c 65 61 73 65 20 36 2e 30 2e 31 20 2d 2d 2d 2d lease 6.0.1 ----
00000020h: 2d 2d 2d 2d 0a 0d 0a 00 54 68 69 73 20 6c 6f 67 ----....This log
00000030h: 20 63 6f 6c 6c 65 63 74 73 20 72 75 6e 6e 69 6e  collects runnin
00000040h: 67 20 73 74 61 74 75 73 20 6f 66 20 48 41 58 4d g status of HAXM
00000050h: 20 64 72 69 76 65 72 2e 0a 0a 0d 0a 00           driver......


{
        "c:/HaxLogs.txt"
        T{ windows-file-info
            { type +regular-file+ }
            { size 93 }
            { size-on-disk 4096 }
            { permissions 32 }
            { created
                T{ timestamp
                    { year 2016 }
                    { month 1 }
                    { day 28 }
                    { hour 22 }
                    { minute 28 }
                    { second 22 }
                }
            }
            { modified
                T{ timestamp
                    { year 2016 }
                    { month 5 }
                    { day 21 }
                    { hour 23 }
                    { minute 54 }
                    { second 25 }
                }
            }
            { accessed
                T{ timestamp
                    { year 2016 }
                    { month 1 }
                    { day 28 }
                    { hour 22 }
                    { minute 28 }
                    { second 22 }
                }
            }
            { attributes { +archive+ } }
        }
    }
```

Trying to read all the files:
```factor
"c:\\" qualified-directory-files [ file-readable? ] map-zip [ second ] reject keys [ '[ _ binary [ ] with-file-reader t ] [ f 2array ] recover ] map-zip ...
{
    { "c:/Config.Msi" { "Access is denied." f } }
    { "c:/Documents and Settings" { "Access is denied." f } }
    { "c:/HaxLogs.txt" t }
    {
        "c:/pagefile.sys"
        {
            "The process cannot access the file because it is being used by another process."
            f
        }
    }
    { "c:/Recovery" { "Access is denied." f } }
    {
        "c:/swapfile.sys"
        {
            "The process cannot access the file because it is being used by another process."
            f
        }
    }
    { "c:/System Volume Information" { "Access is denied." f } }
}
```


I don't have time to debug the ``HaxLogs.txt`` this second but maybe you can take a look?

erg:
Looks like ``open-read`` uses the flags ``GENERIC_READ OPEN_EXISTING`` and these need fewer restrictions than ``open-read-handle``.


```factor
: open-read ( path -- win32-file )
    GENERIC_READ OPEN_EXISTING 0 open-file 0 >>ptr ;

: open-file ( path access-mode create-mode flags -- handle )
    [
        [ share-mode default-security-attributes ] 2dip
        CreateFile-flags f CreateFile opened-file
    ] with-destructors ;


: open-read-handle ( path -- handle/f )
    GENERIC_READ FILE_SHARE_READ f OPEN_EXISTING
    FILE_FLAG_BACKUP_SEMANTICS f CreateFileW
    [ valid-handle? ] keep f ? ;
```

AlexIljin:
@erg, thank you very much, I'll definitely take a look at that.

AlexIljin:
I think I know what the problem is. I'll update the pull request.

AlexIljin:
Note to self: why is open-existing is not just an alias to open-r/w?

AlexIljin:
It would appear that the issue with flags is that `open-read-handle` specifies `FILE_SHARE_READ`, while `open-file` uses `share-mode`, which additionally includes `FILE_SHARE_WRITE` and `FILE_SHARE_DELETE`. I think @erg's `c:/HaxLogs.txt` is open for writing by another process, so the missing `FILE_SHARE_WRITE` flag causes `open-read-handle` to fail.

I'll update the pull request, probably soon. Looking at the Python sources now.

AlexIljin:
I did not find the Python sources for the `os.access` function, but I saw it in action. Using Process Monitor I saw that python.exe it tries to open the file passed to `os.access` with CreateFile, just like we do now. No additional magic functions were involved, so our implementation is solid (apart from the share mode flags fix, which I'm about to roll out).

AlexIljin:
I have updated the pull request. Please, consider.

bjourne:
I think the Python code for access is this: https://github.com/python/cpython/blob/master/Modules/posixmodule.c#L2585

It seems to only use one call to `GetFileAttributesW/A` to determine if access is allowed. It doesn't work so well though because `os.access("C:\\Windows\\System32\\config", os.R_OK)` returns True but the directory is not readable. So opening the file and reading something from it is probably the only way.

AlexIljin:
GetFileAttributes internally calls CreateFile (and subsequently CloseHandle), but instead of GENERIC_READ access it requests some sort of "Read Attributes" access, which I did not see mentioned in the MSDN CreateFile article.

Since we are trying to fix the file-readable? word, it makes sense that we actually request the GENERIC_READ, same as open-read would. In this respect our implementation would work better than Python's for our use case. We are testing the file contents access permission, while Python tests file attributes access.

For example, on my system `os.access("c:\\ntldr", os.R_OK)` returns `True`, but trying to read the file results in "Access denied" message, because I'm not under an admin user. With the patch above in Factor:
`"c:/ntldr" file-readable? .`
`f`

AlexIljin:
Thanks for finding the Python sources, by the way!

erg:
Merged. Thanks!!

AlexIljin:
@erg: don't forget to close issue #1470.

PR #1635, opened 2016-05-27
VM fixes

AlexIljin:
Two fixes. One minor, probably is a no-issue. The other one is actually a broken build under VS 2015.


AlexIljin:
Sorry, I'll fix one of the log messages real quick.

PR #1636, opened 2016-05-27
VM fixes

AlexIljin:
Two fixes. One minor, probably is a no-issue. The other one is actually a broken build under VS 2015.

### DIFF
--- a/vm/os-windows.cpp
+++ b/vm/os-windows.cpp
@@ -37,7 +37,7 @@ void factor_vm::ffi_dlclose(dll* dll) {
 
 BOOL factor_vm::windows_stat(vm_char* path) {
   BY_HANDLE_FILE_INFORMATION bhfi;
-  HANDLE h = CreateFileW(path, GENERIC_READ, FILE_SHARE_READ, NULL,
+  HANDLE h = CreateFileW(path, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL,
                          OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
 
   if (h == INVALID_HANDLE_VALUE) {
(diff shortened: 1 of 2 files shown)


bjourne:
Looks good to me, but I'm curious about the added flags `FILE_SHARE_WRITE | FILE_SHARE_DELETE`? My understanding is that those flags only affects subsequent calls to CreateFile. So they shouldn't be needed because CloseHandle is called almost immediately a few lines later?

AlexIljin:
From MSDN on FILE_SHARE_WRITE: 

> If this flag is not specified, but the file or device has been opened for write access..., the function fails.

That's why CreateFile without the flag will fail to open a file (even for read-only access) if someone else has the same file **already open** for writing.
Same thing with FILE_SHARE_DELETE.
https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858%28v=vs.85%29.aspx
I've been rereading that article several times a day for a few days now : )

bjourne:
How can I test that? If I first open the factor.image file in notepad, will the CreateFileW in windows_stat fail?

AlexIljin:
No, because Notepad doesn't keep the file open.
You simply open a file for writing in one Factor instance, then, while it is still open, try to open it for reading in another Factor instance. I'll try to come up with some code to illustrate this.

AlexIljin:
It turns out you don't even need two instances of Factor. One works fine.
```
USING: io.files.windows windows.kernel32 literals ;
"c:/Programs/Dev/Factor.git/testfile" open-write
```
You will see an alien handle of the created file on the stack.

Then this will fail (returns -1):
`"c:/Programs/Dev/Factor.git/testfile" GENERIC_READ 0 f OPEN_EXISTING 0 f CreateFile`

And this will succeed because of the extra sharing flags:
`"c:/Programs/Dev/Factor.git/testfile" GENERIC_READ flags{ FILE_SHARE_WRITE FILE_SHARE_READ } f OPEN_EXISTING 0 f CreateFile`

bjourne:
Oh right, I see. Then let's merge? @mrjbq7

erg:
Looks good. I merged this and added a unit test so we don't get a regression sometime in the future.

Thanks!

PR #1638, opened 2016-05-28
Fix 1573 (ctrl-break interrupt loops)

AlexIljin:
This is a draft implementation of the fix.


AlexIljin:
The implementation is now complete.

vm/os-windows.cpp
```diff
@@ -304,6 +304,61 @@ void handle_ctrl_c() {
   SetConsoleCtrlHandler(factor::ctrl_handler, TRUE);
 }
 
+static volatile bool stop_on_ctrl_break = false;
+static HANDLE ctrl_break_thread = nullptr;
+
```
bjourne: nullptr? Why not NULL here and in other places?

AlexIljin: I don't know. Because C++11?
typedef void *HANDLE;
http://stackoverflow.com/questions/1282295/what-exactly-is-nullptr

AlexIljin: Same story with bool-BOOL, true-TRUE and false-FALSE.

basis/listener/listener.factor
```diff
@@ -120,7 +125,7 @@ t error-summary? set-global
 
     [
         read-quot [
-            '[ datastack _ with-datastack ]
+            [ '[ datastack _ with-datastack ] with-ctrl-break ] curry
```
AlexIljin: @bjourne @erg @mrjbq7 
Am I right that this line can be simpler, like so:
'[ [ datastack _ with-datastack ] with-ctrl-break ]

vm/os-windows.cpp
bjourne: Ah I see. I never heard of nullptr before. But please use NULL for now and in later pr's we can tidy it up.

vm/os-windows.cpp
```diff
@@ -304,6 +304,61 @@ void handle_ctrl_c() {
...
+}
+
+void factor_vm::primitive_enable_ctrl_break() {
+  handle_ctrl_break();
+}
+
+void factor_vm::primitive_disable_ctrl_break() {
+  ignore_ctrl_break();
+}
+
```
bjourne: You can just put the contents of the functions ignore/handle_ctrl_break in primitive_enable/disable_ctrl_break. No need to add extra wrappings.

basis/listener/listener.factor
jonenst: Yes it should be equivalent:
```
IN: scratchpad 2 [ '[ 1 _ 3 ] ] curry .
[ 2 [ [ 1 ] ] dip [ 3 ] curry compose ]
IN: scratchpad 2 '[ [ 1 _ 3 ] ] .
[ [ 1 2 3 ] ]
IN: scratchpad 2 [ '[ 1 _ 3 ] ] curry optimized.
[ [ 1 2 3 ] ]
IN: scratchpad 2 '[ [ 1 _ 3 ] ] optimized.
[ [ 1 2 3 ] ]
```

vm/os-windows.cpp
bjourne: Then in os-unix.cpp you need to define them as empty.

vm/safepoints.cpp
```diff
@@ -39,6 +40,14 @@ void safepoint_state::handle_safepoint(factor_vm* parent, cell pc) volatile {
...
       parent->end_sampling_profiler();
     std::cout << "Interrupted\n";
+    if (atomic::load(&ctrl_break_p) != 0) {
+      // Ctrl-Break throws an exception, interrupting the main thread, same
+      // as the "t" command in the factorbug debugger. But for Ctrl-Break to
+      // work we don't require the debugger to be activated, or even enabled.
+      atomic::store(&fep_p, false);
+      parent->general_error(ERROR_INTERRUPT, false_object, false_object);
+      FACTOR_ASSERT(false);
+    }
```
bjourne: Changing the signature of `enqueue_fep` I don't think is right. Instead add a flag to the `factor_vm` struct: `volatile cell skip_debugger_p` which you set to true in `primitive_enable_ctrl_break`. Then dispatch on that:

```c++
if (atomic::load(&parent->skip_debugger)) { parent->general_error ... } 
else { parent->factorbug() ... } 
```

vm/os-windows.cpp
```diff
@@ -304,6 +304,61 @@ void handle_ctrl_c() {
...
+        FACTOR_ASSERT(thread_vms.size() > 0);
+        THREADHANDLE thd = thread_vms.begin()->first;
+        factor_vm* vm = thread_vms.begin()->second;
+        /* Check if the VM thread has the same Id as the thread Id of the
+           currently active window. Note that thread Id is not a handle. */
+        DWORD fgThreadId = GetWindowThreadProcessId(GetForegroundWindow(),
+                                                    nullptr);
+        if (!atomic::load(&vm->safepoint.fep_p)
+            && (fgThreadId == vm->thread_id)
+        ) {
```
bjourne: Can't you use `GetThreadId(thd)` to avoid having to store `vm->thread_id`?

AlexIljin: I have considered that, and even had that implemented. Unfortunately, that would mean we drop WinXP support, as [GetThreadId](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683233%28v=vs.85%29.aspx) is only available since Vista.
I, for one, need my WinXP support.

AlexIljin:
Fixed all of the above.

vm/os-windows.cpp
```diff
@@ -304,6 +304,54 @@ void handle_ctrl_c() {
...
+           currently active window. Note that thread Id is not a handle. */
+        DWORD fgThreadId = GetWindowThreadProcessId(GetForegroundWindow(),
+                                                    NULL);
+        if (!atomic::load(&vm->safepoint.fep_p)
+            && (fgThreadId == vm->thread_id)
+        ) {
+          atomic::store(&vm->skip_debugger_p, true);
+          vm->safepoint.enqueue_fep(vm);
+          wake_up_thread(thd); /* Try to wake up the thread. */
+          ctrl_break_handled = true;
```
bjourne: `wake_up_thread` call is not necessary I think.

vm/os-windows.cpp
```diff
@@ -304,6 +304,54 @@ void handle_ctrl_c() {
...
+void factor_vm::primitive_enable_ctrl_break() {
+  stop_on_ctrl_break = true;
+  if (ctrl_break_thread == NULL) {
+    DisableProcessWindowsGhosting();
+    ctrl_break_thread = CreateThread(NULL, 0, factor::ctrl_break_thread_proc,
+                                     NULL, 0, NULL);
+  }
+}
+
 void lock_console() {}
```
bjourne: I think the polling thread should finish when ctrl+break is disabled. E.g your `while (true) ` should be `while (atomic::load(&stop_on_ctrl_break)) {. 

Also I only think you need one flag, the `vm->skip_debugger_p` one. When that flag is on, the loop is running and when off it terminates. More flags just complicate things.

vm/os-windows.cpp
AlexIljin: Shall we get rid of it in the ctrl-c handler as well, or just this place?

vm/os-windows.cpp
```diff
@@ -304,6 +304,54 @@ void handle_ctrl_c() {
   SetConsoleCtrlHandler(factor::ctrl_handler, TRUE);
 }
 
+static volatile bool stop_on_ctrl_break = false;
+static HANDLE ctrl_break_thread = NULL;
+
```
bjourne: Global variables are frowned upon. Instead place `ctrl_break_thread` in `factor_vm`. Like how the `sampler_thread` is handled.

vm/os-windows.cpp
bjourne: Just this place. The other I recall was needed for cancelling system calls.

vm/os-windows.cpp
```diff
@@ -304,6 +304,54 @@ void handle_ctrl_c() {
...
+        /* This proc runs in its own thread without stopping the main
+           thread. Since in practice nobody uses the multi-VM stuff yet,
+           we just interrupt the first VM we can get. */
+        FACTOR_ASSERT(thread_vms.size() > 0);
+        THREADHANDLE thd = thread_vms.begin()->first;
+        factor_vm* vm = thread_vms.begin()->second;
+        /* Check if the VM thread has the same Id as the thread Id of the
+           currently active window. Note that thread Id is not a handle. */
+        DWORD fgThreadId = GetWindowThreadProcessId(GetForegroundWindow(),
+                                                    NULL);
```
bjourne: nitpick: we use `names_like_these` in the vm.

vm/os-windows.cpp
```diff
@@ -304,6 +304,54 @@ void handle_ctrl_c() {
...
+        ) {
+          atomic::store(&vm->skip_debugger_p, true);
+          vm->safepoint.enqueue_fep(vm);
+          wake_up_thread(thd); /* Try to wake up the thread. */
+          ctrl_break_handled = true;
+        }
+      }
+    }
+    Sleep(100);
+  }
```
bjourne: Now I've experimented a lot and 100 ms makes to many ctrl+break presses get lost on my machine. If it doesn't affect performance 30 ms or less would be a better polling interval.

vm/safepoints.cpp
```diff
@@ -39,6 +39,15 @@ void safepoint_state::handle_safepoint(factor_vm* parent, cell pc) volatile {
...
+      // as the "t" command in the factorbug debugger. But for Ctrl-Break to
+      // work we don't require the debugger to be activated, or even enabled.
+      atomic::store(&fep_p, false);
+      atomic::store(&parent->skip_debugger_p, false);
+      parent->general_error(ERROR_INTERRUPT, false_object, false_object);
+      FACTOR_ASSERT(false);
+    }
     parent->factorbug();
     atomic::store(&fep_p, false);
   } else if (atomic::load(&parent->sampling_profiler_p)) {
```
bjourne: If you move `atomic::store(&fep_p, false);` before `if (atomic::load(&parent->skip_debugger_p))` you don't need two calls to it.

bjourne: There seem to be a bug wrt the `fep_p` state. If you run factor.com, so that you have both the gui listener and the console open, then run `[ t ] [ ] while` and switch to the console and press ctrl+c to interrupt it. Then `t` to throw and continue running. Then if you run a new `[ t ] [ ] while` and try to interrupt it with ctrl+break nothing happens.

vm/os-windows.cpp
```diff
@@ -304,6 +304,54 @@ void handle_ctrl_c() {
...
+           we just interrupt the first VM we can get. */
+        FACTOR_ASSERT(thread_vms.size() > 0);
+        THREADHANDLE thd = thread_vms.begin()->first;
+        factor_vm* vm = thread_vms.begin()->second;
+        /* Check if the VM thread has the same Id as the thread Id of the
+           currently active window. Note that thread Id is not a handle. */
+        DWORD fgThreadId = GetWindowThreadProcessId(GetForegroundWindow(),
+                                                    NULL);
+        if (!atomic::load(&vm->safepoint.fep_p)
+            && (fgThreadId == vm->thread_id)
```
bjourne: I don't think you need to check `!atomic::load(&vm->safepoint.fep_p)`. Enqueuing multiple feps has no effect.

vm/safepoints.cpp
AlexIljin: Yes, and I know exactly why that is. It's in this code:
```
parent->factorbug();
atomic::store(&fep_p, false);
```
When you throw the "t", `factorbug()` throws an exception, the second line is not executed and `fep_p` is never set to `false`. In the Ctrl-Break thread I'm explicitly checking  for `fep_p` to avoid double interruption.
To work around that problem in my code I set `fep_p` to `false` before calling `parent->general_error`.

Since `factorbug()` can throw exceptions, the correct implementation would be to clear `fep_p` in a `finally` clause. Unfortunately, there [is no such thing](https://stackoverflow.com/questions/161177/does-c-support-finally-blocks-and-whats-this-raii-i-keep-hearing-about) in C++. Alternatively, we could at least clear `fep_p` before calling `parent->general_error` in `factorbug()`.

bjourne: Yes.. It is very convoluted... But if you check the handler for ctrl+c (`ctrl_handler`) it doesn't check for `fep_p` so I don't think your code should have to do that either?

AlexIljin: I added it because of this code:
```
void safepoint_state::enqueue_fep(factor_vm* parent) volatile {
  if (parent->fep_p)
    fatal_error("Low-level debugger interrupted", 0);
```
I don't want my code to cause an interruption of the debugger, it's only intended for user-code.

vm/os-windows.cpp
AlexIljin: I totally support the sentiment, but would it be right to have a separate Ctrl-Break handler for every VM? Currently it **is** a global application-level service, even if its current implementation only interrupts the first VM in the global list of VMs.

vm/safepoints.cpp
AlexIljin: I am, in fact, checking the wrong flag there. I'm checking safepoint.fep instead of vm->fep. Hmm...

bjourne:
I think our discussion disappeared from the pr for some reason. Anyway you are right about the ctrl+break thread being global. 

But in my mind, it is better to be consistent even when if you are "consistently wrong" and the sampler thread is one per vm instance. An alternative is to make the ctrl+break feature work correctly when multiple vm:s are running. But the multi-vm code doesn't work very well and isn't used at all so I don't think it is worth the bother.

"I don't want my code to cause an interruption of the debugger, it's only intended for user-code." Yes, but when your code is running, won't be able to reach the low-level debugger. `handle_safepoint` will see that `skip_debugger_p` is true and throw the interrupt exception rather than calling `factorbug`. 

"I am, in fact, checking the wrong flag there. I'm checking safepoint.fep instead of vm->fep" I'm as confused as you are. There is a `vm->fep_p` and a `vm->safepoint.fep_p`, I don't know why two are needed or how they fit together.

vm/os-windows.cpp
```diff
@@ -304,6 +304,58 @@ void handle_ctrl_c() {
...
+  }
+}
+
+void factor_vm::primitive_enable_ctrl_break() {
+  stop_on_ctrl_break = true;
+  if (ctrl_break_thread == NULL) {
+    DisableProcessWindowsGhosting();
+    ctrl_break_thread = CreateThread(NULL, 0, factor::ctrl_break_thread_proc,
+                                     NULL, 0, NULL);
+  }
```
bjourne: If you change the 4th parameter to CreateThread to `static_cast<LPVOID>(this)`, then `ctrl_break_thread_proc` would get a reference to the running vm as its first argument.

AlexIljin:
@bjourne: "Yes, but when your code is running, won't be able to reach the low-level debugger. handle_safepoint will see that skip_debugger_p is true and throw the interrupt exception rather than calling factorbug." - It will, in fact, trip over the fep **before** handle_safepoint, see again the code in **enqueue_fep** I quoted:
```
void safepoint_state::enqueue_fep(factor_vm* parent) volatile {
  if (parent->fep_p)
    fatal_error("Low-level debugger interrupted", 0);
```
Essentially, I don't want to even touch `enqueue_fep` if `vm->fep_p == true`. So I'm checking it before the call.

vm/os-windows.cpp
AlexIljin: Good point, I'll use that. Thank you.

vm/vm.hpp
```diff
@@ -127,6 +127,9 @@ struct factor_vm {
   bool fep_disabled;
   bool full_output;
 
+  /* Used for Ctrl-Break handling: interrupt VM, but without debugging. */
+  volatile cell skip_debugger_p;
+
```
bjourne: I think you need to set it to `false` in the constructor.

AlexIljin: Yes, thank you!

mrjbq7:
What's the theory on why it crashes when interrupting background threads?

AlexIljin:
I'm no expert, so better listen to others, but my guess is that cooperative threads replace stack pointer, and whenever there is no exception handler on the stack, the app crashes all the way to the OS.

mrjbq7:
Well maybe that's the thing to fix. If we had a thread exception handler symbol, the ui.tools could set it with something that does get-listener display error or something.

AlexIljin:
That would be very cool!

bjourne:
General comment: It seems to get harder and harder for the thread to "hear" the ctrl+break press. First time I run `[ t ] [ ] while` and press <ctrl+break> it terminates almost immediately. Second time I do it, it is as fast, but third time it feels like it gets less responsive. Even if I lower the Sleep constant to 1 ms, some ctrl+break presses appears to not be noticed. It could be because I'm on a laptop keyboard so I have to press all three of <ctrl>+<fn>+<pause/break> for there to be a ctrl+break.

vm/os-windows.cpp
```diff
@@ -304,6 +304,58 @@ void handle_ctrl_c() {
...
+      THREADHANDLE thd = thread_vms.begin()->first;
+      factor_vm* vm = thread_vms.begin()->second;
+      /* Check if the VM thread has the same Id as the thread Id of the
+         currently active window. Note that thread Id is not a handle. */
+      DWORD fg_thd_id = GetWindowThreadProcessId(GetForegroundWindow(), NULL);
+      if ((fg_thd_id == vm->thread_id) && !vm->fep_p) {
+        atomic::store(&vm->skip_debugger_p, true);
+        vm->safepoint.enqueue_fep(vm);
+        wake_up_thread(thd); /* Try to wake up the thread. */
+        ctrl_break_handled = true;
```
bjourne: Shouldn't this be `wake_up_thread(factor::boot_thread);`?

AlexIljin:
@bjourne: "General comment: It seems to get harder and harder for the thread to "hear" the ctrl+break press."
I have pushed a patch that increases the ctrl-break thread priority relative to the main thread. Could you check it out?

vm/os-windows.cpp
AlexIljin: Technically, it's the same thread, but syntactically it makes sense to wake up the VM thread, because that's what we are trying to interrupt.

bjourne:
I've tried it now and the SetThreadPriority call doesn't seem to have any effect. Ctrl+break keypresses seem to randomly get lost, but I'm not sure if there is a pattern.

Anyway, VM implementation looks quite fine now.

I see that you are setting `vm->skip_debugger_p` to `true` and then back to `false` in `handle_safepoint`. I guess the reason is to differentiate between feps that originate from ctrl+c and ctrl+break keypresses?

I don't think that is necessary. In `handle_safepoint` you just have to check if `stop_on_ctrl_break` is `true` and if so, throw the `ERROR_INTERRUPT` rather than entering the low-level debugger.

The effect that has is that your logic becomes simpler because you can remove the `skip_debugger_flag_p` and it makes it so that if you press ctrl+c on the console it has the same effect as calling ctrl+break with the gui listener in focus. The handling of safepoints is a really complicated part of Factor so trying to make it as simple as possible is worthwhile.

On Linux, I don't think we can realistically trap the ctrl+break keypress. But with the changes I suggest above, with `stop_on_ctrl_break` set to true, you would be able to type `kill -SIGINT <factor-pid>` in a terminal and have that interrupt a gui listener stuck in an infinite loop.

catb0t:
> you would be able to type kill -SIGINT <factor-pid> in a terminal and have that interrupt a gui listener stuck in an infinite loop

One can already do `kill -SIGSEGV <factor-pid>` or `kill -SIGKILL <factor-pid>`, but I guess that at least lets us use CTRL-C on `factor`s opened from terminals.

jonenst:
> > you would be able to type kill -SIGINT in a terminal and have that interrupt a gui listener stuck in an infinite loop

> One can already do kill -SIGSEGV <factor-pid> or kill -SIGKILL <factor-pid>, but I guess that at least lets us use CTRL-C on factors opened from terminals.

Using SIGSEGV for this is quite a hack :) but it does interrupt some busy loops with "Memory protection fault". In my tests, SIGINT correcly interrupted ```: foo ( -- ) [ foo ] [ . ] recover ;``` but SIGSEGV killed the process.
SIGKILL doesn't do the same thing, it terminates the process.

@bjourne So if there is a a terminal with stdin/stdout we get the previous behavior (the low level shell), if there is no terminal and it's during the execution of the user input in the listener we throw directly ? 

And what happens when there is no terminal and we are not executing user code in the listener ? We throw directly also ? I don't know what is the current behavior but it kills the process.

bjourne:
If `-SIGSEGV` occurs during garbage collection, it crashes Factor which is not so good. 

@jonenst The behavior would depend on what sections of the code is wrapped in the `with-ctrl-break` combinator. Whether it then would terminate Factor would depend on if the `ERROR-INTERRUPT` exception was handled or not. 

Afaict, throwing ERROR-INTERRUPT is just as safe as throwing ERROR-DIVIDE-BY-ZERO and shouldn't cause any instability but I'm not 100% sure about that.

AlexIljin:
Hello, all!

I have pushed a new branch in my repo: [fix-1573-ri](https://github.com/AlexIljin/factor/tree/fix-1573-ri). It uses Raw Input API to wait for the Ctrl-Break key press event. No polling is performed, but the code is much more complicated than the polling solution.

Tell me what you think.

catb0t:
@AlexIljin It looks great! That's one platform down, two to go :)

AlexIljin:
I found a way to reduce the number of events used and simplified the code, the fix-1573-ri branch was updated and renamed fix-1573 so it is now the main branch to consider for this pull request. The previous fix-1573 is now known as fix-1573-polling.

AlexIljin:
Just FYI, I'm not planning to do any more work on this PR, unless you have comments or suggestions. So, if you were waiting for things to stabilize before reviewing, the time has come.

bjourne:
I can do it in a while when  I get time.

bjourne:
So do you mean that the code in this pull request (https://github.com/factor/factor/pull/1638/files) is not the latest and you have made a new branch for for the non-polling ctrl+break handling? If so, where do I find the branch with the new code? If I understand you rightly, then maybe you can open a new pull request?

AlexIljin:
This branch has the latest code. If you want the old poling code, it's in another branch.
Sorry for the confusion.

---=====--- 
Alexander

vm/safepoints.cpp
```diff
@@ -36,11 +36,18 @@ void safepoint_state::handle_safepoint(factor_vm* parent, cell pc) volatile {
   parent->faulting_p = false;
 
   if (atomic::load(&fep_p)) {
+    atomic::store(&fep_p, false);
     if (atomic::load(&parent->sampling_profiler_p))
       parent->end_sampling_profiler();
     std::cout << "Interrupted\n";
+    if (parent->stop_on_ctrl_break) {
```
bjourne: Think you need to add `stop_on_ctrl_break` to the vm struct too.

vm/os-windows.cpp
```diff
@@ -304,6 +304,145 @@ void handle_ctrl_c() {
...
+        }
+
+        if (lpb == NULL) {
+          // Memory allocation or need_size sanity check failed.
+          lpb_size = 0;
+          PostQuitMessage(CBT_MALLOC_FAIL);
+        } else if (GetRawInputData((HRAWINPUT)msg.lParam, RID_INPUT, lpb,
+                                   &lpb_size, sizeof(RAWINPUTHEADER))
+                   != lpb_size)
+        {
```
bjourne: I think the check should be `!= need_size`. I.e the return value of the second call to `GetRawInputData` should match the value you acquired in `need_size` in the first call.

bjourne:
Now I understand and I got it to work. But a big problem seem to be that the thread now steals all keyboard input. E.g if you type `readln`, you won't be able to type anything because input is consumed by the `PeekMessage` call in `ctrl_break_thread_proc`.

bjourne:
Also I'm thinking that it probably is no fun never getting your pr merged, so we could use your polling pr and apply that? After you had fixed my latest objections, it was very close to being ready imo. For simplicity's sake, we could also make it so that the listener wouldn't call the `enable/disable-ctrl-break` and require the user to do it for themselves if they want the ctrl+break interruption feature.

vm/os-windows.cpp
AlexIljin: Damn it, you are right! How did I miss that?

vm/safepoints.cpp
AlexIljin: Right. That was a cherry-pick gone wrong.

AlexIljin:
@bjourne: "... the thread now steals all keyboard input."
Yes, there is no way around that. I tried passing the messages through to the main thread, but it has no idea what to do with WM_INPUT. So, unless we want to do raw input processing in the main thread (which might be a good idea for gaming applications), we are stuck.

@bjourne: "we could use your polling pr and apply that."
I have updated the PR branch to point to the polling solution.

@bjourne: "we could also make it so that the listener wouldn't call the... ctrl+break interruption feature."
I can undo the Listener changes, but wouldn't that unfix the original issue? It was: "`[ t ] [ ] while` hangs the Listener if typed in interactive mode, making me forcibly kill Factor. It also hangs the compiler (and thus the Listener) if a file containing` [ t ] [ ] while` is `USE:`d or `reload`ed." (@catb0t)
Making user wrap _everything_ in a `with-ctrl-break`, is not the way to go, especially the module reloading command.

Also, note that currently calls to `with-ctrl-break` can't be nested, because they flip one global `bool` variable. The way it works now is fine for me, but otherwise it's up to you guys to say if that's acceptable or not.

bjourne:
Then I think the polling variant is the best solution. It would get awfully tricky to have one thread that takes all keyboard input and tries to delegate to the main thread. 

Yes, undoing the listener changes would unfix the original issue. But if you care about infinte-loop protection it wouldn't be so much trouble writing `enable-ctrl-break` once when starting the listener, I'd think. I thought running `with-ctrl-break` twice for every line entered would make the listener laggy, but if it doesn't, then no problem!

Another option, is to put `[ listener ] with-ctrl-break` on line 415 of `ui.tools.listener`. Then ctrl+break protection would be on for the duration of the ui listener.

I don't care that `with-ctrl-break` is not nestable. When we implement os threading we will have better ways to solve this problem.

AlexIljin:
@bjourne: "I thought running with-ctrl-break twice for every line entered would make the listener laggy, but if it doesn't, then no problem!"
In my experience there is no lag at all, so I'm inclined to keep the current implementation.

bjourne:
Then I have no objects to the pr being applied as is. @mrjbq7 @erg? I'm not so fond of the name `enable/disable-ctrl-break` because that is just the key-combo and doesn't tell you what feature is enabled/disabled. But I can't come up with a better name myself so it's no big problem.

mrjbq7:
I like the feature but I'm not a fan of having Factor run a thread that wakes up so frequently on my laptop or anything with a battery.

mrjbq7:
The other thing is we are getting close to an 0.98 release so making it opt-in for now or waiting until after release. I'd vote for opt-in because I like new code. 

Any way to make it less battery waking up?

AlexIljin:
@mrjbq7: "Any way to make it less battery waking up?"

Could you please elaborate on your use case?

The way I see it, the background thread only runs while your code is also running interactively in the Listener (or while it is being compiled). This means that you are already doing some work in the main thread anyway (possibly in a busy loop), so it's not fair to say that your CPU would be sleeping if it weren't for the polling. As soon as Factor is done with the interactive task, the polling is shut down completely, and everything can cool down and sleep again.

If you are running a long-term task which is mostly idle, like an HTTP server in Factor, you a) would not use a laptop for that, and b) would not run it interactively in a Listener, I presume.

Or am I wrong? What's the use case that concerns you?

mrjbq7:
Oh the thread is started and stopped for each listener input. 

Sometimes I run a couple listeners and do sometimes do long term http servers but that's okay. Often I do that in `with-thread`.

How does this handle multiple listeners open? Do we end up having multiple threads if two listeners are "processing input"?

AlexIljin:
@mrjbq7: "How does this handle multiple listeners open? Do we end up having multiple threads if two listeners are "processing input"?"

One thread per VM. Since we only have one VM, there is at most one polling thread at any given time within the Factor.exe process.

mrjbq7:
So if two listeners run `with-ctrl-break` at the same time, which listener is an error thrown into? The one with the error or the other one? If the second subsequently errors what happens?

mrjbq7:
I'm traveling for the next week so at hard to respond (sorry!) but am wondering if the polling thread consumes I/O in any way that it would be lost from other factor threads that are running in other listeners or the background.

AlexIljin:
@mrjbq7: "So if two listeners run `with-ctrl-break` at the same time, which listener is an error thrown into?"

The error is thrown into VM, not Listener. If an interactive task yielded, you are most likely screwed, because ctrl-break will happen at an unexpected time and crash a random cooperative thread, bringing down the whole process. There is currently no way around that, as was discussed in this thread, so we are supposed to take that as a known limitation.

@mrjbq7: "wondering if the polling thread consumes I/O in any way that it would be lost from other factor threads that are running in other listeners or the background."

Nope, no I/O is consumed or lost.

mrjbq7:
Okay, that makes sense. Apologies for having you repeat yourself. No guarantees I don't forget and ask again! Hah!

What do you think about making this opt-in for now and documenting that any caller can use `with-ctrl-break` to run code with the ability to interrupt?

erg:
I think it's a cool feature and worth merging. Any plans or way to implement this on linux and mac, or just a sketch of how it would work? (We can merge this PR before that, since the implementation would probably take awhile)

AlexIljin:
@mrjbq7: "Apologies for having you repeat yourself."

You've probably got a lot going on : )

@mrjbq7: "What do you think about making this opt-in for now"

I'd prefer the current behavior (not opt-in), but only because it suits my use cases so well. I think you guys know better and see the bigger picture, so I'm fine with whatever makes sense to you. I understand if don't want to include things that may be considered risky or unproven before a release, but in the long term it makes very little sense as an opt-in feature.

@erg: "Any plans or way to implement this on linux and mac, or just a sketch of how it would work?"

No plans from me, I'm afraid. I have only ever developed stuff for Windows.

jonenst:
I think it should be opt out with some configuration in .factor-rc so that
newcomers have the feature on, and more advanced people who don't want it
can easily disable it. Also add a 'tip of the day' describing this.

I don't know what percentage of people don't have a terminal associated
with the UI listener to drop to the lowlevel shell and press 't' to throw.
Also there are the people who don't know how to send SIGINT to the process.
But I suspect this a good part of the users.
Le 14 juin 2016 21:46, "Alexander Iljin" <notifications@github.com> a
écrit :

bjourne:
@erg I don't know about mac, but the feature is mostly impossible to implement on Linux. The ecosystem is to heterogeneous. But you'll be able to interrupt it like ctrl+break with the kill command: `` kill -SIGINT `pidof factor` ``

jonenst:
The new thing is that kill -SIGINT will work even if there is no stdin/stdout (where we would get the low level shell), right ? Currently it crashes factor.

vm/os-unix.cpp
```diff
@@ -491,6 +491,14 @@ void handle_ctrl_c() {
...
 
+void factor_vm::primitive_disable_ctrl_break() {
+  // TODO
+}
+
+void factor_vm::primitive_enable_ctrl_break() {
+  // TODO
+}
+
 void abort() {
```
bjourne: Think you should set `stop_on_ctrl_break = false;` and `stop_on_ctrl_break = true;` in these primitives.

jonenst:
What about squashing most of the commits together ? 34 commits make a complicated history for a relatively small patch (+88 -12). I'm not sure, but it looks like most of them don't provide much value, except for the fact that you were learning about the vm and factor codebases ?

jonenst:
Also, in the current state, as noted by @bjourne, it doesn't compile on other platforms than windows because ```volatile bool stop_on_ctrl_break;``` is only defined inside the ```#if WINDOWS```

We want it on all platforms, along with the code that triggers general_error(INTERRUPT..), so that SIGINT throws an error for a busy loop even if there is no stdin/stdout for the lowlevel debugger shell.

However, I think we want to keep the lowlevel debugger shell if stdin and stdout are available and SIGINT is sent during the execution of interactive input. I think this is currently not implemented.

AlexIljin:
@jonenst: "What about squashing most of the commits together ?"

I think that would hamper my goal of getting into the [top ten](https://github.com/factor/factor/graphs/contributors) by the end of this year!

Seriously, though, I've cleaned up the history down to 4 commits.

@jonenst: "`stop_on_ctrl_break` is only defined inside the `#if WINDOWS`"

Fixed it.

@jonenst: "However, I think we want to keep the lowlevel debugger shell if stdin and stdout are available and SIGINT is sent during the execution of interactive input. I think this is currently not implemented."

On Windows you have both behaviors. If there is console, both Ctrl-Break and Ctrl-C work; if there is no console, only Ctrl-Break works. On other platforms only Ctrl-C works, and only with a console. On all platforms Ctrl-C interrupts a busy loop first, and only breaks into low-level debugger if there is no interactive code currently running (interactive code is defined as any code executed between the enable-ctrl-break and disable-ctrl-break calls).

Current implementation is as described above, but I only tested it on Windows, and only tried building the project with VS 2015. For additional testing I need help from the others in this thread.

jonenst:
> On all platforms Ctrl-C interrupts a busy loop first, and only breaks into low-level debugger if there is no interactive code currently running (interactive code is defined as any code executed between the enable-ctrl-break and disable-ctrl-break calls).

I think it should not behave like this. I think sometimes we want the low-level debugger during interactive code and with this behavior it's not possible. What do you think ?

Instead, I think it should do this:
  * if there is no console and the code is interactive: interrupt a busy loop
  * otherwise, the old behavior (if there is a console, drop to the low level shell. if there is no console, I'm not sure what the old behavior is; we should find out and decide if it's correct or not)

bjourne:
@jonenst, it is impossible to detect if there is a console or not. Or if it isn't impossible it is at least very hard. Therefore I think if you run `enable-ctrl-break` it is like declaring that you don't have a console and you don't want to use the low-level debugger. So maybe `enable-ctrl-break` should be thought of as `disable-low-level-debugger`...

jonenst:
```USE: unix.ffi 0 isatty``` outputs 1 when there is a console and 0 when there is not for me (ubuntu with unity)

**Comment by AlexIljin** [comment]
@jonenst: "I think sometimes we want the low-level debugger during interactive code and with this behavior it's not possible. What do you think ?"

I see your point about the missing feature. Currently we did lose the ability to use low-level debugger for the interactive code. That feature was explicitly requested by @bjourne, and implemented per that request, overriding the previous behavior. You two should discuss this and make a decision. Here's the relevant quote from above.

@bjourne: "I don't think that is necessary. In handle_safepoint you just have to check if stop_on_ctrl_break is true and if so, throw the ERROR_INTERRUPT rather than entering the low-level debugger.

The effect that has is that your logic becomes simpler because you can remove the skip_debugger_flag_p and it makes it so that if you press ctrl+c on the console it has the same effect as calling ctrl+break with the gui listener in focus."

**Comment by AlexIljin** [comment]
@bjourne: "it is impossible to detect if there is a console or not."

I'm not sure, but I think the STDIN and such pipes would contain INVALID_HANDLE_VALUE if there is no console associated with the process on Windows. We could test this to see if it works.

**Comment by catb0t** [comment]
> It is impossible to detect if there is a console or not

On Unicies, we have `unix.ffi:isatty`, from [the `unistd.h` function of the same name](http://man7.org/linux/man-pages/man3/isatty.3.html). Windows API probably has some equivalent...

**Comment by bjourne** [comment]
Yes, I think it is right that if you have `enable-ctrl-break` on, ctrl+c should *not* start the low-level debugger. That's what allows the `kill -SIGINT <factor-pid>` command to interrupt an infinite loop on Linux in Factor when you don't have a console window. The logic in your pr is very simple and that's what I like. I don't know *when* `enable-ctrl-break` should be turned on. My opinion is more and more that it is something users should activate for themselves since it is so hard for us to agree on what best behavior is.

But if tty detection works that would be better. Though if you run `/factor &` from a console, `isatty` seem to not give the right result, but I'm not sure.

**Comment by AlexIljin** [comment]
Hi, all! I have pushed one more patch. Now F2 command can be interrupted with Ctrl-Break, which closes another UI hole that could get the Listener stuck. This adds to the `USE:`ing and `reload`ing vocabs use cases, which are already covered.

File: basis/ui/tools/listener/listener-docs.factor [hunk 19]
```diff
@@ -97,4 +97,8 @@ TIP: "Scroll the listener from the keyboard by pressing " { $command listener-ga
 
 TIP: "Press " { $command tool "common" refresh-all } " or run " { $link refresh-all } " to reload changed source files from disk. " ;
 
+os windows? [
+    TIP: "Use C+Break to interrupt tight loops in your code started in the Listener, such as" { $code "[ t ] [ ] while" } "Caution: this may crash the Factor runtime if the code uses cooperative multitasking or asynchronous I/O." ;
+] when
+
```
jonenst: Since ```TIP:``` is a parsing word, I think the conditional doesn't do anything

mrjbq7: You could just say, "On Windows, ..." or (currently supported on Windows only..." etc

AlexIljin:
Added "On Windows: " prefix, removed the `os windows?` conditional.

AlexIljin:
I'd like to remind everyone that this does work, and that there are no unresolved issues with the code or documentation in this PR.

basis/listener/listener.factor
```diff
@@ -24,8 +24,13 @@ H{
 M: object prompt.
     nip prompt-style get-global format bl flush ;
 
+: with-ctrl-break ( quot -- )
+    enable-ctrl-break
+    [ disable-ctrl-break ] [ ] cleanup ; inline
+
 : parse-lines-interactive ( lines -- quot/f )
-    [ parse-lines ] with-compilation-unit ;
+    [ [ parse-lines ] with-ctrl-break
```
mrjbq7: So, we were waiting until 0.98 is released to merge this, mostly because of this change right here that enables it... instead of being optional before 0.98.  So, we could either make it optional here, and disabled by default, and merge before 0.98 (which is waiting for a couple fixes and @erg and I have to have some time), or wait until afterward to merge.

AlexIljin: OK, that explains it.

I usually have more branches than just the `master` to handle these things - like the `next` branch for the upcoming release, etc. But that, perhaps, comes with branch managing costs which I enjoy, but others may not like to waste their time on. Anyway, I'll leave this link here just in case: http://nvie.com/posts/a-successful-git-branching-model/

Ah, the 0.98 release... I remember, if I'm allowed to be a bit sentimental, I remember this quote by John Benediktsson as clearly as if it was uttered just on September 22nd, 2015: "Everything that's been committed since 0.97 (see the nightly builds) will be in 0.98 and our plan right now is to fix bugs and cleanup in preparation for a release in a week or two."

Good ol' days... : ))

Excuse me, friends, if I sound a bit frisky. I've been binging on "A bit of Fry and Laurie" lately, that explains my somewhat poetic mood.

mrjbq7: Hah, I've been promising 0.98 for awhile (even longer than that!), but real life has gotten somewhat in the way.  Plus, the two main bugs I want to get fixed are unfixed (Slow GTK resize on @erg's VM, and OpenGL initialization on macOS).

AlexIljin: @mrjbq7: I've added a patch to this PR that makes enable-ctrl-break calls opt-in, off by default.

Will you merge now?

basis/ui/tools/tools.factor
```diff
@@ -1,13 +1,16 @@
...
+ui.tools.listener ui.tools.operations ui.tools.walker
+vocabs.refresh ;
 IN: ui.tools
 
 MAIN: listener-window
 
-\ refresh-all H{ { +nullary+ t } { +listener+ t } } define-command
+: refresh-all-cmd ( -- ) [ refresh-all ] with-ctrl-break ;
+
+\ refresh-all-cmd H{ { +nullary+ t } { +listener+ t } } define-command
```
mrjbq7: Instead of doing this just for ``refresh-all``, maybe we should instead change the ``ui.operations`` code to do this for all commands?

vm/safepoints.cpp
```diff
@@ -35,11 +35,18 @@ void factor_vm::handle_safepoint(cell pc) {
   faulting_p = false;
 
   if (atomic::load(&safepoint_fep_p)) {
+    atomic::store(&safepoint_fep_p, false);
```
mrjbq7: @bjourne is there any problem with moving this store to the top of this block?  I would think it would be better to leave it at the end.

AlexIljin: @mrjbq7: an exception is being thrown below, so the code at the bottom will never be executed.

basis/ui/tools/tools.factor
AlexIljin: I'm not sure if that's a good idea, so I'll let someone with more knowledge of the Factor UI framework to reply.

My goal was to localize the changes and have as little side-effects to parts unknown to me as possible.

AlexIljin:
Why not merge this PR?

vm/safepoints.cpp
bjourne: Actually putting it at the top seem to be right and fixes an unrelated bug. If you press t to throw an exception in the low-level debugger the `safepoint_fep_p` variable gets stuck at true state. So if you then run something triggering safepoints like profiling (`[ 3 seconds sleep ] profile`) it breaks.

mrjbq7:
@bjourne ready to merge this?

bjourne:
Yeah I have no complaints.

vm/os-windows.cpp
```diff
@@ -296,6 +296,50 @@ void handle_ctrl_c() {
   SetConsoleCtrlHandler(factor::ctrl_handler, TRUE);
 }
 
+const int ctrl_break_sleep = 10; /* msec */
```
mrjbq7: Doesn't seem like we should check for interrupted so frequently.  When you press ``Ctrl-Break``, it doesn't have to be "immediate", it can be "almost immediate".

Also, wasn't a no-polling patch in the works?  I recall you mentioning that at some point?

mrjbq7:
I merged this, but would prefer less frequent polling and maybe no-polling if possible.

mrjbq7:
Also, it is unclear how it works when running two Listeners.  So, for example, in the first listener you do this:

```factor
IN: scratchpad 5 seconds sleep
```

In the second listener you do:

```factor
IN: scratchpad [ t ] [ ] while
```

Now, after 5 seconds you try and ``Ctrl-Break``, it shouldn't work because the first one disabled it.

mrjbq7:
@AlexIljin I closed it, but maybe you can comment on the above comments.

mrjbq7:
This seems to have broken ``Ctrl-C`` on Unix.

```
Factor 0.98 x86.64 (1789, heads/fix-1573-1156ea3d8b, Tue Sep 20 06:59:43 2016)
[Clang (GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1))] on macosx
IN: scratchpad ^CInterrupted
^CInterrupted
^CInterrupted
^CInterrupted
^CInterrupted
^CInterrupted
^Z
```

mrjbq7:
Oh, because on Unix, ``handle-ctrl-break`` is now defaulting to ``f``.

vm/os-windows.cpp
AlexIljin: Actually, the delay is not about the reaction speed, it's about catching the Ctrl-Break combo in its depressed state. It turned out to be easy to miss with longer delays. It has been established earlier in this discussion that polling is extremely lightweight and doesn't affect anything.

The no-polling patch was implemented based on Raw Input API, but didn't work, because the watcher stole all keyboard input, so commands like `readln` became impossible in the listener.

@bjourne: "I think the polling variant is the best solution. It would get awfully tricky to have one thread that takes all keyboard input and tries to delegate to the main thread. "

AlexIljin: @mrjbq7: "Also, it is unclear how it works when running two Listeners."

Two listeners can't be supported, unfortunately, because it's the VM that gets interrupted, not a specific Factor coop-thread.

AlexIljin:
@mrjbq7: "This seems to have broken Ctrl-C on Unix."

Are you sure? Anything but Windows should be completely unaffected by this patch: VM functions are empty for other platforms, etc.

mrjbq7:
Yeah, the issue is that ``safepoint_fep_p`` gets set to ``false`` before calling ``factorbug()`` where as before it was set afterwards.  That difference causes ``Ctrl-C`` on Unix to not trigger the debugger unless ``stop_on_ctrl_break`` is set to true.

mrjbq7:
Okay, I moved ``safepoints_fep_p`` to false back to its old location and set it to false before throwing the ctrl-break interrupt.  That brings back the error @bjourne pointed out with profile, but we can fix that separately.

vm/safepoints.cpp
mrjbq7: @AlexIljin is there a reason you can't just ``enqueue_fep()`` and then ``t`` to throw an exception into Factor?  Then we wouldn't need the ``stop_on_ctrl_break`` boolean.

bjourne:
Now I'm not sure what's going on, but on my machine the placement of `atomic::store(&safepoint_fep_p, false);` at the top does not seem to affect the debugger.

vm/safepoints.cpp
AlexIljin: @mrjbq7: "is there a reason you can't just enqueue_fep() and then t to throw an exception into Factor?"

The main use case for the entire feature is when we have no console to `t` in. Otherwise we could just use Ctrl-C in the first place.

mrjbq7:
@AlexIljin you might try using a couple different Listener windows (as thats a common use-case that I personally use).  You can make a new one with ``listener-window``.  It would be nice to support that better.

AlexIljin:
@mrjbq7: I have an idea, but I need to experiment with it. Basically, any time something yields, we should automatically call `disable-ctrl-break`. If that was a background process, it'll have no effect, but if it was some user code started interactively, it'll do the right thing and prevent crashing the VM when user code spawns a thread and never returns.

I'll see what can be done regarding the two listeners. Maybe some further instrumentation in the VM could help us here.

PS: why does Github say `Closed with unmerged commits This pull request is closed, but the AlexIljin:fix-1573 branch has unmerged commits`? I checked, and I don't see anything left out.

mrjbq7:
That's weird, maybe because I rebased on master the commits got new refs?

AlexIljin:
@mrjbq7: From my side, I tried rebasing the branch onto the `master` after your merge, and the list of commits was "noop". Seems like a Github bug to me.

bjourne:
It happens because when the commits are merged, they get your name as "Author" and John's name as "Committer". But in your repo, your name is both on the Author and Commit field of the commits so Github doesn't understand that they are merged. It happened to me too a lot but is nothing to worry about.

PR #1642, opened 2016-06-04
rosetta-code.image-noise, use status bar for FPS

jonenst:
Not sure if there is a simpler way do use the status bar with MAIN-WINDOW: ?


mrjbq7:
Merged, thanks!

PR #1644, opened 2016-06-06
extra: add vocab to wrap tools.test for Exercism.io

catb0t:
The exercism.testing vocabulary makes it easier
for Exercism users and
http://github.com/exercism/xfactor collaborators
to run unit-tests on exercise code, by providing
a common interface to tools.test for both cases.

Because the vocabulary is intended to work on the
current directory, it relies on filesystem state
and unfortunately much of its code is not easily
unit-test-able. However, the vocab itself has
been rigorously tested and verified by yours
truly so it does exactly what it should in all
cases I can think of.

See exercism.testing's documentation for more
information.

**Why do I think this should be in Factor's standard library?**

There are many things that I think make Factor a Joy (heh) to program in, but a big one is its massive standard library which doesn't seem to be afraid of getting bigger. 

Factor's naming conventions coupled with Exercism's naming conventions make writing Factor code in an Exercism context a little bit tricky. The goal of https://github.com/exercism/xfactor is to get people interested in Factor by making it easy and fun to solve Exercism exercises.

Because Factor doesn't yet have an easy way to portably distribute a library to users (think npm for Node.js or pip for Python), and because it's not reasonable to expect all Exercism users to have `git` and the knowledge to install and use a third-party Factor library (and because for most other Exercism tracks, at least the client-side unit testing code is part of the standard library), I think an effective and easy way to distribute this library to Exercism users interested in Factor is to put it in the standard library.

Users have Factor, and they will also have code that lets them use the canonical `tools.test` on Exercism.

Please let me know what you think. :)

(By the way: It's called `exercism.testing` because I'll probably add an [`exercism.client`](https://github.com/exercism/cli) later.)

### DIFF
--- a/extra/exercism/testing/testing.factor
+++ b/extra/exercism/testing/testing.factor
@@ -1,0 +1,232 @@
+! See http://factorcode.org/license.txt for BSD license.
+USING: accessors arrays assocs classes.tuple combinators
+    command-line formatting io io.directories io.encodings.utf8
+    io.files io.files.info io.launcher io.pathnames json.reader
+    kernel locals math multiline parser present sequences sets
+    sorting summary system tools.test vocabs.loader ;
+QUALIFIED: namespaces
+QUALIFIED: sets
+IN: exercism.testing
+
+<PRIVATE
+
+
+: child-directories ( path -- directories )
+    directory-entries
+    [ directory? ] filter
+    [ name>>     ] map ;
+
+CONSTANT: name-clashes { "hello-world" }
+
+TUPLE: config
+        { problems   array }
+        { deprecated array } ; final
+
+SYMBOL: project-env
+ERROR:  wrong-project-env word ;
+
+TUPLE: user-env ; final
+M:     user-env present
+    drop "user-env" ;
+ERROR: not-user-env  < wrong-project-env ; final
+M:     not-user-env  summary
+    word>> name>> "can't use word %s in dev environment" sprintf ;
+
+TUPLE: dev-env ; final
+M:     dev-env present
+    drop "dev-env" ;
+ERROR: not-dev-env  < wrong-project-env ; final
+M:     not-dev-env  summary
+    word>> name>> "can't use word %s in user environment"  sprintf ;
+
+ERROR:  not-an-exercism-folder word ;
+M:      not-an-exercism-folder summary
+    word>> name>> "exercism.testing: %s: current directory is not an exercism folder" sprintf ;
+
+
+HOOK: exercises-folder project-env ( -- dirname )
+M: dev-env  exercises-folder  "exercises" ; inline
+M: user-env exercises-folder  "."         ; inline
+M: f        exercises-folder  \ exercises-folder not-an-exercism-folder ;
+
+
+HOOK: exercise>filenames project-env
+    ( test-name -- example-filename tests-filename )
+M: dev-env exercise>filenames
+    dup exercises-folder prepend-path prepend-path
+    { "-tests.factor" "-example.factor" }
+    [ append ] with map first2 ;
+
+M: user-env exercise>filenames
+    dup prepend-path
+    { "-tests.factor" ".factor" } [ append ] with map
+    first2 ;
+
+: (handle-name-clash) ( -- )
+    vocab-roots namespaces:get reverse vocab-roots namespaces:set ;
+
+HOOK: handle-name-clash project-env ( -- )
+M: dev-env handle-name-clash
+    exercises-folder prepend-path-path add-vocab-root
+    (handle-name-clash) ; inline
+
+M: user-env handle-name-clash
+    add-vocab-root
+    (handle-name-clash) ; inline
+
+
+HOOK: get-config-data project-env ( -- config )
+M: dev-env get-config-data
+    "config.json" path>json
+    { "problems" "deprecated" } [ swap at ] with map
+    config slots>tuple ;
+
+M: user-env get-config-data
+    \ get-config-data not-user-env ;
+
+
+HOOK: exercise-exists? project-env ( exercise -- ? )
+M:: dev-env exercise-exists? ( name -- ? )
+    name
+    [ get-config-data problems>> member? ]
+    [ exercises-folder prepend-path exists? ]
+    bi and
+    [ name exercise>filenames [ exists? ] bi@ and ]
+    [ f ]
+    if ;
+
+M: user-env exercise-exists?
+    dup exercise>filenames [ exists? ] tri@ and and ;
+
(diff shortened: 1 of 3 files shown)


extra/exercism/testing/testing.factor
```diff
@@ -0,0 +1,227 @@
...
+M: user-env exercise-exists?
+  dup exercise>filenames [ exists? ] tri@ and and ;
+
+M: f exercise-exists?
+  drop \ exercise-exists? not-an-exercism-folder ;
+
+
+HOOK: config-exclusive? project-env ( problems deprecated -- ? )
+M: dev-env config-exclusive?
+  sets:intersect { } = ;
```
mrjbq7: Using ``sets:intersects? not`` is more efficient.

extra/exercism/testing/testing.factor
```diff
@@ -0,0 +1,227 @@
...
+
+: (run-exercism-test) ( exercise -- )
+  dup "hello-world" = [ handle-hello-world ] when
+  [ "\ntesting exercise: %s\n\n" printf ]
+  [ exercise>filenames ]
+  bi
+  run-file run-test-file ;
+
+: wd-git-name ( -- name )
+  "git rev-parse --show-toplevel" utf8 [ contents ] with-process-reader*
```
mrjbq7: This isn't probably cross-platform.  Do you intend this to be used on unix platforms only?  And do you expect this to be used with a git checkout of factor versus a binary distribution?

extra/exercism/testing/testing.factor
```diff
@@ -0,0 +1,227 @@
...
+  } [ exists? ] all?
+  dup [ T{ user-env } project-env namespaces:set ] when
+
+  xor
+  [
+    current-directory project-env [ namespaces:get ] bi@
+    "working directory OK: %s is a %s \n" printf
+  ]
+  [
+    current-directory namespaces:get
```
mrjbq7: You shouldn't use ``current-directory set``, it's bad practice because it doesn't restore the previous value when the computation is completed, thus losing it.  Better would be ``path current-directory [ ... ] with-variable``

mrjbq7:
Should use 4-spaces (instead of 2) if wanting to be in the factor distribution.

Can't you distribute this vocabulary as part of the "server delivered" vocabularies you mentioned in one of your earlier posts?

Is this intended to be used as a command-line test harness for exercism?  If so, why not distribute it as a factor script ``./factor exercism-testing.factor blah blah`` or a deployed binary they can download?

I'm a little confused about it's use-case, sorry.

mrjbq7:
Also, I guess the other thing is users can just do ``"leap" test``, but the server can't so this is mainly to support the server?  If that's the case then you don't need to distribute it at all but just keep it in a vocab root accessible by factor on your server?

extra/exercism/testing/testing.factor
catb0t: Will fix, thanks.

extra/exercism/testing/testing.factor
catb0t: I think you may have misread, I am aware of the dangers of `current-directory set` but this is just testing what the current directory *is*.

mrjbq7: Indeed!

extra/exercism/testing/testing.factor
catb0t: `wd-git-name` is supposed to get the "name" of the current git repository. In an Exercism collaborator's local copy of https://github.com/exercism/xfactor, it will be `xfactor`. If the user is not in a git repository or does not have `git` installed, `with-process-reader` returns an exit code above 0 and we know to return empty string (or something that isn't equal to `"xfactor"`.)

What isn't cross platform here? I guess it maybe should be ascii for Windows, but since Windows doesn't know about Unicode anyways it shouldn't? be a problem. I'll change that if you like.

catb0t:
> Should use 4-spaces (instead of 2) if wanting to be in the factor distribution.

Good to know, I'll change that.

> Can't you distribute this vocabulary as part of the "server delivered" vocabularies you mentioned in one of your earlier posts?

In the xfactor git repo, yes, but Exercism users (who don't help maintain xfactor) just get shipped unit tests their implementations should meet. 

> Is this intended to be used as a command-line test harness for exercism? If so, why not distribute it as a factor script `./factor exercism-testing.factor blah blah` or a deployed binary they can download?

Well, the idea of Exercism is to get users to learn to use and work with the language's and platform's default tools (in this case, tools.test). But because of the naming conventions and name clashes between Exercism folder names and Factor vocabulary names, (`hello-world` is a good example), it's tricky to use vanilla tools.test. 

With this in the standard lib, an Exercism user and / or collaborator can sit in the appropriate folder and: 
```
factor -run=exercism.testing run-all
```
or open a listener and:
```
IN: scratchpad USE: exercism.testing "hello-world" run-exercism-test
```
Without external dependencies that *feel* non-standard and third party.

> Also, I guess the other thing is users can just do "leap" test, but the server can't so this is mainly to support the server? If that's the case then you don't need to distribute it at all but just keep it in a vocab root accessible by factor on your server?

Users *can* just do `"leap" test`, but they cannot just do `"hello-world" test` without editing their vocabulary roots  and reversing `vocab-roots`. (see `handle-hello-world`'s implementation, which should be refactored to support handling of any exercism/factor name clashes).

`exercism.testing` makes it easy for Exercism users to test all their Exercism solutions without "third-party" tools and without needing to reverse `vocab-roots` (which might be above a newbie's head to understand) , and makes it easy to verify that the example code on the server (or in collaborator's repos) passes the same tests in the same way.

mrjbq7:
Hmm, I can see what you're looking for, but I wonder if a better solution might be to have them download an exercism support vocab root and add it to Factor (either ``add-vocab-root`` or ``.factor-roots``) or be served that library as part of each exercise?

Regarding naming conflicts.  Is ``hello-world`` the only one?  I might suggest calling it ``hello``, or put the exercises in a namespace (``exercism.hello-world``), or we can rename ours since it's just a sample library...

catb0t:
> have them download an exercism support vocab root and add it to Factor

Hm... okay, I guess that's where we're going. I thought I might be successful here, but I see what you're saying.

> Regarding naming conflicts. Is hello-world the only one? 

The intersection of https://github.com/exercism/x-common and `all-vocabs-really` is: 
```
{ "hello-world" "binary-search" "poker" }
```

> I might suggest calling it hello, or put the exercises in a namespace (exercism.hello-world), or we can rename ours since it's just a sample library...

I guess `hello-world` could be renamed (or perhaps there should be an examples vocab, to put `poker` and `hello-ui` into as well?), but `binary-search` seems important.

PR #1649, opened 2016-06-11
peg.javascript.tokenizer: allow escaped quotes in strings

### LINKED ISSUE #1648: peg.javascript should recognise escaped quotes
I'd fix this in a pull request if I could understand [the tokenizer lines in question](https://github.com/factor/factor/blob/master/extra/peg/javascript/tokenizer/tokenizer.factor#L51) well enough to add this.

```factor
IN: scratchpad HEREDOC: END
"abc\"def\""
END 
javascript
```
Should return: 

```
T{ ast-begin
    { statements V{ T{ ast-string { value "abc\"def\"" } } } }
}
```
But instead it gives:

```
T{ ast-begin { statements V{ } } } 
```

catb0t:
Previously, the tokenizer didn't understand
escaped quotes in string literals. Also added was
a test to ensure the escaping works.

Fixes #1648.

### DIFF
--- a/extra/peg/javascript/tokenizer/tokenizer-tests.factor
+++ b/extra/peg/javascript/tokenizer/tokenizer-tests.factor
@@ -23,3 +23,7 @@ IN: peg.javascript.tokenizer.tests
 { V{ T{ ast-regexp f "<(w+)[^>]*?)/>" "g" } } } [
   "/<(\\w+)[^>]*?)\\/>/g" tokenize-javascript
 ] unit-test
+
+{
+  V{ T{ ast-string { value "abc\"def\"" } } }
+} [ "\"abc\\\"def\\\"\"" tokenize-javascript ] unit-test
--- a/extra/peg/javascript/tokenizer/tokenizer.factor
+++ b/extra/peg/javascript/tokenizer/tokenizer.factor
@@ -8,7 +8,7 @@ IN: peg.javascript.tokenizer
 
 USE: prettyprint
 
-EBNF: tokenize-javascript 
+EBNF: tokenize-javascript
 Letter            = [a-zA-Z]
 Digit             = [0-9]
 Digits            = Digit+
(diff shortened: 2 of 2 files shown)


catb0t:
(My editor trims trailing whitespace on save, I didn't realise those other lines had changed... Trailing whitespace is a health hazard anyways :) )

extra/peg/javascript/tokenizer/tokenizer-tests.factor
```diff
@@ -23,3 +23,10 @@ IN: peg.javascript.tokenizer.tests
 { V{ T{ ast-regexp f "<(w+)[^>]*?)/>" "g" } } } [
   "/<(\\w+)[^>]*?)\\/>/g" tokenize-javascript
 ] unit-test
+
+{
+  V{
+    T{ ast-name { value "abc" } }
+    T{ ast-string { value "def" } }
+  }
+} [ "abc\"def\"" tokenize-javascript ]
```
mrjbq7: This doesn't quite test the escapes, you'd need something like:

```"\"abc\\\"def\"" tokenize-javascript```

catb0t: Well, of course it doesn't. *facepalm*

Next time I'll try *running* the tests first...

catb0t:
There we go.

mrjbq7:
Merged, thanks!!

I added a few more escapes in 29a2a2595476d6ccb04d17c33979ad1a1a090d21.  Didn't add ``\0`` but maybe we should, dunno.

mrjbq7:
P.S. if you want to add support for HexEscapes ``\x..`` and UnicodeEscapes (``\u....''), go for it :-)

http://www.ecma-international.org/ecma-262/5.1/#sec-7.8.4

mrjbq7:
OctalEscapes would be good too, but they look a little complicated.

catb0t:
OctalEscapes are buggy and a non-feature in JS, but I'll get around to it at some point :)

mrjbq7:
Added hex and unicode escapes in e7fa92baacd424e99008087b7eaaa797602456ce.

I think maybe octal isn't worth it!

PR #1663, opened 2016-06-28
For upstream

AlexIljin:
A new batch of updates has arrived.

In this branch:
- removed the blocking copy-file implementation on Windows;
- added default cross-platform move-file implementation;
- added create-guid in windows.ole32;
- added ensure-index in db.queries and related error tuple, with an SQLite implementation and tests;
- reworked checksums.openssl to adhere to the common interface: initialize-checksum-state, add-checksum-bytes/stream, get-checksum;
- plus some documentation fixes and updates.

### DIFF
--- a/basis/checksums/openssl/openssl.factor
+++ b/basis/checksums/openssl/openssl.factor
@@ -1,7 +1,7 @@
 ! Copyright (C) 2008, 2010, 2016 Slava Pestov, Alexander Ilin
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors alien.c-types alien.data checksums
-checksums.stream destructors io kernel openssl openssl.libcrypto
+checksums.stream destructors fry io kernel openssl openssl.libcrypto
 sequences ;
 IN: checksums.openssl
 
@@ -35,6 +35,9 @@ M: evp-md-context dispose*
 : set-digest ( name ctx -- )
     handle>> swap digest-named f EVP_DigestInit_ex ssl-error ;
 
+: checksum-method ( data checksum method: ( ctx data -- ctx' ) -- value )
+    [ initialize-checksum-state ] dip '[ swap @ get-checksum ] with-disposal ; inline
+
 PRIVATE>
 
 M: openssl-checksum initialize-checksum-state ( checksum -- evp-md-context )
@@ -50,7 +53,7 @@ M: evp-md-context get-checksum ( ctx -- value )
     memory>byte-array ;
 
 M: openssl-checksum checksum-bytes ( bytes checksum -- value )
-    initialize-checksum-state [ swap add-checksum-bytes get-checksum ] with-disposal ;
+    [ add-checksum-bytes ] checksum-method ;
 
 M: openssl-checksum checksum-stream ( stream checksum -- value )
-    initialize-checksum-state [ swap add-checksum-stream get-checksum ] with-disposal ;
+    [ add-checksum-stream ] checksum-method ;
--- a/basis/checksums/openssl/openssl.factor
+++ b/basis/checksums/openssl/openssl.factor
@@ -1,8 +1,8 @@
-! Copyright (C) 2008, 2010 Slava Pestov
+! Copyright (C) 2008, 2010, 2016 Slava Pestov, Alexander Ilin
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors byte-arrays alien.c-types alien.data kernel
-continuations destructors sequences io openssl openssl.libcrypto
-checksums checksums.stream classes.struct ;
+USING: accessors alien.c-types alien.data checksums
+checksums.stream destructors io kernel openssl openssl.libcrypto
+sequences ;
 IN: checksums.openssl
 
 ERROR: unknown-digest name ;
@@ -17,8 +17,6 @@ INSTANCE: openssl-checksum stream-checksum
 
 C: <openssl-checksum> openssl-checksum
 
-<PRIVATE
-
 TUPLE: evp-md-context < disposable handle ;
 
 : <evp-md-context> ( -- ctx )
@@ -28,8 +26,7 @@ TUPLE: evp-md-context < disposable handle ;
 M: evp-md-context dispose*
     handle>> EVP_MD_CTX_destroy ;
 
-: with-evp-md-context ( quot -- )
-    maybe-init-ssl [ <evp-md-context> ] dip with-disposal ; inline
+<PRIVATE
 
 : digest-named ( name -- md )
     dup EVP_get_digestbyname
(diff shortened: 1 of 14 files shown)


basis/checksums/openssl/openssl.factor
```diff
@@ -38,27 +35,25 @@ M: evp-md-context dispose*
...
-    name>> swap [
-        [
-            [ set-digest ]
-            [ checksum-loop ]
-            [ digest-value ]
-            tri
-        ] with-evp-md-context
-    ] with-input-stream ;
+M: openssl-checksum checksum-stream ( stream checksum -- value )
+    [ add-checksum-stream ] checksum-method ;
```
mrjbq7: This is a little different.  The old way was 4096 byte block sizes, the new way defaults to 65536.  Does that matter?  Can you benchmark it?

Also, should ``checksum-method`` (not sure that's the best name)`` be in ``checksums`` vocabulary so other checksums could use it?

basis/io/directories/directories.factor
```diff
@@ -68,20 +68,12 @@ HOOK: delete-directory io-backend ( path -- )
...
-
-: move-files-into ( files to -- )
-    '[ _ move-file-into ] each ;
-
 ! Copying files
 HOOK: copy-file io-backend ( from to -- )
 
 M: object copy-file
-    make-parent-directories binary <file-writer> [
+    [ normalize-path ] bi@ make-parent-directories
```
mrjbq7: Why are we calling normalize-path here?  We didn't used to call it...

basis/io/directories/directories.factor
```diff
@@ -93,6 +85,18 @@ M: object copy-file
 : copy-files-into ( files to -- )
     '[ _ copy-file-into ] each ;
 
+! Moving and renaming files
+HOOK: move-file io-backend ( from to -- )
+
+M: object move-file ( from to -- )
+    [ normalize-path ] bi@ dupd copy-file delete-file ;
```
mrjbq7: Same here, should be no need for ``normalize-path``.

basis/io/directories/windows/windows.factor
```diff
@@ -36,10 +36,6 @@ M: windows delete-file ( path -- )
     [ (delete-file) ]
     [ \ file-delete-failed boa rethrow ] recover ;
 
-M: windows copy-file ( from to -- )
-    make-parent-directories
-    [ normalize-path ] bi@ 0 CopyFile win32-error=0/f ;
```
mrjbq7: Is our new ``copy-file`` faster than the old one that used ``CopyFile``?  What is the main reason we want to not call ``CopyFile``?

basis/windows/ole32/ole32-tests.factor
```diff
@@ -29,3 +29,7 @@ IN: windows.ole32.tests
 [ "{01234567-89ab-cdef-0123-456789abcdef}" ]
 [ "{01234567-89ab-cdef-0123-456789abcdef}" string>guid guid>string ]
 unit-test
+
+[ f ] [
+    f create-guid 10 [ create-guid [ guid= ] keep -rot or swap ] times drop
```
mrjbq7: It might be better to use ``any?`` so you don't have to do the stack shuffling:

```factor
{ f } [ create-guid 10 iota [ drop create-guid guid= ] with any? ]
```

basis/io/directories/directories.factor
AlexIljin: It was in the Windows copy-file implementation, for which this code is now a replacement.
I don't know if it *should* be there, but it was.

basis/io/directories/directories.factor
AlexIljin: Why not? This method is callable by user, to presumably it should sanitize/preprocess the parameters? Or is there a wrapper that does this?

basis/io/directories/windows/windows.factor
AlexIljin: WinAPI CopyFile is synchronous. 'nuff said.

basis/checksums/openssl/openssl.factor
AlexIljin: I don't think it matters, my main point was to remove custom loop implementation in favor of the common code. I can benchmark.

As to whether to publish the checksum-method, it's your call, but it might be useful if all the checksums implement the common API pattern.

basis/windows/ole32/ole32-tests.factor
AlexIljin: @mrjbq7: that's brilliant, thank you! I still haven't wrapped my head around the `with` construct.

AlexIljin: @mrjbq7: I've updated the patch to use your code.

basis/io/directories/directories.factor
mrjbq7: Because it delegates to copy-file and delete-file, presumably if there is file path sanitization required, those will have to do it also so keep this simple.

AlexIljin: Ah, good point.

basis/checksums/openssl/openssl.factor
```diff
@@ -28,8 +26,7 @@ TUPLE: evp-md-context < disposable handle ;
 M: evp-md-context dispose*
     handle>> EVP_MD_CTX_destroy ;
 
-: with-evp-md-context ( quot -- )
-    maybe-init-ssl [ <evp-md-context> ] dip with-disposal ; inline
+<PRIVATE
```
mrjbq7: why are you making ``evp-md-context`` part of the public interface of ``checksums.openssl``?

basis/io/directories/directories.factor
mrjbq7: it's not necessary here because you are using ``make-parent-directories`` which calls ``make-directory`` which normalizes, and you are using the ``<file-reader>`` and ``<file-writer>`` which should sanitize if necessary any user paths.  so we should remove these calls to ``normalize-path``.

basis/io/directories/directories.factor
```diff
@@ -93,6 +85,18 @@ M: object copy-file
 : copy-files-into ( files to -- )
     '[ _ copy-file-into ] each ;
 
+! Moving and renaming files
+HOOK: move-file io-backend ( from to -- )
+
+M: object move-file ( from to -- )
+    dupd copy-file delete-file ;
```
mrjbq7: Is this implementation here so that your new backend works?

Seems like we should remove it for now since the unix backend and the windows backend already define ``move-file``?

basis/windows/ole32/ole32-tests.factor
```diff
@@ -29,3 +30,7 @@ IN: windows.ole32.tests
 [ "{01234567-89ab-cdef-0123-456789abcdef}" ]
 [ "{01234567-89ab-cdef-0123-456789abcdef}" string>guid guid>string ]
 unit-test
+
+{ f } [
+    create-guid 10 iota [ drop create-guid guid= ] with any?
```
mrjbq7: For what it's worth, another way to do this is make 10 of them and check there are no duplicates:

```factor
{ 0 } [ 10 [ create-guid ] replicate duplicates length ] unit-test
```

basis/checksums/openssl/openssl.factor
AlexIljin: @mrjbq7: "why are you making evp-md-context part of the public interface of checksums.openssl?"

So that `add-checksum-bytes` and `get-checksum` can be called for it.

basis/io/directories/directories.factor
AlexIljin: @mrjbq7: "Is this implementation here so that your new backend works?"

No, I just thought it would be nice to have a generic implementation.

basis/windows/ole32/ole32-tests.factor
AlexIljin: @mrjbq7: I think your new suggestion is even better.

basis/checksums/openssl/openssl.factor
mrjbq7: that's not necessary, you can provide generic methods on "private" classes.

basis/io/directories/directories.factor
mrjbq7: dunno, code we don't use and don't test is weird to have.

AlexIljin: @mrjbq7: "dunno, code we don't use and don't test is weird to have"

I agree, I should revert that.

mrjbq7:
Merged, thanks!

AlexIljin:
Great! And thank _you_ for the added fixes.

PR #1664, opened 2016-06-28
io.launcher.windows: hide window of the launched process

AlexIljin:
As a result of this patch, all the child processes start completely hidden from view: their main windows are not displayed on screen, and they are not shown on the taskbar. This is to avoid flashing of the console windows when a program in Factor calls external utilities.

What do you think? Is there a better way to include this feature? Maybe make it optional?

### DIFF
--- a/basis/io/launcher/launcher-docs.factor
+++ b/basis/io/launcher/launcher-docs.factor
@@ -11,6 +11,9 @@ ARTICLE: "io.launcher.detached" "Running processes in the background"
 "By default, " { $link run-process } " waits for the process to complete. To run a process without waiting for it to finish, set the " { $snippet "detached" } " slot of a " { $link process } ", or use the following word:"
 { $subsections run-detached } ;
 
+ARTICLE: "io.launcher.hidden" "Running hidden processes"
+"By default, child processes can create and display their own (console and other) windows. To signal to a process that it should stay hidden, set the " { $snippet "hidden" } " slot of the " { $link process } " before running it. The processes are free to ignore this signal." ;
+
 ARTICLE: "io.launcher.environment" "Setting environment variables"
 "The " { $snippet "environment" } " slot of a " { $link process } " contains an association mapping environment variable names to values. The interpretation of environment variables is operating system-specific."
 $nl
@@ -338,6 +341,7 @@ ARTICLE: "io.launcher" "Operating system processes"
     "io.launcher.lifecycle"
     "io.launcher.command"
     "io.launcher.detached"
+    "io.launcher.hidden"
     "io.launcher.environment"
     "io.launcher.redirection"
     "io.launcher.priority"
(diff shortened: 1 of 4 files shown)


bjourne:
Isn't that dependent on if you start Factor using factor.com or factor.exe?

erg:
I think it should be an option on ``process`` in "io.launcher".

AlexIljin:
@bjourne: "Isn't that dependent on if you start Factor using factor.com or factor.exe?"

Yes, it is. If you run factor.com, all the child console processes will reuse its window, and there will be no flashing. But if you run factor.exe explicitly because you don't want to see the console window(s), all the child processes will create their console windows anyway, and not even in the background - they will flash in your face.

@erg: "I think it should be an option on process in "io.launcher"."

Could you suggest a name for the slot? `hidden`?

AlexIljin:
OK, I've updated the proposal. Now there is the `hidden` slot. Added documentation.

What I actually needed was that processes spawned by `<process-stream>` were hidden, hence the second patch. What do you say?

bjourne:
Looks good to me. A unit test or two testing the feature would be nice though.

AlexIljin:
@bjourne: "A unit test or two testing the feature would be nice though."

I don't know how to test for the absence of a window. Or did you have something else in mind?

erg:
You could just make sure it still launches with either option and let someone complain is the window shows.

bjourne:
Yep. Didn't mean something fancy. If I understand your pr, you could add something simple like:

    { t } [
        "dir" utf8 (process-stream) hidden>> swap stream-contents drop
    ] unit-test

to `io.launcher.tests`.

AlexIljin:
@erg, @bjourne: OK, I'll do it sometime soon.

AlexIljin:
OK, done.

mrjbq7:
Question: should we default ``hidden`` to true or false?  The new (default) behavior is different than the old (default) behavior.

bjourne:
Probably best to keep it off by default. Otherwise when you run a windowed program like notepad.exe, the window won't show up.

AlexIljin:
@mrjbq7: the new default behavior is only changed for processes started using `(process-stream)`, i.e. by `<process-stream>` and `with-process-stream*`. Only in `(process-stream)` is `hidden` set to `t`, otherwise it defaults to `f` and to the old behavior.

The presumption is that you only do `process-stream` for console processes that work with STDIO streams. That's generally not the case with GUI apps, such as Notepad.

mrjbq7:
That sounds reasonable, I think.  @bjourne?

Merged, thanks!!

bjourne:
Yeah, seems right. It probably should be mentioned in the docs also.

AlexIljin:
@bjourne: added the mention in PR #1714.

@mrjbq7: thanks for merging! Close this PR?

PR #1665, opened 2016-06-28
Win32 error cleanup

AlexIljin:
Hello, dear readers of my full-pull blog!

  In this issue I'd like to unify the throwing of win32 errors in different parts of code, and also use the existing INVALID_HANDLE_VALUE constant instead of the "-1 <alien>" that some people came up with.

  For some reason, applying the last commit in this series leads to gdi+-error being thrown an startup. I, for the life of me, can't track the reason. Can you help me?

### DIFF
--- a/basis/windows/errors/errors.factor
+++ b/basis/windows/errors/errors.factor
@@ -717,9 +717,6 @@ CONSTANT: FORMAT_MESSAGE_MAX_WIDTH_MASK   0x000000FF
     [ drop "Unknown error 0x" id 0xffff,ffff bitand >hex append ]
     [ alien>native-string [ blank? ] trim ] if ;
 
-: win32-error-string ( -- str )
-    GetLastError n>win32-error-string ;
-
 ERROR: windows-error n string ;
 
 : (win32-error) ( n -- )
--- a/basis/windows/errors/errors.factor
+++ b/basis/windows/errors/errors.factor
@@ -744,7 +744,7 @@ ERROR: windows-error n string ;
     win32-error-string throw ;
 
 : check-invalid-handle ( handle -- handle )
-    dup INVALID_HANDLE_VALUE = [ throw-win32-error ] when ;
+    dup INVALID_HANDLE_VALUE = [ win32-error ] when ;
 
 CONSTANT: expected-io-errors
     ${
--- a/basis/windows/errors/errors.factor
+++ b/basis/windows/errors/errors.factor
@@ -719,23 +719,18 @@ CONSTANT: FORMAT_MESSAGE_MAX_WIDTH_MASK   0x000000FF
 
 ERROR: windows-error n string ;
 
-: (win32-error) ( n -- )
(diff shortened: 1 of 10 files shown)


basis/windows/errors/errors.factor
```diff
@@ -758,11 +755,7 @@ CONSTANT: expected-io-errors
     expected-io-errors member? ;
 
 : expected-io-error ( error-code -- )
```
mrjbq7: this is weird that it drops the error code and throws ``GetLastError``.... is that the correct?

AlexIljin:
I have added some more patches, but the code is not ready for review yet. I think I'll rearrange the commits and give better log messages.
I pushed it now just to have the work on GitHub, and so that you could see where I'm going with this.

mrjbq7:
I wonder if you should simplify the words into two layers. One that gets last error and then defers to another word that checks and throws based on the error code.  We seem to have some duplication between ``n>windows-error-check`` and ``windows-error<>0`` for example.

mrjbq7:
Also, not a big fan of ``n>`` as a word prefix, it's a little meaningless :-)

erg:
I think it's pretty descriptive when you constrain ``n`` to always be an integer.

mrjbq7:
What's the status of this PR?  Ready to merge?

AlexIljin:
No, it's on pause. I'll notify when ready.

catb0t:
did you end up not finding these changes useful? maybe if they're merged they can clean up other windows code we now have (well, aside from merging being a lot of work)

AlexIljin:
I kind of lost interest in this work, though hoping to finish it some day.
Some of these changes I think are valuable, like replacing `-1 <alien>` with the `INVALID_HANDLE_VALUE` constant, because they make code more canonical and readable to a WinApi developer.
I think I need someone to help me with this.

AlexIljin:
Mostly I wanted to reduce duplication and make all words for a neat tree.

bjourne:
Looks like fine changes to me. Although some unit tests probably would be useful.

mrjbq7:
@AlexIljin this seems like it's ready to merge as is?

AlexIljin:
@mrjbq7, most of it is OK to merge, but the last patch throws up the `gdi+-error` when you try to `refresh-all`, and I still couldn't figure out why.

mrjbq7:
Maybe it's an issue with ``refresh-all`` and you can still bootstrap into a good VM?

mrjbq7:
Merged!

Let's see what it looks like on the build bots.

mrjbq7:
(For certain changes that affect the UI on macOS and Linux, I sometimes have to do a console listener ``refresh-all save`` then start the UI.  Maybe that's required here.)

AlexIljin:
I did `save` and restarted, then I saw this message on startup.

mrjbq7:
Ok, cool. Will find and fix!

PR #1671, opened 2016-07-03
Listing apache httpd

jonenst:
html table listing + sort specs

Also while doing this, I needed a word to dynamically sort ascending or descending, I think we don't have one yet. Maybe add something like
```
: sort-with* ( seq quot asc? -- sortedseq )
      [ sort-with ] [ [ reverse ] unless ] bi* ; inline
! or <reversed> ??
! or reverse! ?
```

I also tried
```
: sort-with* ( seq quot: ( elt -- key ) asc? -- sortedseq )
    [ sort-with ] [ inv-sort-with ] if ; inline
```
but this generates each call to sort seem to generate 3kb of code so this adds 3kb of code...

Or maybe
```
: sort-with* ( seq quot asc? -- sortedseq ) '[ _ compare _ [ invert-comparison ] unless ] sort ; inline
```
but this does the test at every comparision..

and find a better name than sort-with* :)

### DIFF
--- a/basis/http/server/static/static.factor
+++ b/basis/http/server/static/static.factor
@@ -9,7 +9,7 @@ sorting logging calendar.format accessors splitting io io.files
 io.files.info io.directories io.pathnames io.encodings.binary
 fry xml.entities destructors urls html xml.syntax
 html.templates.fhtml http http.server http.server.responses
-http.server.redirection xml.writer ;
+http.server.redirection xml.writer locals ;
 QUALIFIED: sets
 
 TUPLE: file-responder root hook special index-names allow-listings ;
@@ -58,18 +58,54 @@ TUPLE: file-responder root hook special index-names allow-listings ;
 
 \ serve-file NOTICE add-input-logging
 
+:: file-html-template ( href size modified -- xml )
+    [XML
+        <tr>
+            <td><a href=<-href->><-href-></a></td>
+            <td align="right"><-modified-></td>
+            <td align="right"><-size-></td>
+        </tr>
+    XML] ;
+
 : file>html ( name -- xml )
-    dup link-info directory? [ "/" append ] when
-    dup [XML <li><a href=<->><-></a></li> XML] ;
+    dup link-info [
+        dup directory?
+        [ drop "/" append "-" ]
+        [ size>> number>string ] if
+    ] [ modified>> ] bi file-html-template ;
+
+: parent-dir-link ( -- xml )
+    "../" "" "" file-html-template ;
+
+: ?parent-dir-link ( -- xml/f )
+    url get [ path>> "/" = [ "" ] [ parent-dir-link ] if ] [ "" ] if* ;
+
+: listing-title ( -- title )
+    url get [ path>> "Index of " prepend ] [ "" ] if* ;
+
+:: listing-html-template ( title listing ?parent -- xml )
+    [XML <h1><-title-></h1>
+        <table>
+            <tr>
+                <th>Name</th>
+                <th>Last modified</th>
+                <th>Size</th>
+            </tr>
+            <tr><th colspan="5"><hr/></th></tr>
+            <-?parent->
+            <-listing->
+            <tr><th colspan="5"><hr/></th></tr>
+        </table>
+    XML] ;
+
+: listing ( path -- seq-xml )
+    [ natural-sort [ file>html ] map ] with-directory-files ;
+
+: listing-body ( title path -- xml )
+    listing ?parent-dir-link listing-html-template ;
 
 : directory>html ( path -- xml )
-    [ file-name ]
-    [ drop f ]
-    [
-        [ file-name ] [ [ natural-sort [ file>html ] map ] with-directory-files ] bi
-        [XML <h1><-></h1> <ul><-></ul> XML]
-    ] tri
-    simple-page ;
+    [ listing-title f over ] dip listing-body simple-page ;
 
 : list-directory ( directory -- response )
     file-responder get allow-listings>> [
--- a/basis/http/server/static/static.factor
+++ b/basis/http/server/static/static.factor
@@ -9,7 +9,7 @@ sorting logging calendar.format accessors splitting io io.files
 io.files.info io.directories io.pathnames io.encodings.binary
 fry xml.entities destructors urls html xml.syntax
 html.templates.fhtml http http.server http.server.responses
-http.server.redirection xml.writer locals ;
+http.server.redirection xml.writer locals combinators ;
 QUALIFIED: sets
 
 TUPLE: file-responder root hook special index-names allow-listings ;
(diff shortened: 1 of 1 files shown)


jonenst:
Ooops, broke the tests, fixed now.

jonenst:
Also for sort-with* I didn't consider a dynamic quot and call( X y -- <=> ) becuse I suspect it would be slower. Something like (untested)
```
: sort-with* ( seq quot asc? --  sortedseq )
[ [ compare ] curry ] dip
[ [ invert-comparison ] compose ] unless
[ call( x y -- <=> ) ] curry sort
```

bjourne:
Looks much better than before. You could write your sorting like this:

```factor
: sort-key ( {file,info} -- key )
    sort-column {
        { "M" [ second modified>> ] }
        { "S" [ second size-without-directories ] }
        { "N" [ first ] }
    } case ; inline

: sort-listing ( zipped-files-infos -- sorted )
    [ sort-key ] sort-with sort-asc? [ reverse ] unless ; inline
```

`sort-with*` could be something like:
```factor
: sort-with* ( seq quot: ( elt -- key ) asc? -- seq' )
    swap '[ _ [ swap ] unless _ compare ] sort ; inline
```

jonenst:
Using ```[ sort-key ] sort-with``` is doing the sort-order get and the case at each comparison, which I was trying to avoid. But in my code there is the problem that sort is called 3 times so that's 9kb of inline code.. I guess it's better your way for directory files because they are small lists anyway. Is that what you had in mind ?

Also, you really like it better when the case throws an exception for an invalid column request ?

Finally, regarding sort-with*
> sort-with* could be something like:
```
: sort-with* ( seq quot: ( elt -- key ) asc? -- seq' )
    swap '[ _ [ swap ] unless _ compare ] sort ; inline
```

This also does the test at every comparison, and I don't think it's better than what I wrote
```
: sort-with* ( seq quot asc? -- sortedseq )
'[ _ compare _ [ invert-comparison ] unless ] sort ; inline
```

which uses words from the linear protocol instead of assuming that ```swap <=>``` is equivalent to ```<=> invert-comparison```

bjourne:
Well, my suggestion was just for removing a little code duplication I saw. But it is true that it runs the `case` for each element and therefore is a bit slower.  Another way would be using the "call trick":

```factor
: key-quot ( sort-column -- key-quot )
    {
        { "M" [ second modified>> ] }
        { "S" [ second size-without-directories ] }
    } at [ first ] or ;

: sort-quot ( -- sort-quot )
    sort-column key-quot '[ _ compare ]
    sort-asc? [ \ invert-comparison suffix ] unless ;

: sort-listing ( zipped-files-infos -- sorted )
    sort-quot '[ _ call( obj1 obj2 -- <=> ) ] sort ;
```

jonenst:
> I didn't consider a dynamic quot and call( X y -- <=> ) because I suspect it would be slower.

So I don't think the "call trick" would solve the slowness problem :) And it adds complexity.

I think we can pull this as is, or apply your refactoring. Both are fine. It's not like we're sorting lists of millions of files.


For reference here's a quick benchmark:

 ```factor
SYMBOL: foosort
: a ( arr -- sorted )
foosort get
 {
   { "A" [ [ first ] sort-with ] }
   [ drop [ second ] sort-with ]
} case ; 
: b ( arr -- sorted )
 [ foosort get {
   { "A" [ first ] }
   [ drop second ]
} case ] sort-with ; 
: c ( arr -- sorted )
 foosort get [  {
   { "A" [ first ] }
   [ drop second ]
} case ] curry sort-with ; 
: d ( arr -- sorted )
 foosort get {
   { "A" [ [ first ] ] }
   [ drop [ second ] ]
} case [ compare ] curry [ call( x y -- <=> ) ] curry sort ; 

"B" foosort set
1000000 [ 2 [ random-32 ] replicate ] replicate
gc [ a ] time drop
1000000 [ 2 [ random-32 ] replicate ] replicate
gc [ b ] time drop
1000000 [ 2 [ random-32 ] replicate ] replicate
gc [ c ] time drop
1000000 [ 2 [ random-32 ] replicate ] replicate
gc [ d ] time drop
```
Running time: 1.047180931 seconds
Running time: 1.661074428 seconds
Running time: 1.305837008 seconds
Running time: 3.297530013 seconds

File: basis/http/server/static/static.factor [hunk 1]
```diff
@@ -58,18 +58,98 @@ TUPLE: file-responder root hook special index-names allow-listings ;
...
-    simple-page ;
+        dup directory?
+        [ drop "/" append "-" ]
+        [ size>> number>string ] if
+    ] [ modified>> ] bi file-html-template ;
+
+: parent-dir-link ( -- xml )
+    "../" "" "" file-html-template ;
+
+: ?parent-dir-link ( -- xml/f )
```
mrjbq7: When does this return ``f``? It looks like it returns empty string?

basis/http/server/static/static.factor
```diff
@@ -58,18 +58,98 @@ TUPLE: file-responder root hook special index-names allow-listings ;
...
+
+: sort-listing ( zipped-files-infos -- sorted )
+    sort-column {
+        { "M" [ sort-with-modified ] }
+        { "S" [ sort-with-size ] }
+        [ drop sort-with-name ]
+    } case ; inline
+
+: zip-files-infos ( files -- zipped )
+    dup [ link-info ] map zip ;
```
mrjbq7: Can also do ``[ dup link-info ] { } map>assoc``...

basis/http/server/static/static.factor
```diff
@@ -58,18 +58,98 @@ TUPLE: file-responder root hook special index-names allow-listings ;
...
+                <th><a href=<-CO-S->>Size</a></th>
+            </tr>
+            <tr><th colspan="5"><hr/></th></tr>
+            <-?parent->
+            <-listing->
+            <tr><th colspan="5"><hr/></th></tr>
+        </table>
+    XML] ;
+
+: sort-column ( -- column ) params get "C" of "N" or ;
```
mrjbq7: Do these query parameters match apache? I haven't looked...

basis/http/server/static/static.factor
```diff
@@ -58,18 +58,98 @@ TUPLE: file-responder root hook special index-names allow-listings ;
...
+    [ first ] listing-sort-with ;
+
+: sort-with-modified ( {file,info} -- sorted )
+    [ second modified>> ] listing-sort-with ;
+
+: size-without-directories ( info -- size )
+    dup directory? [ drop -1 ] [ size>> ] if ;
+
+: sort-with-size ( {file,info} -- sorted )
+    [ second size-without-directories ] listing-sort-with ;
```
mrjbq7: Another idea might be to make a tuple that has the ``{ name modified size }`` slots and then use ``sorting.slots`` to do a secondary sort when sorting by size or modified in case of identical size/modified slots.

```{ "size" "name" } sort-by```

```{ "size" "modified" } sort-by```

mrjbq7:
I merged this PR, if you want to address any of the suggestions/comments I made in a second PR that would be fine.

Thanks!!

mrjbq7:
Also, I suspect "sort speed" isn't important for this -- wouldn't the list of files have to be huge before that mattered?

PR #1678, opened 2016-07-25
windows.fonts: call reset-memoized for windows-fonts on startup

AlexIljin:
Please, accept this humble contribution. That's all I have, but I see that you need it more than I do.


mrjbq7:
Merged, thanks.

PR #1680, opened 2016-07-30
NaNs

jonenst:
So I set out to improve factor's printing of NaNs...
Before this PR, all nans prettyprintted as "NAN: payload" (no sign), and `number>string`'d as 0/0.. So for prettyprint, both negative and positive nans with the same payload had the same prettyprinted representation.
Also, entering 0/0. in the listener showed as `NAN: 8000000000000`

At commit b875554ff03, both these issues are fixed (there's a catch, I made the prettyprinter use  `0/0. fp-bitwise=` and I'm not sure if 0/0. is the same value on all platform. So a float might prettyprint as 0/0. on some platform and NAN: xxx on another.. Or the prettyprinted representation "0/0." might unparse to different floats, whereas before NAN: XXX was more precise. Not sure if that's a problem) 

After that, I was thinking that glibc prints nans as "-nan" or "nan", so we could use "0/0." and "-0/0." for number>string. And prettyprint "-0/0." as "-0/0." instead of `NAN: fff8000000000000`
But on my hardware, 0/0. is actually NAN: fff8000000000000 (negative nan) (same for -1 fsqrt ; or other operations with NaNs). So I had to add some code to force the sign of nans in the parse for 0/0. and -0/0.

So depending on what people like, we can merge this PR fully, or just until b875554ff03. What do you think ?

### DIFF
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -83,7 +83,7 @@ M: real pprint*
 
 M: float pprint*
     dup fp-nan? [
-        \ NAN: [ fp-nan-payload >hex text ] pprint-prefix
+        \ NAN: [ double>bits >hex text ] pprint-prefix
     ] [
         number-base get {
             { 10 [ number>string text ] }
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -82,8 +82,8 @@ M: real pprint*
     } case ;
 
 : nan-with-payload? ( f -- ? )
-    [ fp-nan? ] [ 0/0. fp-bitwise= not ]
-    bi and ; inline
+    [ fp-nan? ] [ 0/0. fp-bitwise= not ] [ -0/0. fp-bitwise= not ]
+    tri and and ; inline
 
 M: float pprint*
     dup nan-with-payload? [
--- a/basis/prettyprint/backend/backend.factor
+++ b/basis/prettyprint/backend/backend.factor
@@ -81,8 +81,12 @@ M: real pprint*
         [ unsupported-number-base ]
     } case ;
 
+: nan-with-payload? ( f -- ? )
+    [ fp-nan? ] [ 0/0. fp-bitwise= not ]
+    bi and ; inline
+
 M: float pprint*
-    dup fp-nan? [
+    dup nan-with-payload? [
         \ NAN: [ double>bits >hex text ] pprint-prefix
     ] [
         number-base get {
--- a/basis/prettyprint/prettyprint-tests.factor
+++ b/basis/prettyprint/prettyprint-tests.factor
@@ -20,6 +20,7 @@ IN: prettyprint.tests
 { "0x1.0p3" } [ 16 number-base [ 8.0 unparse ] with-variable ] unit-test
 { "1267650600228229401496703205376" } [ 1 100 shift unparse ] unit-test
 { "NAN: 7ff80000deafbeef" } [ NAN: 7ff80000deafbeef unparse ] unit-test
+{ "0/0." } [ 0/0. unparse ] unit-test
 
 { "+" } [ \ + unparse ] unit-test
 
--- a/basis/prettyprint/prettyprint-tests.factor
+++ b/basis/prettyprint/prettyprint-tests.factor
@@ -21,6 +21,7 @@ IN: prettyprint.tests
 { "1267650600228229401496703205376" } [ 1 100 shift unparse ] unit-test
 { "NAN: 7ff80000deafbeef" } [ NAN: 7ff80000deafbeef unparse ] unit-test
 { "0/0." } [ 0/0. unparse ] unit-test
+{ "-0/0." } [ -0/0. unparse ] unit-test
 
 { "+" } [ \ + unparse ] unit-test
 
--- a/basis/prettyprint/prettyprint-tests.factor
+++ b/basis/prettyprint/prettyprint-tests.factor
@@ -19,6 +19,7 @@ IN: prettyprint.tests
 [ 8 number-base [ 8.0 unparse ] with-variable ] [ unsupported-number-base? ] must-fail-with
 { "0x1.0p3" } [ 16 number-base [ 8.0 unparse ] with-variable ] unit-test
 { "1267650600228229401496703205376" } [ 1 100 shift unparse ] unit-test
+{ "NAN: 7ff80000deafbeef" } [ NAN: 7ff80000deafbeef unparse ] unit-test
 
 { "+" } [ \ + unparse ] unit-test
 
--- a/core/math/parser/parser-tests.factor
+++ b/core/math/parser/parser-tests.factor
@@ -1,4 +1,5 @@
-USING: layouts literals math math.parser sequences tools.test ;
+USING: kernel layouts literals math math.parser
+math.parser.private sequences splitting tools.test ;
 
 { f }
 [ f string>number ]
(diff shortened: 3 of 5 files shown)


bjourne:
It's a very tricky subject but great that someone is spending time with it. I wonder do we need negative NaNs at all? I'm reading this thread (https://bugs.python.org/issue23185) and it seems NaNs, in contrast to positive and negative infinity, can be thought of to be signless. I don't understand why we even have the NAN: 1234... syntax because all NaNs are the same and it doesn't matter what the payload bits are. Like why not use `0/0.` for NaN everywhere?

mrjbq7:
The thought was that two NaN's that are not bit wise equal should look different. I kinda like `0/0.` for all NaN's with maybe throwing `NaN` Into the ring.

mrjbq7:
I'm also confused about the sign of NaN's.  I think they should be unsigned because they aren't numbers.

erg:
Negative nans are subjective, but nan payloads are a parlour trick/compiler optimization opportunity.

Factor doesnt use them, but it could. I believe luajit uses the nan payload for some sort of optimization (google it).

erg:
Literally not a number.

jonenst:
I like "0/0." more than "NaN" for symmetry with "1/0." . I also like that number>string gives a simpler version than the prettyprinter.
I think the sign bit of nans is as important as important as any other bit of the payload. So if we prettyprint nans with their payloads, we should print the sign too.

So I think I'm leaning towards not printing the sign of nans in number>string and printing the full 64 bits of nans in the prettyprinter, except for the one default nan NAN: fff8000000000000 (ie just merge b875554 , not the full PR)

Another point that could be improved is that the signaling bit is more important than the sign bit.
Maybe add something like QNAN: and SNAN: ? And An advantage of the "NaN" syntax over "0/0." is that you can also have "qNaN" and "sNaN". I wonder if we can have both the signaling/quiet information and a syntax like "1/0." ...

mrjbq7:
I kind of think number>string should be consistent with the pretty printer.

mrjbq7:
I'd love to merge these PR's if it makes sense... where were we with them?

jonenst:
For me, the first commit doesn't raise any question. It should go in. There's no reason not to show the sign bit of the nan when already writing the full payload in hex (except backwards compatibility..?)

For the rest, we basically when need to decide things:
- should number>string behave like the glibc and print the negative sign of nans with "-" ?
- should number>string print the signalling bit in a specific manner (like the sign bit with "-". Maybe "0/0!" vs "0/0." ?)
- should number>string print all the payload and sign so that string>number can round trip the exact nan ?
- should number>string switch to NaN instead of 0/0. ? (with possible variants qNaN sNaN -qNaN -sNan ?)

- should the prettyprinter print all NaNs as NAN: XXXX or special case some NaNs (output the litterals from number>string, like 0/0. for example)

- should 0/0. have the same bits on all platforms ?
- should 0/0. be the positive quiet nan on all platforms ?
- should 0. 0. / or -1 fsqrt be number>string'd "-0/0." ? prettyprinted "-0/0." ? 

The current approach is the simplest: number>string always prints 0/0. ; prettyprinter always print NAN: XXXX; 0/0. is platform dependant (negative on x86_64) because it is the same as ```0. 0. /```.

I think we should make 0/0. guaranteed to be the positive quiet nan on all platforms (also guarantee the values of -0/0. and the signalling variants (0/0! and -0/0!)), display the sign and signaling bit in number>string, accept those representations in string>number, and special case the prettying printer so that those nans are printed as their litteral value. This would make factor a very powerful language to deal with floats (it would go well with the math.floats.env which allows a very fined grain control over floats). The cost is extra complexity, people might be surprised when they see negative nans or signaling nans (eg bugs like http://stackoverflow.com/questions/3772835/getting-a-negative-nan-on-g-4-4-3-is-this-standard ).

Another approach is to keep the current simple system. I'm not sure how useful NaNs are. Having factor handle them with the previous approach would be impressive and empowering, but might have a practical negative net effect because more people want simple things.

jonenst:
Also note that the PR in its current state does not do what I proposed. When we agree on what to do, we can write the code :)

mrjbq7:
I love the idea of best-in-class proper NaN handling (qNaN, sNan, etc.).  I'm not sure what the best convention is (`0/0` seems okay), but one thing that I'd like if possible is that visible NaN's have a single bit-representation, or rather that a prettyprinted NaN can be converted back into the exact float that it came from (one reason I like the change to print the payload even though it's a little weird).

mrjbq7:
I suck at your NaN pull requests.  It's hard to remember what we wanted to merge and what was still under discussion.  Sorry!

jonenst:
Ha no worries, deciding things like this is hard ! And not much has been decided yet :)

A) nan printing: need to decide how smart we want to be to represent nans. The simplest/dumbest is to always print NAN: XXX (full 64 bits). The smartest is to represent the "default" nans as some variation of "0/0." or some other literal "NaN" with the sign and the quiet/signaling bit. 

B) if we go for some kind of "[-]0/0." we need to decide if we want the bitwise value to be cross platform or not. For example now "0/0." actually does "0. 0. /f" which uses a floating point division, so it uses the value defined in the hardware of the machine compiling the source code (for x86_64, that's a NaN with a quiet nan with a null payload and a negative sign 0xfff8000000000000). 

For A, I think we should aim for the smartest ! sign and signaling and "nice" default nans. Quiet nans as "[-]0/0" .for signaling nans, I think I like "0./0" the most ? Or "0//0." ? Or "0./0!" ? I don't like "0/0" for signaling NaN because I find it to close to "0 0 /" which is integer math... I also thought of things like "0/0.!" or variations but find them all ugly.

For B, I think we should set them:
0/0. : 0x7ff8000000000000
-0/0. : 0xfff8000000000000
0./0 : 0x7ff4000000000000
-0./0 : 0xfff4000000000000

Also, look at the questions I asked in the previous post, they are all about these 2 subjects.

Cheers,
Jon

mrjbq7:
Merged all the things that make sense in abc67c73d29ab66997b51767f08e44e14f9b1999.

Belated thanks!

jonenst:
Thank you ! Sorry for not being able to "finish" this

mrjbq7:
Do you have any tips on fixing this?

```
IN: scratchpad 0.0 0.0 float/f .
-0/0.
```

jonenst:
I will try to remember what I found when I investigated this. I mentionned this
> "0/0." actually does "0. 0. /f" which uses a floating point division, so it uses the value defined in the hardware of the machine compiling the source code (for x86_64, that's a NaN with a quiet nan with a null payload and a negative sign 0xfff8000000000000).

Will get back to you

mrjbq7:
Okay, workaround here in 187c3240fbac2dd086cd24a89304b52d097731a3.

jonenst:
So yeah, we could choose to display the quiet *negative* nan with 0 payload as "0/0." because its the result of 0. 0. float/f (and also -1 fsqrt and many others) on the most common platform (x86) (not sure about other platforms). But "0/0." is already used by the quiet *positive* nan with 0 payload. So for homoiconicity we should choose a new name for this one, like NAN: 0 or something? That would be a non backwards compatible change.

However I don't know if the benefits are worth the costs. How many people will be tripped up by the fact that they think that "0/0." should be a positive nan? To be compared with the number of people who will be tripped up by the fact that their hardware returns a negative nan for `0. 0. float/f`.

Side note, the way I implemented and what is in current master is that `"0/0." string>number` is still platform-dependant (only the sign bit is forced after the 0. 0. float/f). Maybe we should make it fully portable by not doing `0. 0. float/f` at all and hardcoding the resulting nan (sign, quiet=0, payload=0) ?

PR #1682, opened 2016-08-09
Doc fixes

AlexIljin:
Some documentation fixes and minor improvements.

### DIFF
--- a/core/assocs/assocs-docs.factor
+++ b/core/assocs/assocs-docs.factor
@@ -328,6 +328,8 @@ HELP: assoc-partition
 }
 { $description "Calls a predicate quotation on each key of the input assoc. If the test yields true, the key/value pair is added to " { $snippet "true-assoc" } "; if false, it's added to " { $snippet "false-assoc" } "." } ;
 
+{ assoc-partition partition } related-words
+
 HELP: assoc-any?
 { $values { "assoc" assoc } { "quot" { $quotation ( ... key value -- ... ? ) } } { "?" boolean } }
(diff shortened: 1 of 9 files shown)


core/assocs/assocs-docs.factor
```diff
@@ -328,6 +328,8 @@ HELP: assoc-partition
 }
 { $description "Calls a predicate quotation on each key of the input assoc. If the test yields true, the key/value pair is added to " { $snippet "true-assoc" } "; if false, it's added to " { $snippet "false-assoc" } "." } ;
 
+{ assoc-partition partition } related-words
```
mrjbq7: I'm not sure we should make all the ``assoc-foo`` and ``foo`` words related.  Long term I think we should have a common cursors/containers interface so operating on hashtables and arrays isn't so different.  But maybe this isn't a big deal. Dunno.

AlexIljin: The thing is, I knew about `partition`, and after some thinking I knew I needed something similar for an `assoc`, but I did not know that `assoc-partition` was already there. AFAIR, I found out it exists after asking on the mailing list. That's why I put a link in the docs.

It may be unnecessary even for me, now that I know to look for `assoc-foo` for every `foo`, but in general I prefer seeing more links in the "see also" section rather than fewer.

You may omit this commit if you like it better without it.

AlexIljin:
I've added a few more commits to this PR.

mrjbq7:
Merged, thanks!

I reverted the ``assoc-partition`` related words change for right now.

PR #1683, opened 2016-08-10
ui.gadgets.tables: implement cell-dim and draw-cell for the f class

AlexIljin:
I'd like to propose this update for the table gadget. I was creating a window based on the error-list tool design. I needed to display a sequence of tuples as rows, with some slots as the columns. I found that without this patch if there are `f` objects in the cells, the table will throw

```
Generic word cell-dim does not define a method for the POSTPONE: f class.
Dispatching on object: f
```

and fail to render the world. With this fix it renders the table with some empty cells.

What do you think?

### DIFF
--- a/basis/ui/gadgets/tables/tables.factor
+++ b/basis/ui/gadgets/tables/tables.factor
@@ -67,6 +67,9 @@ rows ;
 GENERIC: cell-dim ( font cell -- width height padding )
 GENERIC: draw-cell ( font cell -- )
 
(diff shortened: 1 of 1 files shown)


mrjbq7:
Seems okay.  Maybe nicer if something else just checked the column as ``f`` and didn't do work, but I see that this is a fairly clean way to do it since the logic is split across dims and drawing.

Merged, thanks!

PR #1685, opened 2016-08-16
windows.com.syntax: remove return-type-name import

AlexIljin:
This fixes the bootstrap, which was broken in 907d63c16b.


mrjbq7:
Merged, thanks!!

PR #1686, opened 2016-08-16
checksums.crc16: new vocabulary

AlexIljin:
I've been working with the [Modbus](https://en.wikipedia.org/wiki/Modbus) protocol recently, and I've added this checksum. It works.


mrjbq7:
Want to do crc64 also? :-)

mrjbq7:
Merged!!

PR #1690, opened 2016-08-17
twitter: Changed twitter-url to use >url

xzenf:
obtain-request-token needs a url and fails with a string.


mrjbq7:
Looks good, merged.   Thanks!!

PR #1696, opened 2016-08-21
Change vm comments

AlexIljin:
All comments I could find were changed from /**/ to //, except for a couple embedded in multiline macros.
I've tested the build, and factor.exe/com/dll/lib are built with no errors or warnings. I could not test other OSes.


bjourne:
Looks pretty nice to me. What do you guys say?

bjourne:
@erg, @mrjbq7? If we want to merge this it is probably better to do it sooner rather than later because it touches so many files.

mrjbq7:
I'm +1, I think @erg liked it and @slavapestov doesn't get a vote right now! :-)

bjourne:
I tried doing the merge with the button :) It didn't turn out right.

mrjbq7:
You can't merge because git.factorcode.org is the true master and we push it to github. So I always have to pull/rebase/merge/push manually.

bjourne:
Ah ok. Doing it that way was a little tricky. Hm..

bjourne:
I pushed it using cherry-pick. Seems to work.

Issue #1697, opened 2016-08-21
Editor gadgets ignore row and column restrictions


masweeney:
Factor 0.98 x86.64 (1781, heads/master-eb8ce16277, Fri Aug 19 12:19:48 2016) (Linux, Debian)
Factor 0.97 x86.64 (stable release) (Windows 10)

It appears that the editor gadget ignores the `min/max-rows/cols` values unless it is contained in a viewport. Otherwise, it  #appears as a 1x14 pixel vertical line, and stretches to fill any text entered into it.

Here is an example:
```
<multiline-editor>
    10 >>min-rows
    20 >>max-rows
    60 >>min-cols 
    120 >>max-cols
    COLOR: blue <solid> >>interior 
gadget.
```

And if we set the text first:
```
<multiline-editor>
    10 >>min-rows
    20 >>max-rows
    5 >>min-cols 
    10 >>max-cols
    COLOR: blue <solid> >>interior 
dup gadget.
"any string longer than the maximum" swap set-editor-string
```

This also happens when using `open-window` instead of `gadget.`, and when manually typing the text instead of setting it.

bjourne:
I'm not sure this is a bug. Those slots are only intended for size calculations when the editor is placed inside a viewport, such as a scroller.  Other gadgets such as tables work the same way. Instead, you can use pref-dim to set a preferred gadget size: `<multiline-editor> { 800 600 } >>pref-dim`

masweeney:
Then would it make sense to move those properties to the viewport instead? I would actually prefer to be able to specify the editor's size based on columns and rows rather than setting the dimension in pixels.

mrjbq7:
In the meantime we could either merge your patch or add an `editor-min-dim` that does the calculation and you can do `dup editor-min-dim >>pref-dim`...?

mrjbq7:
I have to get some time to look at the implications of the patch still.

AlexIljin:
@masweeney: "I would actually prefer to be able to specify the editor's size based on columns and rows rather than setting the dimension in pixels."

How would that work for variable width fonts?

masweeney:
It probably makes the most sense to just move the calculations I put in pref-dim* into a separate function and managing the size with `dup editor-min-dim >>pref-dim` so that it behaves the same way as other gadgets.

@AlexIljin also raises a good point about variable width fonts (though it makes me wonder what the viewport does about that).

Thanks for the suggestions!

masweeney:
[...]

PR #1698, opened 2016-08-22
Editor sizing

masweeney:
This fixes issue #1697, so that the editor gadget now respects the min/max-row/col parameters when calculating its preferred size.

P.S. new to `git` - let me know if I screwed this up

### DIFF
--- a/basis/ui/gadgets/editors/editors.factor
+++ b/basis/ui/gadgets/editors/editors.factor
@@ -2,7 +2,7 @@
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors arrays assocs calendar colors.constants
 combinators combinators.short-circuit documents
-documents.elements fry grouping kernel locals make math
+documents.elements fry grouping kernel strings locals make math
 math.functions math.order math.ranges math.rectangles
 math.vectors models models.arrow namespaces opengl sequences
 sorting splitting timers ui.baseline-alignment ui.clipboards
--- a/basis/ui/gadgets/editors/editors.factor
+++ b/basis/ui/gadgets/editors/editors.factor
@@ -213,9 +213,24 @@ M: editor draw-gadget*
         [ draw-lines ] [ draw-caret ] bi
     ] with-variable ;
 
+: dim-by-minmax ( editor -- dim ) 
+    [ font>> ] keep
+        [ dup max-rows>> "" resize-string [ drop CHAR: \n ] map
+            swap max-cols>> "" resize-string [ drop CHAR: x ] map append text-dim ]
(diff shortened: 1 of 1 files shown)


AlexIljin:
I'll comment on the git log messages. The accepted template is:

```
<vocab(s)>: <short-summary>

[long-summary]
```

All lines should not exceed 75 characters. For example: `ui.gadgets.editors: editor pref-dim* uses max/min-row/column preferences`

Another note: no need to include the merge commit in the pull request. It's up to the receiving repository to decide whether to pull as merge or as rebase.

Finally, all these three commits could be squashed together, if there is no value in them being separate.

basis/ui/gadgets/editors/editors.factor
```diff
@@ -213,9 +213,37 @@ M: editor draw-gadget*
         [ draw-lines ] [ draw-caret ] bi
     ] with-variable ;
 
+<PRIVATE
+
+: row,col-dim ( font r c -- dim )
+    "" resize-string [ drop CHAR: x ] map
```
mrjbq7: This should just be `CHAR: x <string>`.

basis/ui/gadgets/editors/editors.factor
```diff
@@ -213,9 +213,37 @@ M: editor draw-gadget*
...
+    swap "" resize-string [ drop CHAR: \n ] map 
+    append text-dim ;
+
+: min-dim ( font editor -- dim )
+    [ min-rows>> 0 [ and ] most ]
```
mrjbq7: Could do `[ 0 ] unless`, I think.

basis/ui/gadgets/editors/editors.factor
```diff
@@ -213,9 +213,37 @@ M: editor draw-gadget*
...
+    
+PRIVATE>
+
 M: editor pref-dim*
     ! Add some space for the caret.
-    [ font>> ] [ control-value ] bi text-dim { 1 0 } v+ ;
+    ! Now respects min/max-row/col parameters
+    [ font>> ] keep
+    [ max-dim ]
+    [ txt-dim ]
```
mrjbq7: I realize you are lining up with min and max but we try not to do too much abbreviation.

basis/ui/gadgets/editors/editors.factor
AlexIljin: Yes, but with a star: `[ 0 ] unless*`.

masweeney:
Thank you for the feedback. I submitted a different pull request (#1699) with these changes, in addition to those suggested in issue #1697.

PR #1699, opened 2016-08-22
ui.gadgets.editors: add editor-constrained-dim function to editor

masweeney:
This adds a function to calculate the dimensions of an editor with respect to the min/max-row/col values as discussed in issue #1697.


mrjbq7:
Merged, thanks!!

PR #1700, opened 2016-08-22
Editor Sizing

masweeney:
Hopefully the last pull request for issue #1697. The first commit shortens one of the helper functions for computing the dimension, fixes an error in computing the height. The second replaces the old `pref-dim*` with `editor-constrained-dim` so that by default editors obey their size restrictions - if it is decided that the original behavior is preferred, then this commit should be ignored.

### DIFF
--- a/basis/ui/gadgets/editors/editors.factor
+++ b/basis/ui/gadgets/editors/editors.factor
@@ -215,25 +215,22 @@ M: editor draw-gadget*
 
 <PRIVATE
 
-: row,col-dim ( font r c -- dim )
-    CHAR: x <string> swap
-    CHAR: \n <string>
-    append text-dim ;
+: col,row-dim ( font c r -- dim )
+    2array swap "m" text-dim v* ;
 
 : min-dim ( font editor -- dim )
-    [ min-rows>> [ 0 ] unless* ]
-    [ min-cols>> [ 0 ] unless* ] bi
-    row,col-dim ;
-
+    [ min-cols>> [ 0 ] unless* ]
+    [ min-rows>> [ 0 ] unless* ] bi
+    col,row-dim ;
+    
 : max-dim ( font editor -- dim )
-    ! hopefully no one goes over 5000
-    [ max-rows>> [ 5000 ] unless* ]
-    [ max-cols>> [ 5000 ] unless* ] bi
-    row,col-dim ;
-
+    [ max-cols>> [ 1/0. ] unless* ]
+    [ max-rows>> [ 1/0. ] unless* ] bi
+    col,row-dim ;
+    
 : txt-dim ( font editor -- dim )
     control-value text-dim ;
-
+    
 PRIVATE>
 
 : editor-constrained-dim ( editor -- dim )
--- a/basis/ui/gadgets/editors/editors.factor
+++ b/basis/ui/gadgets/editors/editors.factor
@@ -242,7 +242,7 @@ PRIVATE>
 
 M: editor pref-dim*
     ! Add some space for the caret.
-    [ font>> ] [ control-value ] bi text-dim { 1 0 } v+ ;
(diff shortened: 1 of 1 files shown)


mrjbq7:
I can see why using ``>>pref-dim`` didn't work.  I had forgotten about the dimension layout caching.

Okay, but I think I've also changed my thoughts on this pull request.  Because ``min-rows`` and ``max-rows`` are in ``line-gadget``, which already has ``line-gadget-height`` and ``line-gadget-width``, which are used in the ``pref-viewport-dim``.

This code is duplicative of that, and not using any slots that are ``editor`` specific.

The real question seems to be: should editors have a preferred dimension restricted separate from inside of a viewport?  Right now the preferred dimension is the size to display the contents.  The viewport dimension is an artificial ``"m"`` based content restriction based on ``min/max-rows`` and ``min/max-cols``.

I'm leaning towards "no", because I think it's a misuse of the meaning of preferred dimension to separate it from the contents the editor is holding (even if you intend to return a value larger because of some maximum theoretical size).

Could you explain your use-case again?  You have a ``multiline-editor`` in a window with some min and max values set and you want the window to contain the editor without using a ``scroller``?

Maybe I would suggest temporarily using your own subclass?

```factor
TUPLE: constrained-editor < multiline-editor ;

M: constrained-editor pref-dim* pref-viewport-dim ;
```

mrjbq7:
Alternatively, we can see that ``pref-viewport-dim`` was implemented just for ``scroller``, and only is implemented for ``line-gadget``, so we could take the view that we should push that logic "up" to ``pref-dim`` (and remove ``pref-viewport-dim`` completely) and I might not be opposed to that.

masweeney:
The use case: I wanted to be able to place some `multiline editor` in a pile layout with fixed widths, but the ability to resize vertically as more text is entered. Right now, this isn't possible; they will have dimension `{ 1 14 }` unless inside a scroller or viewport

I agree that the preferred size should depend on the contents of the editor, and it will when none of these constraints are set, or when the text occupies space between them. However, many text boxes occupy nonzero space even when nonempty, which is what I initially expected the constraints to be used for.

On a different note: if these properties are only used for computing the size of the viewport, should they be moved up such that `viewport` has the `min/max-row/col` properties instead of the line-gadget? 

You are right that this does duplicate `pref-viewport-dim*`, so in any case the current `editor-constrained-dim` can be removed or replaced with that.

mrjbq7:
Do you think we should make a patch like this:

```factor
M: line-gadget pref-dim* pref-viewport-dim

M: editor pref-dim*
    [ call-next-method ] keep ! at least as big as our min-rows/min-cols
    ! Add some space for the caret.
    [ font>> ] keep dup draw-default-text?
    [ default-text>> ] [ control-value ] if
    text-dim { 1 0 } v+ vmax ;
```

And maybe make sure negative min/max-rows and min/max-cols don't affect things:

```diff
diff --git a/basis/ui/gadgets/line-support/line-support.factor b/basis/ui/gadgets/line-support/line-support.factor
index b1cab467c1..98e913d3ff 100644
--- a/basis/ui/gadgets/line-support/line-support.factor
+++ b/basis/ui/gadgets/line-support/line-support.factor
@@ -85,7 +85,7 @@ GENERIC: draw-line ( line index gadget -- )
 <PRIVATE
 
 : clamp ( dim unit min max -- dim' )
-    [ -1/0. or * ] [ 1/0. or * ] bi-curry* bi
+    [ -1/0. or 1 max * ] [ 1/0. or 1 max * ] bi-curry* bi
     [ max ] [ min ] bi* ;
 
 : em ( font -- x ) "m" text-width ;
```

mrjbq7:
Fixed in 6f34ec8d727db8b9072d0d730dc2daa0543151fd.

Issue #1701, opened 2016-08-23
Path joining should use the native separator


bjourne:
It would be nice if the native path separator was used for path joining, for Windows sake:
```factor
"C:\\tmp" "foo\\bar" append-path .
"C:\\tmp/foo\\bar"
```
It's a purely cosmetic issue because forward slashes work fine on Windows, but here `"C:\\tmp\\foo\\bar"` would look better.

mrjbq7:
Fixed by #1702 (b477c432932784e9bb8f62d804b9135ce2c694db).

bjourne:
Hopefully, someone are not using append-path for url manipulation. E.g. `"http://www.google.com/" "blaha" append-path` wont work anymore... I didn't think of that. Of course we have `url-append-path` for those cases..

PR #1702, opened 2016-08-23
io.pathnames: fix separator when appending paths in Windows

masweeney:
This fixes issue #1701.

### DIFF
--- a/core/io/pathnames/pathnames-tests.factor
+++ b/core/io/pathnames/pathnames-tests.factor
@@ -107,14 +107,28 @@ os windows?
 ] with-test-file
 
 ! aum's bug
-H{
-    { current-directory "." }
-    { "resource-path" ".." }
-} [
-    [ "..\\core\\bootstrap\\stage2.factor" ]
-    [ "resource:core\\bootstrap\\stage2.factor" absolute-path ]
-    unit-test
-] with-variables
+os windows?
+    [
+        H{
+            { current-directory "." }
+            { "resource-path" ".." }
+        } [
+            [ "..\\core\\bootstrap\\stage2.factor" ]
+            [ "resource:core\\bootstrap\\stage2.factor" absolute-path ]
+            unit-test
+        ] with-variables
+    ]
+    [
+        H{
+            { current-directory "." }
+            { "resource-path" ".." }
+        } [
+            [ "../core/bootstrap/stage2.factor" ]
+            [ "resource:core/bootstrap/stage2.factor" absolute-path ]
+            unit-test
+        ] with-variables
+    ]
+if
 
 { t } [ cwd "misc" resource-path [ ] with-directory cwd = ] unit-test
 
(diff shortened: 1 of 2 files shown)


mrjbq7:
Merged, thanks!!

mrjbq7:
Reverted, this breaks test cases.

masweeney:
Right now I am modifying the tests to call `normalize-path` on both paths and check for equality. This is pretty inelegant, so let me know if you have a better idea for getting the tests to work more consistently across platforms.

mrjbq7:
I'd have to look at it. Elegant is better than inelegant. Having normalize path equality doesn't check the intermediate result.

masweeney:
Unfortunately, test cases like `"/usr" "lib" append-path` don't make much sense on windows anyway. The updated version returns `/usr\\lib`, which I suppose is correct given the input, but not very meaningful. To check the intermediate, each test could branch on OS and check for equality separately, which might be the best option.

mrjbq7:
I could be find to have something like:

```factor
os windows? [
    { "C:\\foo" } [ "C:" "foo" append-path ] unit-test
] [
    { "/usr/foo" } [ "/usr" "foo" append-path ] unit-test
] if
```

bjourne:
I say good work! You can also use the literal syntax:
```factor
${ "/usr" "foo" path-separator glue } [ "/usr" "foo" append-path ]  unit-test
``` 
Also updating the tests and documentation is really important.

masweeney:
I've updated the tests, so hopefully everything works now.

I was not able to run this test on windows because it's an older version, but everything else worked.
```
[| path |
    { } [ 2 [ path touch-file ] times ] unit-test
] with-test-file
```

bpollack:
Okay @masweeney, sorry for the lag; tests all look good both on Windows and Linux, so I'm going to go ahead and merge this if you're okay with that. I am going to collapse your two commits into just one, though, since it seems to me that your second commit is strictly cleaning up behavior that should've been in the first patch.

mrjbq7:
Do we care what happens with mixed separators?  (Meaning the paths have mixed separators and stuff still works).

Currently:

```factor
IN: scratchpad "C:\\Users" "John" append-path .
"C:\\Users/John"
```

Proposed (note, I changed the arguments a little bit):

```factor
IN: scratchpad "C:/Users" "John" append-path .
"C:/Users\\John"
```

AlexIljin:
@mrjbq7: I think the mixed separators are not an issue of this PR. We had them before, and we have them still. This PR is a step in the right direction, so I like it.

When I want my paths fully compliant, I use `normalize-path`, which switches all instances of `/` to `\` on Windows. After that the proposed `append-path` won't break introduce new `/`s, so that's good. That's my view on the matter.

bpollack:
This landed in 35f82c90f3b22457b42143143bc08d5298462ca6; I'm going to close it.

PR #1712, opened 2016-09-14
vocabs.loader: prevent vocab-exists? from throwing bad-vocab-name

### LINKED ISSUE #1711: Listener stuck with bad-vocab-name error
Start the Listener and type `USE: aoeu/u`. After that you'll get the following error open in a separate window: `Vocab name cannot contain ":/\ "`, and it'll be nearly impossible to type anything in the Listener. The only two workarounds I found are

- to select a piece of the vocab name and press Spacebar, and
- click the Restart Listener in the toolbar.

AlexIljin:
This fixes #1711.

### DIFF
--- a/core/vocabs/loader/loader.factor
+++ b/core/vocabs/loader/loader.factor
@@ -56,7 +56,8 @@ PRIVATE>
     ] cache ;
 
 : vocab-exists? ( name -- ? )
-    dup lookup-vocab [ ] [ find-vocab-root ] ?if ;
+    [ dup lookup-vocab [ ] [ find-vocab-root ] ?if ]
+    [ dup bad-vocab-name? [ 2drop f ] [ rethrow ] if ] recover ;
 
 : vocab-append-path ( vocab path -- newpath )
     swap find-vocab-root [ prepend-path ] [ drop f ] if* ;


core/vocabs/loader/loader.factor
```diff
@@ -56,7 +56,8 @@ PRIVATE>
     ] cache ;
 
 : vocab-exists? ( name -- ? )
-    dup lookup-vocab [ ] [ find-vocab-root ] ?if ;
+    [ dup lookup-vocab [ ] [ find-vocab-root ] ?if ]
+    [ dup bad-vocab-name? [ 2drop f ] [ rethrow ] if ] recover ;
```
mrjbq7: I'd prefer not to use ``recover`` here.  It looks like ``find-vocab-root`` calls ``vocab-name``, which throws.  Perhaps we make a ``?vocab-name`` which returns ``name/f`` and then ``find-vocab-root`` can skip the finding work if not found?

Other suggestions?

mrjbq7:
I accidentally started a review, some new Github feature.  Anyway, I think my comment still shows up above.

core/vocabs/loader/loader.factor
AlexIljin: Actually, ` lookup-vocab` also calls `vocab-name`, so they both throw here.

If you see usage of `vocab-exists?`, there are only two places. Both can use the fix, and it seems consistent with the word name: an invalid vocab name is a just a special case of a non-existing vocab, so it should just return `f`. We are not obliged to throw reasons _why_ the vocab doesn't exist.

bjourne: You could perhaps put the recover wrapping in `(word-at-caret)` instead.

mrjbq7:
Okay, for now I changed ``(word-at-caret)`` to recover and return ``f`` so your UI issue is fixed.

PR #1713, opened 2016-09-18
Checksums.process

AlexIljin:
I've created this new vocab specifically to call external console checksum utilities, such as those from the GNU CoreUtils package. It works! : )
Today I finally had time to write good documentation.

I put it in the `extra` folder for your consideration. Please, provide feedback, if you have any.

### DIFF
--- a/extra/checksums/process/process.factor
+++ b/extra/checksums/process/process.factor
@@ -9,23 +9,23 @@ TUPLE: checksum-process launch-desc ;
 INSTANCE: checksum-process block-checksum
 C: <checksum-process> checksum-process
 
-TUPLE: process-state < disposable proc result ;
+TUPLE: process-state < disposable process result ;
 
 M: checksum-process initialize-checksum-state ( checksum -- checksum-state )
-    launch-desc>> binary <process-stream> process-state new-disposable swap >>proc ;
+    launch-desc>> binary <process-stream> process-state new-disposable swap >>process ;
 
 M: process-state dispose* ( process-state -- )
-    proc>> [ dispose ] when* ;
+    process>> [ dispose ] when* ;
 
 M: process-state add-checksum-bytes ( process-state bytes -- process-state' )
-    over proc>> stream-write ;
+    over process>> stream-write ;
 
 : trim-hash ( str -- str' )
     dup empty? [ dup [ blank? ] find drop [ head ] when* ] unless ;
 
 M: process-state get-checksum ( checksum-state -- value )
     dup result>> [
-        dup proc>> [
+        dup process>> [
             [
                 [ out>> dispose ] keep
                 stream-contents trim-hash hex-string>bytes
--- a/extra/checksums/process/process.factor
+++ b/extra/checksums/process/process.factor
@@ -1,8 +1,8 @@
 ! Copyright (C) 2016 Alexander Ilin.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors checksums checksums.common combinators destructors
-io io.encodings.binary io.launcher kernel math.parser sequences
-strings ;
+USING: accessors ascii checksums checksums.common combinators
+destructors io io.encodings.binary io.launcher kernel math.parser
+sequences strings ;
 IN: checksums.process
 
 TUPLE: checksum-process launch-desc ;
@@ -20,14 +20,15 @@ M: process-state dispose* ( process-state -- )
 M: process-state add-checksum-bytes ( process-state bytes -- process-state' )
     over proc>> stream-write ;
 
-: trim-hash ( str -- str' ) dup " *-" swap start head ;
+: trim-hash ( str -- str' )
+    dup empty? [ dup [ blank? ] find drop [ head ] when* ] unless ;
 
 M: process-state get-checksum ( checksum-state -- value )
     dup result>> [
         dup proc>> [
             [
                 [ out>> dispose ] keep
-                stream-contents >string trim-hash hex-string>bytes
+                stream-contents trim-hash hex-string>bytes
             ] with-disposal
         ] [ B{ } clone ] if*
         [ >>result ] keep
--- a/extra/checksums/process/process.factor
+++ b/extra/checksums/process/process.factor
@@ -1,0 +1,34 @@
+! Copyright (C) 2016 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: accessors checksums checksums.common combinators destructors
+io io.encodings.binary io.launcher kernel math.parser sequences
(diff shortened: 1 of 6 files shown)


AlexIljin:
This PR works best with #1664: that way your console hash processes don't flash in you face, stealing focus from whatever else you are doing.

extra/checksums/process/process.factor
```diff
@@ -0,0 +1,34 @@
...
+M: process-state add-checksum-bytes ( process-state bytes -- process-state' )
+    over proc>> stream-write ;
+
+: trim-hash ( str -- str' ) dup " *-" swap start head ;
+
+M: process-state get-checksum ( checksum-state -- value )
+    dup result>> [
+        dup proc>> [
+            [
+                [ out>> dispose ] keep
```
mrjbq7: Why do you dispose the ``output-stream`` here like this?

extra/checksums/process/process.factor
```diff
@@ -0,0 +1,34 @@
...
+                stream-contents >string trim-hash hex-string>bytes
```
mrjbq7: Why do you convert the ``stream-contents`` to a string before trimming?  Can't you just treat it as a byte-array of ascii bytes?

extra/checksums/process/process.factor
```diff
@@ -0,0 +1,34 @@
...
+            ] with-disposal
+        ] [ B{ } clone ] if*
```
mrjbq7: Maybe this should be an error instead of returning an empty byte array?

extra/checksums/process/process.factor
AlexIljin: This signals end of STDIN stream to the app, and makes it produce the output.

extra/checksums/process/process.factor
AlexIljin: Only because I want to use hex-string>bytes on it.

extra/checksums/process/process.factor
AlexIljin: Could be. Do you think it would be better to throw something?
To my mind, this is the case of a hashing-state before the hashing got started. No hashing - empty result. As good as any. Didn't want to complicate things by adding exceptions.
Other errors may also produce empty result, e.g. if the console app returned nothing:

```
"" hex-string>bytes .
B{ }
```

So, it's a generic empty result, without interrupting the control flow.

AlexIljin: I'm open to suggestions. Teach me the ways of the Factor. : )

AlexIljin: Incidentally, why is this not throwing an error?

```
"aoeu" hex-string>bytes .
B{ 184 254 }
```

extra/checksums/process/process.factor
```diff
@@ -0,0 +1,34 @@
...
+M: checksum-process initialize-checksum-state ( checksum -- checksum-state )
+    launch-desc>> binary <process-stream> process-state new-disposable swap >>proc ;
+
+M: process-state dispose* ( process-state -- )
+    proc>> [ dispose ] when* ;
+
+M: process-state add-checksum-bytes ( process-state bytes -- process-state' )
+    over proc>> stream-write ;
+
+: trim-hash ( str -- str' ) dup " *-" swap start head ;
```
jonenst: In gnu coreutils, the "*" means binary mode and "-" means the input comes from stdin.

But maybe it's better to just look for the first non hexadecimal character [^0-9a-fA-F] so that more tools work with it ? Or maybe simply the first ```blank?``` character ?

extra/checksums/process/process.factor
mrjbq7: Partly because we use ``digit>``, which just assumes offset from ``a``:

```factor
IN: scratchpad CHAR: o digit> .
24
```

Rather than aborting on non-hex values:

```factor
IN: scratchpad "aeou" hex> .
f
```

Maybe that should throw or return ``f``...

extra/checksums/process/process.factor
mrjbq7: I think ``hex-string>bytes`` will work on byte-arrays:

```factor
IN: scratchpad "1234" hex-string>bytes .
B{ 18 52 }

IN: scratchpad "1234" >byte-array hex-string>bytes .
B{ 18 52 }
```

mrjbq7: This new GitHub review button is a little hard to get used to.

extra/checksums/process/process.factor
```diff
@@ -0,0 +1,34 @@
...
+USING: accessors checksums checksums.common combinators destructors
+io io.encodings.binary io.launcher kernel math.parser sequences
+strings ;
+IN: checksums.process
+
+TUPLE: checksum-process launch-desc ;
+INSTANCE: checksum-process block-checksum
+C: <checksum-process> checksum-process
+
+TUPLE: process-state < disposable proc result ;
```
AlexIljin: Done in the updated PR.

extra/checksums/process/process.factor
```diff
@@ -20,14 +20,15 @@ M: process-state dispose* ( process-state -- )
 M: process-state add-checksum-bytes ( process-state bytes -- process-state' )
     over proc>> stream-write ;
 
-: trim-hash ( str -- str' ) dup " *-" swap start head ;
+: trim-hash ( str -- str' )
+    dup empty? [ dup [ blank? ] find drop [ head ] when* ] unless ;
```
AlexIljin: @jonenst : I've reimplemented `trim-hash` using `blank?`. If there is a simpler way, please advise. I did not find a way to use `unless-empty` here.

extra/checksums/process/process.factor
AlexIljin: I've removed `>string` in the updated PR. It still works, @mrjbq7 were right.

extra/checksums/process/process.factor
```diff
@@ -0,0 +1,35 @@
...
+    launch-desc>> binary <process-stream> process-state new-disposable swap >>process ;
+
+M: process-state dispose* ( process-state -- )
+    process>> [ dispose ] when* ;
+
+M: process-state add-checksum-bytes ( process-state bytes -- process-state' )
+    over process>> stream-write ;
+
+: trim-hash ( str -- str' )
+    dup empty? [ dup [ blank? ] find drop [ head ] when* ] unless ;
```
mrjbq7: Sometimes I find it cleaner to not check for empty, this would work without it:

```factor
:  trim-hash ( str -- str' )
    dup [ blank? ] find drop [ head ] when* ;
```

mrjbq7:
Merged, thanks!

PR #1714, opened 2016-09-20
Docs updates

AlexIljin:


mrjbq7:
Merged, thanks!

I'm not sure we want the docs for ``help`` to pull in ``help.lint``, but I'm going to bootstrap with your changes and see if thats a problem.  If it is, we can soft-link to the help-lint docs.

Issue #1281, opened 2015-02-08
broken readln in UI listener


mrjbq7:
From the mailing list:

> How to read multiple lines from standard input (in Listener)?
>
> I tried to write something like "readln readln", then copy-paste two lines of text to Listener, but I got an unexpected result. I got first line on stack, then program waited for my input. I pressed the "Enter" key and, of course, got empty string on stack. So, I got "Line1" "" instead of "Line1" "Line2". How to get both lines? For example, "Line1\nLine2" or something else. Maybe I chose the wrong input method?

bjourne:
Works in the console and in FUEL, so the UI must be doing it wrong.

mrjbq7:
Yep, I think it's something to do with ``M\ interactor stream-readln``.

erg:
It's that ``interactor-read`` calls ``interactor-finish`` which calls ``clear-editor``. Instead, clear-editor should be called outside of the read code.

I don't know why ``ctrl-d`` makes it always return ``f`` though.

```factor
: interactor-yield ( interactor -- obj )
    dup thread>> self eq? [
        {
            [ t >>waiting drop ]
            [ flag>> raise-flag ]
            [ mailbox>> mailbox-get ]
            [ f >>waiting drop ]
        } cleave
    ] [ drop f ] if ;

: interactor-finish ( interactor -- )
    [ history>> history-add ] keep
    [ print-input ]
    [ clear-editor drop ]
    [ model>> clear-undo drop ] 2tri ;

: interactor-read ( interactor -- lines )
    [ interactor-yield ] [ interactor-finish ] bi ;
```

mrjbq7:
Somehow clear should happen after all the pending code is executed, not as each ``stream-read`` or ``stream-readln`` is called I would guess?

erg:
Here's a good demo of the problem:

``readln readln``

<img width="265" alt="Screen Shot 2021-12-21 at 11 46 38 AM" src="https://user-images.githubusercontent.com/27430/146975285-ea12a3f6-9a8d-4448-afd0-8930e7399ca9.png">

Capital-EX:
Looking into it it seems like `stream-readln` will read all lines at once. I can confirm this by modifying `M: interactor stream-readln` to the following:
```
M: interactor stream-readln
    interactor-read dup . ?first ;
```
When I run a single `readln`, all lines I C-c/C-v will be consumed:
[...]

PR #1724, opened 2016-10-09
interactor: fix issues with readln

viswans83:
This fixes one part of the issue #1281 where multiple `readln` discarded available user input. One drawback to this fix is that it introduces a change in behavior on the listener for the case where there are fewer `readln` invocations on the `interactor` being evaluated than the number of lines pasted into the listener. This behavior is similar to how `decoder` works outside the UI.

E.g. Enter on the listener `readln readln`, then paste the below input:

```
Line1
Line2
Line3
```

Then will cause an error to appear soon after due to the remaining extra input:

```
1: Line3
       ^
No word named “Line3” found in current vocabulary search path
```

### DIFF
--- a/basis/documents/documents.factor
+++ b/basis/documents/documents.factor
@@ -72,6 +72,9 @@ CONSTANT: doc-start { 0 0 }
 : doc-end ( document -- loc )
     [ last-line# ] keep line-end ;
 
+: doc-empty? ( document -- ? )
+    doc-end doc-start = ;
+
 <PRIVATE
 
 : (doc-range) ( from to line# document -- slice )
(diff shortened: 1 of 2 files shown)


viswans83:
The last commit fixes how the `interactor` handles `C+d` as well.

mrjbq7:
Very cool!  I'm going to try and get some time to play around with this patch later today or tomorrow.

viswans83:
@mrjbq7, would you like me to investigate any areas relating to this patch?

mrjbq7:
Hi @viswans83!  No, I just wanted some time to think about the change and whether it has any disadvantages besides the "Line3" example.

mrjbq7:
It looks like this adds history elements incorrectly (due to where it calls ``interactor-finish``, probably).

Try this:

```
IN: scratchpad readln readln readln
1
2
3
```

Then if you type ``Ctrl-R`` to get a history popup, the last three entries are ``3``, ``2``, and ``1``.

mrjbq7:
Also, I think this doesn't fix ``read1 read1 read1`` in the listener.

PR #1730, opened 2016-10-24
json: error on invalid json object keys

viswans83:


jonenst:
It's nice to accept slightly invalid inputs in some situation, so maybe we don't want to apply this patch as is. Maybe an option to configure this behavior ?
That being said, python throws on 
```python
>>> json.loads('{3:2}')
ValueError: Expecting property name: line 1 column 2 (char 1)
```

So I guess it's ok.
Also it only works for numbers..
"{foo:2}" json>
json-error

viswans83:
> It's nice to accept slightly invalid inputs in some situation, so maybe we don't want to apply this patch as is. Maybe an option to configure this behavior ?

I'm not really sure if there will be too many uses to allow of the configuration to accept non-literal-string object keys. Most JSON in the wild will use strings as keys.

> Also it only works for numbers..

Actually, we get `json-error` when the literal starts with t, f and n where the `scan` word expects true, false and null respectively. So the following inputs give `json-error`:
 - `'{tam:10}'`
 - `'{fam:10}'`
 - `'{none:10}'`

But, the following inputs give `not-a-json-object-key`:
 - `'{a:1}'`
 - `'{blue:20}'`

The current parser implementation does not do a lot of validity checks. So in this case, it might make more sense to simply throw `json-error` and not the new error `not-a-json-object-key` I introduced. I plan to do some more work to make the parsing more strict, which might resolve these inputs as well.

viswans83:
I removed the `not-a-json-number` and `not-a-json-object-key` errors for now since they were not being thrown appropriately. Work is needed to throw more appropriate errors.

mrjbq7:
Glad you're cleaning it up!

Keep an eye on the benchmark performance as/when you make major parsing changes, mostly so we know what the impact is.

```factor
USE: benchmark.json
gc [ json-benchmark ] time
```

mrjbq7:
Was there more work you wanted to do as part of this PR?  I was waiting before reviewing...

viswans83:
Yes, let me close off this pull request without merging and I can open another when I think its more complete.

PR #1735, opened 2016-10-28
docs: typos: occurence -> occurrence

jonenst:
Didn't fix the one at extra/webapps/fjsc/www/termlib/readme.txt
because it's from upstream


mrjbq7:
Good find! Merged, thanks!

PR #1737, opened 2016-10-29
ODBC

AlexIljin:
The `odbc` vocab now compiles, and possibly even works fine, but I have no experience with ODBC to test correctly, so I don't know if my tests are bad or the vocab contains errors. Experienced people wanted!

  Here's how I tried to test it:

```
"select * from authors.txt;"
"Driver={Microsoft Text Driver (*.txt; *.csv)};Dbq='C:\\Programs\\Dev\\factor.git\\extra\\odbc\\';Extensions=asc,csv,tab,txt;"
odbc-query
```

I have copied the DSN string from http://www.htmlgoodies.com/primers/database/connect-to-csv-data-files-using-microsoft-data-access-components.html.


mrjbq7:
Cool!  I guess it would be nice to get a test showing it working before merging.  Also, maybe a squashed commit with all the "making it compile" changes?  Although that's not super necessary.

AlexIljin:
Yess!!

Using VS2015 (which I previously installed only to make changes to Factor VM) I have created a small project pulling data from ODBC. That gave me the subtle knowledge of the DSN strings, the expected txt-file format and the SQL query details.

Having that under my belt I was able to fix the few remaining bugs in the `odbc` vocabulary, and have it pull the same data from the same file as the VS2015 project did.

Unfortunately, my test case involved some local configuration made in the ODBC Data Sources (in Control Panel\Administrative Tools), so I can't (yet?) make it into a unit-test. (I've created a local DSN string using [this article](http://www-01.ibm.com/support/docview.wss?uid=swg21419207) as a hint.)

As to the squashing, I'm a bit reluctant to do it, because it basically shows some of the changes that happened to Factor between 2007 and today. I see it as helpful info for when someone else tries to pull stuff from unmaintained. I certainly did a lot of git-log digging, because I'm fairly new to the scene. I had no idea we had direct access to the retain stack back in the day (>r/r>) - today I couldn't even find it explained in the docs what the retain stack is and what it's for.

Also, did `10 [ . ] each` use to output numbers 0 through 9?

I'd appreciate it if you accepted my branch without squashing.

mrjbq7:
Integers used to be iterable, so ``10 [ . ] each`` is the same as ``10 iota [ . ] each``.

mrjbq7:
Is this branch ready to merge?

AlexIljin:
Yes, it is.

mrjbq7:
Awesome, merged. Thanks!!

catb0t:
@mrjbq7 Out of curiosity, why were integers ever iterable, or if the answer is "because it was cool", was the feature removed because it led to obscure bugs?

mrjbq7:
I think mostly because it was cool.  I think a better form in the future would support syntax like Swift's number ranges (e.g., ``0..10``, ``0..<10``).

PR #1739, opened 2016-11-01
Doc fixes

AlexIljin:
Some more documentation updates and fixes.


basis/threads/threads-docs.factor
```diff
@@ -181,9 +181,10 @@ $nl
     "A simple thread that adds two numbers:"
     { $code "1 2 [ + . ] 2curry \"Addition thread\" spawn" }
     "A thread that counts to 10:"
+    ! Don't use $example below: it won't pass help-lint.
     { $code
       "USING: math.parser threads ;"
-      "[ 10 iota [ number>string write nl yield ] each ] \"test\" spawn"
+      "[ 10 iota [ number>string write nl yield ] each ] \"test\" spawn drop"
```
mrjbq7: Why are you dropping the spawned thread?

basis/threads/threads-docs.factor
```diff
@@ -181,9 +181,10 @@ $nl
     "A simple thread that adds two numbers:"
     { $code "1 2 [ + . ] 2curry \"Addition thread\" spawn" }
```
mrjbq7: We don't drop it here...

basis/threads/threads-docs.factor
AlexIljin: Well, it's irrelevant to the example, and is not mentioned in the code output. I thought the missing `drop` was an oversight. Either that, or the stated output of the code.

mrjbq7:
Merged...

PR #1744, opened 2016-11-10
Distributed messaging Fix

doublec:
This fixes distributed messaging so that it works again. Changes in commit f20ee7a53b264fda515d2a55c8fbea26261de016 broke some use cases - in particular the remote channels functionality.

The test uses a hard coded port. I couldn't work out how to do a randomly generated port that worked with threaded-server.

### DIFF
--- a/basis/concurrency/distributed/distributed-docs.factor
+++ b/basis/concurrency/distributed/distributed-docs.factor
@@ -1,5 +1,12 @@
-USING: help.markup help.syntax concurrency.messaging threads ;
+USING: help.markup help.syntax concurrency.messaging io.servers threads ;
 IN: concurrency.distributed
+
+HELP: local-node
+{ $var-description "A variable containing the " { $link threaded-server } " the current node is running on." } ;
+
+HELP: start-node
+{ $values { "addrspec" "an addrspec to listen on" } }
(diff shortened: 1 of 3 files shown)


doublec:
I've updated [factor-articles.pdf](https://bluishcoder.co.nz/factor-articles.pdf) so that the remote channels example works with these changes if you're looking for usage.

mrjbq7:
I don't understand why this requires a global variable?  Why would that not be up to the user to decided how/where to create and start one and when/how to stop it?

Is the issue with randomized port, trying to know ahead of time by a client what the port is?

doublec:
It is up to the user to decide how/when to create a node. Once it's created distributed messages can be sent between nodes and they access it via the variable.

Currently (pre-this commit) this is done via the threaded-server variable which requires wrapping usage in a with-threaded-server combinator. That combinators starts and stops the server. What is needed though is a persistent server that stays around so other nodes can connect to it. Using a different variable also prevents clashes when using distributed messages from within a threaded server.

The port issue is for the test only. The distributed messaging code lets the user decide the port when calling start-node by having them pass an addrspec.

mrjbq7:
I see, thanks!  Merged!!!

PR #1752, opened 2016-11-27
Avl

jonenst:
Hi,
```factor
{
    { 0 0 }    { 1 1 }    { 3 3 }    { 4 4 }
    { 6 6 }    { 2 2 }    { 7 7 }    { 5 5 }
} >avl
```
throws retain-stack overflow. This is because ```rotate``` creates cycles in the tree and you get infinite recursion.

The error is actually visible with ```AVL{ { 0 0 } { 1 1 } { 2 2 } { 3 3 } { 4 4 } { 5 5 } } ...``` . This tree contains the node '0' twice and lost the node '2'.

This PR fixes it and improve the test coverage of rotate to avoid future regressions.

Note: for all the permutations of 0-8, the probability to throw is 0.7% ... Found the testcase by bruteforcing them. But as the tree gets bigger, the probability to throw increases exponentially. A random tree with 50 nodes is basically impossible to build..

### DIFF
--- a/extra/trees/avl/avl-tests.factor
+++ b/extra/trees/avl/avl-tests.factor
@@ -3,7 +3,7 @@ assocs accessors trees.avl.private trees.private ;
 IN: trees.avl.tests
 
 { "key1" 0 "key2" 0 } [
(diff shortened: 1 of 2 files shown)


mrjbq7:
Merged, thanks!!

PR #1755, opened 2016-11-27
Fix trails

AlexIljin:
I found that the `trails` vocab leaves its thread behind, and it eats 100% CPU in the background.
This fixes it.

### DIFF
--- a/extra/trails/trails.factor
+++ b/extra/trails/trails.factor
@@ -54,7 +54,7 @@ TUPLE: trails-gadget < gadget paused points ;
     ]
     loop
   ]
-  in-thread ;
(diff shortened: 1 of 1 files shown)


mrjbq7:
Merged, thanks!!

bjourne:
Perhaps you also need to delete the thread? Or is stopping it enough?

mrjbq7:
The thread quits it's loop if "paused" is true.  But if the trails gadget is ever re-grafted to the UI, it doesn't restart the thread.  So it's a little fragile/broken.  But at least it doesn't use 100% cpu.  I would welcome a cleanup PR if anyone wants to make it a little more better.

PR #1761, opened 2016-12-13
Fix set-query-params usage

AlexIljin:
The first commit here fixes an actual issue with `google.charts`, the second one sort of proactively tries to do the right thing elsewhere.


mrjbq7:
Merged, thanks!!

PR #1762, opened 2016-12-13
Misc fixes

AlexIljin:
These are mostly documentation fixes and updates, and one source formatting fix.

### DIFF
--- a/basis/command-line/command-line-docs.factor
+++ b/basis/command-line/command-line-docs.factor
@@ -115,7 +115,7 @@ ARTICLE: ".factor-roots" "Additional vocabulary roots file"
 $nl
 "A word to create this file:"
 { $subsections scaffold-factor-roots }
-"A word to run this file from an existing Factor session:"
+"A word to load this file from an existing Factor session:"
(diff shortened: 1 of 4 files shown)


basis/command-line/command-line-docs.factor
```diff
@@ -112,7 +113,9 @@ $nl
 ARTICLE: ".factor-roots" "Additional vocabulary roots file"
 "The vocabulary roots file is named " { $snippet ".factor-roots" } ". If it exists, it is loaded every time Factor starts. It contains a newline-separated list of " { $link "vocabs.roots" } "."
 $nl
-"A word to run this file from an existing Factor session:"
+"A word to create this file:"
+{ $subsections scaffold-factor-roots }
```
mrjbq7: I think this would cause ``tools.scaffold`` to be loaded (since this creates a new dependency), not sure we want that...

mrjbq7:
Ok, merged everything except the command-line doc change because I didn't want ``tools.scaffold`` to load in the default bootstrap.

PR #1765, opened 2016-12-24
google.charts: fix tests, add comments

AlexIljin:
In commit 294f75f0c the tests got broken - in the sense that they no longer detected what they were meant to detect (which was the missing `clone` call in `chart>url`), even though they all passed. The problem was with the reordering of constructors that eliminated the side-effects that were tested.

This restores the functionality.


mrjbq7:
Sounds good, thanks!

PR #1769, opened 2017-01-02
grouping-docs: add wrap-around note

shanepelletier:
Added "wrapping around the end of the sequence" to the description of <circular-clumps>, for consistency with the description of circular-clump.


mrjbq7:
Merged, thanks!!

PR #1770, opened 2017-01-03
core layouts, remove hardcoded tag width

jonenst:
109aa88b060 showed the places were it was hardcoded.

This changes core/layouts/layouts.factor but the values stay the same (just taking the folded symbol value into account). The max-array-capacity is still optimied to a constant.


mrjbq7:
Merged, thanks!

PR #1771, opened 2017-01-03
compiler.cfg: add peephole opt to never untag the input of ##bit-coun…

### LINKED ISSUE #1764: unnecessary untagging of fixnums for bit-count
I noticed that bit-count untags the input (sar rax 0x4) but it doesn't need to:
```factor
[ { array-capacity } declare bit-count ] disassemble
```

```
00007f4f7acd06d6: 498b06        mov rax, [r14]
00007f4f7acd06d9: 48c1f804      sar rax, 0x4
00007f4f7acd06dd: f3480fb8c0    popcnt rax, rax
00007f4f7acd06e2: 48c1e004      shl rax, 0x4
00007f4f7acd06e6: 498906        mov [r14], rax
00007f4f7acd06ef: c3            ret
```

I tried to fixed it with the following patch but I don't know the compiler enough to know if that's the correct way to do it. What do you think @bjourne, I think you are the one who spent the most time working on the compiler recently?
```diff
diff --git a/basis/compiler/cfg/representations/peephole/peephole.factor b/basis/compiler/cfg/representations/peephole/peephole.factor
index f505cb7..a165a37 100644
--- a/basis/compiler/cfg/representations/peephole/peephole.factor
+++ b/basis/compiler/cfg/representations/peephole/peephole.factor
@@ -281,3 +281,12 @@ M: ##not optimize-insn
         }
         [ call-next-method ]
     } cond ;
+
+M: ##bit-count optimize-insn
+    {
+        { [ dup src>> rep-of tagged-rep? ] [
+            [ no-use-conversion ]
+            [ finish ]
+            [ emit-def-conversion ] tri ] }
+        [ call-next-method ]
+    } cond ;
```

After this patch, I get the correct assembly:
```
00007f4f7ad1a6a6: 498b06        mov rax, [r14]
00007f4f7ad1a6a9: f3480fb8c0    popcnt rax, rax
00007f4f7ad1a6ae: 48c1e004      shl rax, 0x4
00007f4f7ad1a6b2: 498906        mov [r14], rax
```

Jon

jonenst:
…t. fixes #1764

I'm not sure how the value of vreg-counter is chosen in the test.. I know it will be used to generate a unique vreg for the newly created untag instruction, but the file has values like 5, 3, 6 etc. What's going on ??

### DIFF
--- a/basis/compiler/cfg/representations/peephole/peephole.factor
+++ b/basis/compiler/cfg/representations/peephole/peephole.factor
@@ -281,3 +281,6 @@ M: ##not optimize-insn
         }
         [ call-next-method ]
     } cond ;
+
+M: ##bit-count optimize-insn
+    [ no-use-conversion ] [ finish ] [ emit-def-conversion ] tri ;
--- a/basis/compiler/cfg/representations/representations-tests.factor
+++ b/basis/compiler/cfg/representations/representations-tests.factor
@@ -897,3 +897,21 @@ cpu x86.64? [
         T{ ##replace f 1 D: 0 }
     } test-peephole
 ] unit-test
+
+! untag elimination for ##bit-count
+2 vreg-counter set-global
+
+{
+    V{
+        T{ ##peek f 0 D: 0 }
+        T{ ##bit-count f 3 0 }
+        T{ ##shl-imm f 1 3 $[ tag-bits get ] }
+        T{ ##replace f 1 D: 0 }
+    }
+} [
+    V{
+        T{ ##peek f 0 D: 0 }
+        T{ ##bit-count f 1 0 }
+        T{ ##replace f 1 D: 0 }
+    } test-peephole
+] unit-test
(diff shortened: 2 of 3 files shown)


bjourne:
It's just a monotonically increasing number. You set it to 2 in your test so the next value is 3. Pr looks good, but in the future it is better to have code like `2 vreg-counter set-global` inside the unit tests because it makes them easier to move around or run in isolation.

mrjbq7:
Merged, thanks!

jonenst:
> It's just a monotonically increasing number. You set it to 2 in your test so the next value is 3.

So the rest of the file just randomly sets it to different values ...?

> in the future it is better to have code like 2 vreg-counter set-global inside the unit tests because that makes them easier to move around or run in isolation.

Yes, I used this structure only because that is what was used in the rest of the file. It's simpler to refactor and improve the whole thing in a separate commit.

PR #1773, opened 2017-01-05
Doctype

jonenst:
The first commit adds DOCTYPE html everywhere where it made sense, continuing the work of 4ef4235

The second commit fixes the mason report page (eg http://builds.factorcode.org/report?os=linux&cpu=x86.64 ) , it has a spurious <html><body>  next to the report table.

The third commit fixes mason report unit tests. Does someone know if these tests are running on the build farm ? I'm not seeing the failures on the report. Or I misunderstood something and they somehow pass on the build-farm ??

### DIFF
--- a/extra/mason/report/report-tests.factor
+++ b/extra/mason/report/report-tests.factor
@@ -10,6 +10,7 @@ mason.report mason.common mason.config namespaces tools.test xml xml.writer ;
      [ ] [ "report" delete-file ] unit-test ;
 
 "builds" temp-file builds-dir [
+    builds-dir get
     [
         "resource:extra/mason/report/fake-data/" "." copy-tree
 
@@ -29,5 +30,5 @@ mason.report mason.common mason.config namespaces tools.test xml xml.writer ;
         [ status-error ] [ 1236 test-failed ] unit-test
         verify-report
 
-    ] with-temp-directory
+    ] with-directory
 ] with-variable
--- a/extra/webapps/imagebin/upload-image.xml
+++ b/extra/webapps/imagebin/upload-image.xml
@@ -1,4 +1,4 @@
-<?xml version='1.0' ?>
+<!DOCTYPE html>
 <html>
 <head><title>Upload</title></head>
 <body>
--- a/extra/webapps/imagebin/upload-image.xml
+++ b/extra/webapps/imagebin/upload-image.xml
@@ -1,3 +1,4 @@
+<?xml version='1.0' ?>
 <!DOCTYPE html>
 <html>
 <head><title>Upload</title></head>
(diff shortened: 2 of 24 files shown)


mrjbq7:
``mason.report`` tests run for me locally, whether they test what they should, I'm not sure.  Was the test failing for you?

extra/mason/report/report-tests.factor
```diff
@@ -29,5 +30,5 @@ mason.report mason.common mason.config namespaces tools.test xml xml.writer ;
         [ status-error ] [ 1236 test-failed ] unit-test
         verify-report
 
-    ] with-temp-directory
```
mrjbq7: I think we want to use a ``temp-directory`` to copy the fake data into, not the ``builds-dir``.

mrjbq7:
I pulled the first two commits, but not the ``mason.report`` test change because I'm not sure that one is right.

extra/webapps/imagebin/upload-image.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version='1.0' ?>
+<!DOCTYPE html>
 <html>
```
jonenst: Now I'm not sure about removing this line (same for uploded-image.xml) Maybe it should be added back ? Sorry for the noise

jonenst:
> mason.report tests run for me locally, whether they test what they should, I'm not sure. Was the test failing for you?

Yes it was failing.

It's because the tests fail if the /tmp/factor-temp/builds file/directory doesn't exist, and it is created by mason/disk/disk-tests.factor

So running "mason" test works (by chance, because it does disk-tests first), but running "mason.report" test doesn't (if the factor-temp is empty before running it).

So after the third commit, the directory builds is created by mason.report-test
Also all the fake data files are put in there, so it's a little cleaner.

I didn't understand this when I made the commit, but now I think it's a good commit.

mrjbq7:
That's weird because I just did this on macOS and it worked fine:

```
USE: mason.report
"mason.report" test
```

But it seems to fail on linux.  I still think we shouldn't test into the ``builds-dir``, that seems wrong -- better to have a temporary directory?

jonenst:
I did break the tests ! I tried runng them but must have missed this one. Found it with https://travis-ci.org/jonenst/factor/jobs/189610859 (look for "Build report" :
                         Build report for linux/x86.64
                               Unit test failures
     * resource:basis/html/templates/fhtml/fhtml-tests.factor
)

The following patch fixes it:

```diff
commit 5518ed7b0e4c18e699062b5c6fe67cb9e2d80da0
Author: Jon Harper <jon.harper87@gmail.com>
Date:   Fri Jan 6 22:13:12 2017 +0100

    Ooops broke the tests with doctype

diff --git a/basis/html/templates/fhtml/test/example.html b/basis/html/templates/fhtml/test/example.html
index a65c78f..409c9d8 100644
--- a/basis/html/templates/fhtml/test/example.html
+++ b/basis/html/templates/fhtml/test/example.html
@@ -1,3 +1,4 @@
+
 <!DOCTYPE html>
 <html>
     <head><title>Simple Embedded Factor Example</title></head>


```

mrjbq7:
No worries, already fixed it in 9836dd77344cdc548dc5b969533ebc89f2757fb4.

PR #1774, opened 2017-01-05
Next batch of fixes

AlexIljin:
Hi!

  I have a couple of documentation updates here, and a couple of changes to the PDF vocab.

### DIFF
--- a/basis/concurrency/distributed/distributed-docs.factor
+++ b/basis/concurrency/distributed/distributed-docs.factor
@@ -10,27 +10,29 @@ HELP: start-node
 
 ARTICLE: "concurrency.distributed.example" "Distributed Concurrency Example"
 "In this example the Factor instance associated with port 9000 will run "
-"a thread that receives and prints messages "
-"in the listener. The code to start the thread is:"
-{ $examples
-    { $unchecked-example
-        ": log-message ( -- ) receive . flush log-message ;"
-        "[ log-message ] \"logger\" spawn dup name>> register-remote-thread"
-    }
+"a thread that receives and prints messages in the listener. "
+"The code to run the server is:"
+{ $code
+  "USING: io.servers ;"
+  "9000 local-server <node-server> start-server drop"
 }
-"This spawns a thread waits for the messages. It registers that thread as "
-"able to be accessed remotely using " { $link register-remote-thread } "."
+"The code to start the thread is:"
+{ $code
+    "USING: concurrency.messaging threads ;"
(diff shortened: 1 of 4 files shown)


basis/concurrency/distributed/distributed-docs.factor
```diff
@@ -39,9 +41,9 @@ $nl
 $nl
 "It is also possible to use " { $link send-synchronous } " to receive a "
 "response to a distributed message. When an instance of " { $link thread } " "
-"is marshalled it is converted into an instance of " { $link remote-thread }
+"is marshalled, it is converted into an instance of " { $link remote-thread }
 ". The receiver of this can use it as the target of a " { $link send }
-" or " { $link reply } " call." ;
+", " { $link send-synchronous } " or " { $link reply-synchronous } " call." ;
```
mrjbq7: Why did you remove ``send`` and ``reply`` in favor of the synchronous versions?  Maybe we should mention both versions.

AlexIljin: Ain't no such thing as `reply`, my friend.
Also, I didn't remove `send`.

extra/pdf/layout/layout.factor
```diff
@@ -359,7 +359,7 @@ TUPLE: pdf-info title timestamp producer author creator ;
     pdf-info new
         now >>timestamp
         "Factor" >>producer
-        "USER" os-env "unknown" or >>author
+        "USER" os-env \ developer-name get-global "unknown" or or >>author
```
mrjbq7: I wonder if we should instead have symbols that are for the pdf-author, instead of trying to guess this way.

```factor
SYMBOLS: pdf-producer pdf-author pdf-creator ;
```

AlexIljin: Good idea.

basis/concurrency/distributed/distributed-docs.factor
mrjbq7: Ahh, I missed that.

extra/pdf/layout/layout.factor
mrjbq7: Want to make that change? Or should I pull these commits and then make it myself?

AlexIljin: OK, I'll do it.

AlexIljin: Done!

mrjbq7:
Merged!

PR #1775, opened 2017-01-06
io.directories: improve file moving words

### LINKED ISSUE #1772: error when unix move-file accross mount points
I have separate partitions for / and /home
```
$ mount
/dev/sda1 on / 
/dev/sda5 on /home
```
When running the following
```factor
"/tmp/bar" "/home/jon/bar" move-file
```
I get this exception
```
Unix system call “rename” failed:

Invalid cross-device link (18)

It was called with the following arguments:

"/tmp/bar2"
"/home/jon/bar2"
```

```M\ unix move-file``` should check for EXDEV and default to copy then delete

bpollack:
`move-file` previously had inconsistent behavior on Windows and unixes. This unifies the behavior to the common case (just get 'er done), while also introducing an obviously named word, `move-file-atomically`, to handle the case where you need an atomic file primitive.

Fixes #1772

### DIFF
--- a/basis/io/directories/directories-docs.factor
+++ b/basis/io/directories/directories-docs.factor
@@ -104,7 +104,13 @@ HELP: touch-file
 
 HELP: move-file
 { $values { "from" "a pathname string" } { "to" "a pathname string" } }
-{ $description "Moves or renames a file." }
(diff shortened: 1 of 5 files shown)


mrjbq7:
Merged, thanks!!!

PR #1780, opened 2017-01-13
ui.backend.gtk, use a separate gtk_drawing_area for drawing opengl. Fixes #1487

### LINKED ISSUE #1487: Resizing ui in gtk takes a very long time
When resized, the original window size is still drawn and Factor seems to redraw the window for each tick of the resize, taking several seconds. In the space that the resize is attempting to grow to is just black pixels.

jonenst:
Wow this bug was hard to debug ! After poking around changing random settings in the factor initialization, I started by trying to reproduce it in plain C. It took me a while to reproduce but I finally found that doing the opengl draws on a drawing_area instead of the window during the configure callback was triggering the bug (the official examples at https://github.com/tdz/gtkglext/blob/master/examples/simple.c also use the drawing_area, so I guess it's the best way to do it)

In factor, I did the draws on a drawing area, but it didn't work. It took me again a lot of time to figure out that you have to setup 2 configure callbacks, one on the window and one on the drawing area. The configure callback on the window shouldn't draw (even call select-gl-context!), but the configure callback on the drawing_area can.

So we should test this patch on other platforms to see if it causes regression, because nothing comes from official documentations, it's all trial and error.. At least it's fixed on my computer with Ubuntu 14.04 unity

### DIFF
--- a/basis/ui/backend/gtk/gtk.factor
+++ b/basis/ui/backend/gtk/gtk.factor
@@ -15,9 +15,9 @@ IN: ui.backend.gtk
 
 SINGLETON: gtk-ui-backend
 
-TUPLE: window-handle window im-context fullscreen? ;
+TUPLE: window-handle window drawable im-context fullscreen? ;
 
-: <window-handle> ( window im-context -- window-handle )
+: <window-handle> ( window drawable im-context -- window-handle )
     f window-handle boa ;
 
 : connect-signal-with-data ( object signal-name callback data -- )
@@ -219,14 +219,14 @@ icon-data [ default-icon-data ] initialize
 ! Window state events
 
 : on-expose ( win event user-data -- ? )
-    2drop window relayout t ;
+    2drop gtk_widget_get_toplevel window relayout t ;
 
-: on-configure ( win event user-data -- ? )
-    drop swap window dup active?>> [
-        swap GdkEventConfigure memory>struct
-        [ event-loc >>window-loc ] [ event-dim >>dim ] bi
-        relayout-1
-    ] [ 2drop ] if f ;
+: on-configure ( window event user-data -- ? )
+    drop swap dup gtk_widget_get_toplevel [ = ] keep window dup active?>> [
+        swap [ swap GdkEventConfigure memory>struct ] dip
+        [ event-loc >>window-loc drop ]
+        [ event-dim >>dim relayout-1 ] if
+    ] [ 3drop ] if f ;
 
 : on-map ( win event user-data -- ? )
     2drop window t >>active? drop t ;
@@ -234,11 +234,16 @@ icon-data [ default-icon-data ] initialize
 : on-delete ( win event user-data -- ? )
     2drop window ungraft t ;
 
+: connect-configure-signal ( winhandle -- )
+    [ window>> ] [ drawable>> ] bi "configure-event"
+    [ on-configure yield ] GtkWidget:configure-event
+    [ connect-signal ] 2curry bi@ ;
+
+: connect-expose-sigal ( drawable -- )
+    "expose-event" [ on-expose yield ]
+    GtkWidget:expose-event connect-signal ;
+
 :: connect-win-state-signals ( win -- )
-    win "expose-event" [ on-expose yield ]
-    GtkWidget:expose-event connect-signal
-    win "configure-event" [ on-configure yield ]
-    GtkWidget:configure-event connect-signal
     win "delete-event" [ on-delete yield ]
     GtkWidget:delete-event connect-signal
     win "map-event" [ on-map yield ]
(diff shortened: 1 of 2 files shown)


jonenst:
On xfce, the resizing is a lot smoother.

jonenst:
Don't pull this in yet, there's a regression on xfce when putting the factor window on another desktop workspace, then changing workspace by dragging on another window(any other window) to the workspace with the factor window. The factor window will then be black (or other glitchy pixels)

jonenst:
Ok, the expose event wasn't fired anymore from the window (makes sense, it was not associate with the opengl context anymore), but from the drawable child, I had to change the connect.

I fixed it and pushed -f, should be good for everyone to test, and then merge.

jonenst:
By the way, to update, you can launch factor -run=listener, then "ui.backend.gtk" reload, then [ "ui.tools" run ] with-ui, or save directly and relaunch factor normal. Or you can mboostrap.

mrjbq7:
Awesome!! I'll try and take a look in a couple days. Traveling this weekend.

basis/ui/backend/gtk/gtk.factor
```diff
@@ -219,26 +219,31 @@ icon-data [ default-icon-data ] initialize
...
+        [ event-dim >>dim relayout-1 ] if
+    ] [ 3drop ] if f ;
 
 : on-map ( win event user-data -- ? )
     2drop window t >>active? drop t ;
 
 : on-delete ( win event user-data -- ? )
     2drop window ungraft t ;
 
+: connect-configure-signal ( winhandle -- )
```
mrjbq7: Maybe we should hook up two different words to the two different elements.

``
: on-configure-window ( window -- )

: on-configure-drawable ( drawable -- )
```

so you don't have to do the ``gtk_widget_get_toplevel [ = ] keep`` ?

File: basis/ui/backend/gtk/gtk.factor [hunk 2]
```diff
@@ -420,9 +425,11 @@ M: window-handle flush-gl-context ( handle -- )
 
 M:: gtk-ui-backend (open-window) ( world -- )
     GTK_WINDOW_TOPLEVEL gtk_window_new :> win
+    gtk_drawing_area_new :> drawable
+    win drawable gtk_container_add
```
mrjbq7: So the main idea is we have a top level container that we add everything into?  Whereas before we used the window as the top level?

basis/ui/backend/gtk/gtk.factor
jonenst: That's what I did originally but then changed it so that it's clearer that we do either the position update or the relayout. Both ways are fine.

basis/ui/backend/gtk/gtk.factor
jonenst: Yes kind of. We only need gtk to give us an opengl context. So before we were calling gtk_widget_set_gl_capability and gtk_widget_get_gl_context on the window. The window had no children.

Now the window only has 1 child (the child has no children). We call the 2 gtk functions on the child.

So the only difference is that those 2 contexts are treated differently inside the gtk library (see https://git.gnome.org/browse/gtkglext/tree/gdk/gdkglwindow.c vs https://git.gnome.org/browse/gtkglext/tree/gdk/gdkglpixmap.c I think)

jonenst: In particular, if any opengl calls happen during the configure signal on the window, gtk (on unity) will send the events for all the frames. But if you do your opengl calls during the configure on the drawable, gtk will coalesce the events. At least that's what I saw empirically, didn't find docs for that.

mrjbq7:
In the example program you link to they only linked the configure event to the drawable.  I wonder if that would be sufficient for us.

jonenst:
If you link it only to the drawable, there is no event when the window is
moved and the factor variable for the window position will not be updated.

Le 21 janv. 2017 7:21 PM, "John Benediktsson" <notifications@github.com> a
écrit :

In the example program you link to they only linked the configure event to
the drawable. I wonder if that would be sufficient for us.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/factor/factor/pull/1780#issuecomment-274278635>, or mute
the thread

mrjbq7:
Merged, thanks!!

bjourne:
Cool it  works!! I really, really don't understand how. Were we previously responding to configure events during  resize we aren't anymore?

jonenst:
It's mostly internal to the gtk/unity/x11 librairies and their
interactions. They all behave differently when you do opengl calls during
the configure signal connected to a window with opengl capabilities vs
connected to a drawable_area with opengl capabilities..

Also since the window didn't have opengl capabilities anymore, it didn't
fire the expose event. I had to connect the drawable expose event instead.

Also, since I had moved the configure to the drawable, there was no event
fired when the window was moved. So I reconnected the configure event of
the window and now moving is smoother because we don't draw on move, we
only update the factor value of the window position.


Le 30 janv. 2017 12:37 PM, "Björn Lindqvist" <notifications@github.com> a
écrit :

PR #1784, opened 2017-01-23
core.math, bignum/f, shift subnormals before rounding. Fixes #1782

### LINKED ISSUE #1782: float subnormals are truncated instead of rounded
A strinking example is that the smallest non zero float is wrongly parsed in decimal form.

1 bits>double number>string string>number . ! 0.0, should be 4.940656458412465e-324

This is because the smallest float actually has the following digit after what we represent:
4.940656458412465e-324    ! what we represent
4.94065645841246544...2e-324    ! actual value

So when you enter it in decimal form, it's slightly lower and truncated to 0. But this is a problem with all subnormals.

The following patch fixes most of the problems, but is wrong when rounding happens twice in opposite directions. I'll leave it here as a proof of concept but don't apply it. Instead the rounding should be correct the first time we do it because we lose the remainder information after that.
```diff
diff --git a/core/math/integers/integers.factor b/core/math/integers/integ
index bf36076..5e6f0b6 100644
--- a/core/math/integers/integers.factor
+++ b/core/math/integers/integers.factor
@@ -180,11 +180,14 @@ M: bignum (log2) bignum-log2 ; inline
 ! Fourth step: post-scaling
 ! Because of rounding, our mantissa with guard bit is now in the
 ! range [2^54;2^53], so we have to handle 2^54 specially.
+: round-shift-subnormal ( mantissa scale -- mantissa' )
+    1 + shift f swap 0 round-to-nearest 2/ ;
+
 : scale-float ( mantissa scale -- float' )
     ! At this point, the scale value is the exponent minus 1.
     {
         { [ dup 1024 > ] [ 2drop 1/0. ] }
-        { [ dup -1021 < ] [ 1021 + shift bits>double ] } ! subnormals and
+        { [ dup -1021 < ] [ 1021 + round-shift-subnormal bits>double ] } 
         [ [ 52 2^ 1 - bitand ] dip 1022 + 52 shift bitor bits>double ]
     } cond ; inline
 

```

Note: python does this correctly:
```python
>>> 2.470656458412465e-324
5e-324
>>> 2.460656458412465e-324
0.0
>>> float.fromhex('0x0.8000000000000000000000000000000001p-1074')
5e-324
>>> float.fromhex('0x0.80p-1074')
0.0
```

jonenst:


mrjbq7:
Merged!

Issue #1754, opened 2016-11-27
Add range/closer/slurp operations to trees


jonenst:
Currently trees only support the assoc protocol. But compared to hashtables, they can do more.

Looking at java https://docs.oracle.com/javase/8/docs/api/java/util/NavigableMap.html
  * range filter operations : headMap, subMap, tailMap
  * extreme keys: firstKey lastKey firstEntry lastEntry
  * closest keys: ceilingKey (>=) higherKey(>), floorkey lowerkey , same with Entries.
  * in order "slurping" : pollFirstEntry pollLastEntry
    * this is like a priority queue (slurp-heap). It doesn't have O(1) peek though, so you can't entirely treat it as a priority queue..

This is just a reminder for someone who would want to contribute an implementation :)

PR #1785, opened 2017-01-24
Trees

jonenst:
#1754 
I fixed some bugs in the extra:trees and added the features I was talking about.
I didn't go full meta-programming so the code is verbose, please improve if you can !
Also, I wasn't totally happy with the naming of the words, if anyone has better suggestions, please tell.

### DIFF
--- a/extra/trees/avl/avl-docs.factor
+++ b/extra/trees/avl/avl-docs.factor
@@ -1,4 +1,4 @@
-USING: help.syntax help.markup assocs ;
+USING: assocs help.markup help.syntax trees ;
 IN: trees.avl
 
 HELP: AVL{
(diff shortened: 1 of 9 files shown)


jonenst:
Hmm wait I just realized that pop and slurp won't maintain the properties of subclasses of tree, that's not good. Let me rework that tomorow.

jonenst:
Okay, now it's correct for all trees. I also changed >tree >avl and >splay in the case the input is a tree to attempt to keep the shape by choosing the order of the alist derived from the input (bfs for regular tree and avl, reverse bfs for splay tree).

I was thinking of adding a tree.traversal vocabulary with visitors for inorder, preorder, postorder and bfs traversals, along with >alist equivalents (and move the code that I added for bfs in trees.factor), what do you think ?

mrjbq7:
Awesome! It sounds maybe useful to put in a traversal vocabulary. Maybe a stylistic thing, but if you are leaning toward it, sure.

jonenst:
> It sounds maybe useful to put in a traversal vocabulary. Maybe a stylistic thing, but if you are leaning toward it, sure.

I didn't understand... What's the stylistic alternative ?

Anyway, if you have time to review the code in this PR that'd be great. Especially come up with better names :)

mrjbq7:
> I was thinking of adding a tree.traversal vocabulary with visitors for inorder, preorder, postorder and bfs traversals, along with >alist equivalents (and move the code that I added for bfs in trees.factor), what do you think ?

I was just thinking that whether you have a massive ``trees`` vocab or split out to ``trees.traversal`` is a stylistic judgment.  I do like the idea of the ``trees.traversal`` (similar to ``xml.traversal``) because the API seems to be growing and it could be a little confusing looking at all the words in ``trees``?

mrjbq7:
By the way, I merged this.  Thanks!!

New features, bug fixes - love it!

PR #1791, opened 2017-02-21
fixed variable name mismatch

thron7:
There is a variable name mismatch in build.sh's `gcc_version_ok()`, which prevents proper version checking for GCC. My patch fixes that.

Besides that, there is another issue with GCC's `-dumpversion` option: In my version of GCC (4.8.5) it only returns "4.8", missing the patch version. I could fix this in my copy of build.sh with
~~~
   GCC_VERSION=`gcc --version | head -1 | cut -d" " -f"4"`
~~~
but as I don't know how other GCC versions might behave, I left it out of the PR.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -139,8 +139,8 @@ gcc_version_ok() {
 
 	if [[ $GCC_MAJOR -lt 4
 		|| ( $GCC_MAJOR -eq 4 && $GCC_MINOR -lt 7 )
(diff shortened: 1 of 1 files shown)


mrjbq7:
Merged, thanks!

(Does the test fail for your ``4.8`` version?)

thron7:
No, with `GCC_VERSION` properly initialized and the variable names fixed, the version check runs fine, i.e. my version 4.8.5 is accepted.

Issue #1158, opened 2014-10-10
Some floats could be prettier


jonenst:
{ 0.07 0.56 } .
{ 0.07000000000000001 0.5600000000000001 }

See the 2007-2008 interesting discussion about the same problem (solved) for python:
http://bugs.python.org/issue1580

mrjbq7:
Fixed by Dragonbox!

PR #1792, opened 2017-02-26
Formatting+integer logs+fixes

jonenst:
Hi,
Lot's of stuff in this PR, it started small but grew to this ! please review :)
Mainly,
it adds support to print all ratios and integers with %f and %e.

It fixes some bugs in rounding
```-4/5 round``` was 1 instead of -1
```-0.1 round``` was 0.0 instead of -0.0

in formatting, it changes the meaning of %d %b and %o to not coerce the input to an integer, so that they behave like %x and can print floats and ratios in their bases.

it also changes %f and %e when the precision is 0 to remove the added ".0". This makes it more apparent that the precision is 0, not 1.

Also, some small things like documentation for "% d" ( it's like "%+d" but with a space ), and a small bugfix for "%+e" when the number is positive but less than 1 (the plus sign was not added)

Please feel free to say that some commits should not go in !
Cheers,
Jon

### CODE REVIEW

File: basis/math/functions/integer-logs/integer-logs.factor [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+! this to the first 27 elements..
+CONSTANT: log10-guesses {
+    0 0 0 0 1 1 1 2 2 2 3 3 3 3
+    4 4 4 5 5 5 6 6 6 6 7 7 7 8
+    8 8 9 9 9 9 10 10 10 11 11 11
+    12 12 12 12 13 13 13 14 14 14
+    15 15 15 15 16 16 16 17 17
+}
+
+! This table will hold a few unusued bignums on 32 bits systems...
```
mrjbq7: s/unusued/unused/

jonenst:
Hi,
Yeah as I mentioned in the commit message, there's a choice to be made
here.. The choice of the libc is to display "22"...
But what format most naturally conveys the idea that you have only this
amount of precision (0 digits after the decimal point) ? 22 looks like an
integer which has no notion of precision.
With "22." at least it's obvious that some kind of precision is implied.

Also, it's the same for "%e", do we want "2.e1" or "2e1" ? Although here
2e1 is commonly understood as having a defined precision.

In the end, I don't have a strong attachment to either solutions
(formatting with %0f is a pretty rare case after all), so feel free to
change it ! (If you change it, also change 2e1, although the code is in
common so it should change automatically, you just have to update the unit
tests)

Le 28 févr. 2017 19:00, "Björn Lindqvist" <notifications@github.com> a
écrit :

bjourne:
My medium to strong opinion about signed zero is that it is not a great idea. Because I have never had any use for it and I have never seen any code use that feature of ieee 754. Plus, it makes an expression like `x * 0.0` harder to optimize. So I think it would be better to leave `truncate` as is, and in the future we can change `-0.3 round` and other functions to evaluate to `0.0` instead of `-0.0`. I'll defer to @mrjbq7's opinion though.

basis/formatting/formatting.factor
```diff
@@ -113,9 +113,9 @@ fmt-C     = "C"                  => [[ [ 1string >upper ] ]]
 fmt-s     = "s"                  => [[ [ present ] ]]
 fmt-S     = "S"                  => [[ [ present >upper ] ]]
 fmt-u     = "u"                  => [[ [ unparse ] ]]
-fmt-d     = "d"                  => [[ [ >integer number>string ] ]]
-fmt-o     = "o"                  => [[ [ >integer >oct ] ]]
-fmt-b     = "b"                  => [[ [ >integer >bin ] ]]
+fmt-d     = "d"                  => [[ [ number>string ] ]]
+fmt-o     = "o"                  => [[ [ >oct ] ]]
+fmt-b     = "b"                  => [[ [ >bin ] ]]
 fmt-e     = digits "e"           => [[ first '[ _ format-scientific ] ]]
```
bjourne: I like this. Now it shows if you make a mistake and outputs a float with the  `"%d"` specifier.

mrjbq7: But doesn't every other language convert the float argument to an integer before printing with ``%d``?

mrjbq7: I think it should mean "print this as an integer"... essentially the same as ``%.0f``.  Otherwise, it should throw an error for non-integer inputs (which would be weird, too).  What it shouldn't do is print ``1.23456789``...

jonenst:
Ha, I like the idea of signed zeroes ! Especially with rounding, where you don't expect this operations to change the sign.. But I've never used it in practice

jonenst:
Also, I like pure and theoretical systems... Please don't hesitate to reject the parts of my ideas that prevent factor from being a useful and practical system :)

basis/formatting/formatting.factor
jonenst: I did this because
```factor
1.5 "%x\n" printf
1.8p0
```

So I made "%d", "%o" and "%b" work like "%x".

Maybe we should do the other way around, add >integer to %x ?
In C, they use "%a" for hexadecimal floats. So if we want to look like C so that people are not surprised by the library, we can do this and drop binary and octal floats ?

jonenst:
Hi,
I changed "%.0f" and "%.0e" to the more practical version where there is no radix character.
So now "%.0f" should output "22" instead of "22." like bjourne suggested.

Also fixed a bug where %e didn't work:
    99/10 0 format-scientific-simple was outputting 1.0e0 because
    the mantissa was rounded up, so it had one extra character, and
    the exponent was wrong.

jonenst:
Is there anything I should change to get this merged ?

mrjbq7:
@bpollack want to take a look at this and see what you think?

bjourne:
@jonenst, as I wrote I really think the -0.0 should be dropped. Signed zero is (more or less) an implementation detail of the ieee 754 standard and serves no practical purpose. But it does cause some edge case issues which from a users perspective is unexpected. For example, in `x 0.0 *` you can't fold the expression to `0.0` because if `x = -0.0` it would evaluate to that. Eventually, I think we should remove signed zeroes from everywhere in the code base. It would make optimizing floating point arithmetic easier.

bpollack:
I agree at a low-level with @bjourne on this one, @mrjbq7. I also want to do a pass through on code formatting/etc., which I didn't quite have time to do yesterday, but should have tomorrow; if I don't see anything, then, with the removal of `-0.0`, I think I'm on board.

bpollack:
Okay, I should not respond to emails right after getting up.

IEEE-751 says `-0` is a thing, and while I kind of theoretically agree that it'd be neat if it weren't (since it's mostly just an implementation detail of how we happen to represent floating point in modern hardware), Factor will need to support it if we're going to work with existing IEEE-751 hardware and if we want to interface with existing libraries. We're thus going to need to support it in the language core, which in turn means I think we do really want to have full support for displaying and parsing in Factor core, too.

Beyond that: if we decide, later, that we do in fact want to eliminate -0 from Factor conceptually, I fail to see how @jonenst's patch series makes that _harder_ to do than it would be right now. And in terms of compiler optimizations, I think we'd get a ton more benefit out of endeavors like squeezing more high-level optimizations out of `TYPED:` call chains than we could on realizing we can fold `x 0.0 *` to `0.0`.

So I still want to review this patch series as I noted before merging, but I do not think we need to change how this series handles 0 v. -0. That seems fine.

bjourne:
Yes, but even if we would want to keep -0.0 in the language for compatibility with C libraries (which I'm not sure is needed), that does not mean Factor words needs to create -0.0 values themselves.

In C, the round() function returns -0.0 for the value -0.3, but there is no requirement that other language's rounding functions work similarly. For example in Java, Math.round(-0.3) == 0 and it is the same in the Python and Ruby where both round(-0.3) and truncate(-0.3) returns 0. In Python 2.7, round(-0.3) used to return -0.0 but it was changed in Python 3.

In Factor right now, `-0.3 truncate` and `-0.3 round` both evaluates to `0.0` which I think is right. Jon's pr changes `-0.3 truncate` to return `-0.0` which  I  think is wrong. 

I'm on board with all the other commits in the pr which I  think are good fixes to math and formatting.

jonenst:
Feel free to cherry pick only the good parts !

Le 30 avr. 2017 09:33, "Björn Lindqvist" <notifications@github.com> a
écrit :

Yes, but even if we would want to keep -0.0 in the language for
compatibility with C libraries (which I'm not sure is needed), that does
not mean Factor words needs to create -0.0 values themselves.

In C, the round() function returns -0.0 for the value -0.3, but there is no
requirement that other language's rounding functions work similarly. For
example in Java, Math.round(-0.3) == 0 and it is the same in the Python and
Ruby where both round(-0.3) and truncate(-0.3) returns 0. In Python 2.7,
round(-0.3) used to return -0.0 but it was changed in Python 3.

In Factor right now, -0.3 truncate and -0.3 round both evaluates to 0.0
which I think is right. Jon's pr changes -0.3 truncate to return -0.0 which
I think is wrong.

I'm on board with all the other commits in the pr which I think are good
fixes to math and formatting.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/factor/factor/pull/1792#issuecomment-298216852>, or mute
the thread

bjourne:
That's a good idea. :+1: I guess I should try and do the cherry picking since I had objections. I'll see if I can figure out how to do it, it was a long time since i committed anything to factor.

basis/formatting/formatting-docs.factor
```diff
@@ -78,8 +80,10 @@ HELP: printf
         "##12" }
     { $example
         "USING: formatting ;"
-        "1234 \"%+d\" printf"
-        "+1234" }
+        " 1234 \"%+d\\n\" printf"
```
mrjbq7: I wonder if it's not better to put a ``+`` instead of a space for positive integers?

Like Python:

```python
>>> '%+d' % 1234
'+1234'
>>> '%+d' % -1234
'-1234'
```

basis/formatting/formatting.factor
mrjbq7: In Python, ``%d`` and ``%x`` and ``%o`` format integers:

```
>>> '%d' % 123.456
'123'
>>> '%x' % 123.456
'7b'
>>> '%o' % 123.456
'173'
```

I realize, they don't have our ``%x`` for floats, so maybe ``%a`` is a better approach?

basis/formatting/formatting-docs.factor
mrjbq7: Oh, maybe this example is just wrong? It should be ``+1234``?

jonenst: The example looks correct. The whitespace before "1234" on line 83 is not in a string, it's in factor code. It's just so that examples align. The last line (line 86, "+1234\n-1234\n 1234" ) show the results with the correct '+' or ' ' in front of the number.

Compare what happens when you click on the example in the doc and it pastes it to the listener with and without the space.. It's just comestic..
formatting
formatting_unaligned

Anyway, I ran help-lint and it passed.

mrjbq7:
@bjourne agree with approach - let me know if you want help merging it. You could also merge it all and then revert that one patch.

mrjbq7:
So, I merged all of these patches and then applied a single patch (57de1f5ab003ad410032288616ec872627b11944) to change ``-0.3 truncate`` to return ``0.0``.  I'm sort of ambivalent about this patch.  I really like being compatible with other languages when it makes sense. But we also have current precedent for using ``-0.0`` in cases like ``-0.3 ceiling`` which returns ``-0.0.``.

mrjbq7:
@jonenst thanks for having the patience to work on this and make fixes and help us get it merged!!!

bjourne:
Thanks for doing this! I apologize for taking it upon myself and then not have any time to carry it out.

mrjbq7:
No worries!

jonenst:
Glad that we reached some kind of consensus!

Le 10 juin 2017 15:00, "John Benediktsson" <notifications@github.com> a
écrit :

erg:
It looks like ``%d`` is broken after this patch:

```
3.3 "%d" sprintf
"3.3"
```

From http://en.cppreference.com/w/cpp/io/c/fprintf

``%d`` ``converts a signed integer into decimal representation [-]dddd.``

I don't see any circumstances where it makes sense to break the C behavior of ``%d``. If we want to fix ``%d`` then it should do a ``integer?`` guard and throw an exception.

Is there any justification for breaking this?

mrjbq7:
I have a patch for it,wait for me to fix that.  I think I'm going to make x and o do integer formatting also (but we should add something that formats the way @jonenst was trying to do, it's kind of neat).

mrjbq7:
Fixed in 0fddbc2af16dec30d7031d146f2a5cabe5ce899e.

But why is this happening on Windows?

```
Unit Test: { { "1.2" } [ 1+1/4 "%.1f" sprintf ] }

=== Expected:
"1.2"
=== Got:
"1.3"
```

mrjbq7:
@jonenst ?

AlexIljin:
1+1/4 = 1.25 ~1.3
Maybe some rounding is in effect?

mrjbq7:
Odd it's only on Windows and maybe related to some of the rounding patches in this PR?

jonenst:
Will look into the rounding issue on monday when I get acces to a windows
computer. Maybe the windows printf doesn't do round to even? This is a new
test right? I don't think I changed the behavior but will double check.

For 1.3 %d", I changed it for symmetry with 1.3 "%x" which prints as a
float or an int.


Le 10 juin 2017 22:49, "John Benediktsson" <notifications@github.com> a
écrit :

Odd it's only on Windows and maybe related to some of the rounding patches
in this PR?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/factor/factor/pull/1792#issuecomment-307589419>, or mute
the thread

jonenst:
Yup I can confirm that on windows 10 using factor 0.97 stable, you get 
```factor
1+1/4 "%.1f" sprintf
"1.3"
```
So printf on windows doesn't do NearestTiesToEven rounding on base10 conversions. It's not surprising at all, @erg was talking about how microsoft pledged to improve their floating point handling ( also this lead me to read this hilarious bug report https://sourceware.org/bugzilla/show_bug.cgi?id=4943 ).

So this is has nothing to do with this patch, except for the added test ! Maybe add ```os window? [ ... ] unless``` ?

Regarding the behavior "%d" on floats and ratios, I don't think it "breaks" the C behavior because C just outputs garbage in this case, so both coercing to int and formatting as a float are differences to C. A good argument to keeping the coercion is backwards compatibility. But factor being a very dynamic language, I like having something in printf to print a "number" (float, int, ratio, complex).. It could be "%d" or something else. I think using "%d" has the advantage of simplifying the system (it fits neatly with printing in bases other than 10).
However I think discussion and consensus are the most important things and will gladly accept any solution :)
cheers,
Jon

jonenst:
Also I plan to integrate
https://github.com/google/double-conversion
after the coming release. This will fix the rounding issue from this thread, and #1158 and #1155

Or do we want it for this release ?

mrjbq7:
I don't understand what the rounding issue is, or why it's only on Windows?

mrjbq7:
@jonenst 

This adds "alternate" forms of formatting (similar to how Python works):

```python
Python 2.7.13 (default, Dec 18 2016, 07:03:39) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> '%#3x' % -12345
'-0x3039'
>>> '%#15x' % -12345
'        -0x3039'
```

Using this patch:

```diff
diff --git a/basis/formatting/formatting.factor b/basis/formatting/formatting.factor
index 54c131c191..b928ebdb7e 100644
--- a/basis/formatting/formatting.factor
+++ b/basis/formatting/formatting.factor
@@ -137,7 +137,15 @@ strings   = pad width strings_   => [[ <reversed> compose-all ]]
 numbers_  = fmt-d|fmt-o|fmt-b|fmt-e|fmt-E|fmt-f|fmt-x|fmt-X
 numbers   = sign pad numbers_    => [[ unclip-last prefix compose-all [ fix-sign ] append ]]
 
-types     = strings|numbers
+fmt-#b    = "b"                  => [[ [ >bin "0b" prepend ] ]]
+fmt-#o    = "o"                  => [[ [ >oct "0o" prepend ] ]]
+fmt-#x    = "x"                  => [[ [ >hex "0x" prepend ] ]]
+fmt-#X    = "X"                  => [[ [ >hex >upper "0x" prepend ] ]]
+
+alt_      = fmt-#b|fmt-#o|fmt-#x|fmt-#X
+alt       = "#" pad alt_         => [[ [ third [ fix-sign ] ] [ second ] bi 3append ]]
+
+types     = strings|numbers|alt
 
 lists     = "[%" types ", %]"    => [[ second '[ _ map ", " join "{ " prepend " }" append ] ]]
 
```

Works like this:

```factor
IN: scratchpad -122345 "%#x\n" printf
-0x1dde9
IN: scratchpad -122345.0 "%#x\n" printf
-0x1.dde9p16
IN: scratchpad -122345.0 "%#o\n" printf
-0o1.673644p16
IN: scratchpad -122345.0 "%#X\n" printf
-0x1.DDE9P16
IN: scratchpad -122345.0 "%#b\n" printf
-0b1.1101110111101001p16
```

Thoughts?

jonenst:
> I don't understand what the rounding issue is

When you want to format 0.25 to a decimal value with only 1 digit of precision, you have to choose between "0.2" and "0.3".. Different people made different choices. In math, we are used to rounding to the nearest, breaking ties away from zero (ie "0.3"). The IEEE754 standard says that we should round to the nearest, rounding ties to even (ie "0.2").  

> , or why it's only on Windows?

We use the system's functions (in 0.97, it was snprintf on unix and _snprintf on windows; currently we use std::ostringstream). The osttringstream (it was the same for snprintf) implementation is provided by the different implementations of libstdc++ (windows, gcc, clang etc). They don't have the same behavior apparently.

Note that I tested windows with 0.97 (using _snprintf), so it's not very relevant to seeing if there was a regression by this pull request. Can someone test ```1+1/4 "%.1f" sprintf``` on windows with a recent factor that doesn't have this pull request merged (2ffa446f5b5 or before) so that it uses windows's ostringstream?

jonenst:
> This adds "alternate" forms of formatting (similar to how Python works):
> Thoughts?

I think it's nice to add features using the same syntax as other languages because this way you reduce complexity a bit.

However, I still find it surprising that %x works for ints and floats, but %d doesn't. But yeah, if other systems use %a (C and python at least), maybe that's the way to go ? As you said, python coerces to int for %d, %x and %o.

Maybe to print a "number", "%u" is enough ? But it uses the pretty printer variables, which is suprising..
```
3.1 "%u\n" printf
16 number-base [ 3.1 "%u\n" printf ] with-variable
0x1.8cccccccccccdp1
```

mrjbq7:
If we had %x print as hex whatever you passed in like your PR wouldn't we want a "print as an integer hex" option anyway? So the question would be which should be easier / simpler to specify?

jonenst:
I was initially thinking that coercing a number to an integer and printing it in hex would be
```
>integer "%x" printf :)
```

But maybe it's a nice way to reach down the stack in case of complicated formatting, for example the following is verbose.
```
[ >integer ] 4 ndip "the %x is %s %s %s %s " printf
```

I agree with you that it comes down to a good choice of defaults !

mrjbq7:
Hmm, that's a fair point.

bjourne:
I tested with an old Factor and it also outputs "1.3" for 1+1/4 "%.1f" sprintf. So it is not this pr that is doing it.

PR #1796, opened 2017-03-16
Doc fixes

AlexIljin:
Some more documentation updates, including a brand new article on classes.maybe!

### DIFF
--- a/core/sequences/sequences-docs.factor
+++ b/core/sequences/sequences-docs.factor
@@ -678,7 +678,7 @@ HELP: replace-slice
 { $description "Replaces a range of elements beginning at index " { $snippet "from" } " and ending before index " { $snippet "to" } " with a new sequence." }
 { $errors "Throws an error if " { $snippet "new" } " contains elements whose types are not permissible in " { $snippet "seq" } "." } ;
 
-{ push prefix suffix } related-words
+{ push push-either push-if pop pop* prefix suffix suffix! } related-words
 
 HELP: suffix
 { $values { "seq" sequence } { "elt" object } { "newseq" sequence } }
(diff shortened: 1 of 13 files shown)


core/sequences/sequences-docs.factor
```diff
@@ -678,7 +678,7 @@ HELP: replace-slice
 { $description "Replaces a range of elements beginning at index " { $snippet "from" } " and ending before index " { $snippet "to" } " with a new sequence." }
 { $errors "Throws an error if " { $snippet "new" } " contains elements whose types are not permissible in " { $snippet "seq" } "." } ;
 
-{ push prefix suffix } related-words
+{ push push-all push-either push-if pop pop* prefix suffix suffix! } related-words
```
mrjbq7: I'm not sure how many related words we should have, for example ``push`` puts something at the end of a sequence, so I think ``suffix`` and ``suffix!`` make sense, but maybe ``prefix`` isn't close enough?  Not sure where the line is, but a list of all "modifies a sequence" is probably too wide.

Also ``push-all`` pushes several things, so is it more related to ``append`` and ``append!`` rather than ``push``?

AlexIljin: Personally, I like lots of cross-references, as they help discover related functionality when you know approximately what you need, but not sure of what exactly your options are. But it's up to you if you think I went too far in this case.

As to the specifics, the `prefix` and `suffix` were in one list before I chimed in, so it's not like I smashed together unrelated things that weren't there. I merely expanded the list of synonyms. That's how I see it, anyway.

Good point about the `push-all` and `append`! That was a little thoughtless of me.

AlexIljin: Yeah, it's a difficult line to draw. Maybe we say that the list should contain all words that are adding one item at either end of a sequence?

mrjbq7: That sounds good. I don't really have a strong opinion just think we should have a general guidance that's all.

AlexIljin:
I have moved the `push-all` to the `append`s, and added a typo fix in the `animations-docs`.

mrjbq7:
Merged, thanks!

I might want ``pop`` and ``pop*`` to be related but not to the push words.

PR #1799, opened 2017-03-23
French number>text modern usage + ratios

samueltardieu:
When I originally wrote the `number>text` French implementation, I used an older form for formatting numbers. The new one is taught in France, Belgium, Quebec, Switzerland, etc.

Also, this adds support for ratios.


mrjbq7:
Merged, thanks!!!

PR #1804, opened 2017-04-07
pos: delete vocab from unmaintained

AlexIljin:
The `pos` vocab was only useful with the `self` vocab, which was deleted a long time ago.


mrjbq7:
Merged, thanks!!

PR #1811, opened 2017-05-05
slides: map 'b' key to previous slide (like unix less)

viswans83:
'space' and 'b' are close together and make slide navigation similar to unix less.

### DIFF
--- a/extra/slides/slides.factor
+++ b/extra/slides/slides.factor
@@ -120,6 +120,7 @@ SYNTAX: STRIP-TEASE:
     { T{ button-down } [ request-focus ] }
     { T{ key-down f f " " } [ next-page ] }
     { T{ key-down f f "DOWN" } [ next-page ] }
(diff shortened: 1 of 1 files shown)


mrjbq7:
``less`` supports a couple more:

```
SPACE or ^V or f or ^F
              Scroll forward N  lines,

b or ^B or ESC-v
              Scroll backward N lines
```

Want to add support for those?

(We could change the ``f`` binding to not ``toggle-fullscreen``, maybe a different binding for that...)

viswans83:
Only the 'b' key binding made sense to me, mainly because 'space' and 'b' are in close proximity and that 'b' could mean 'back', along with 'space' being a large and prominent key on the keyboard. I imagine users would either use the arrow keys or use 'space' and 'b' keys to navigate slides. I doubt users would want to reach for the Ctrl key just to navigate slides.

mrjbq7:
Merged, thanks!

PR #1812, opened 2017-05-05
test runner: set exit code to number of failed test cases

robx:
This makes it easier to detect unit test failures from the command line.

Very new to factor, let me know if this should be done differently. Would it cause trouble on Windows?

### DIFF
--- a/basis/tools/test/test.factor
+++ b/basis/tools/test/test.factor
@@ -6,7 +6,7 @@ generalizations io io.files.temp io.files.unique kernel lexer
 locals macros namespaces parser prettyprint quotations sequences
 sequences.generalizations source-files source-files.errors
 source-files.errors.debugger splitting stack-checker summary
-tools.errors unicode vocabs vocabs.files vocabs.metadata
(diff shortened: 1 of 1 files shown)


mrjbq7:
That's a cool idea. Mostly we run unit tests inside a Listener but I added the command line test runner and this would make it useful!

I will merge when I get a moment this afternoon.

mrjbq7:
Thanks!

jonenst:
Exit codes are limited to 255. They are not meant to be used as a value. It would probably be more useful to return 0 for no errors and 1 for errors.

Also, I think most users rarely use the command line test runner. Instead, in the GUI listener, you can type "myvocab" test, and errors will be shown in a separate window (the same one with compile errors) that you can access by pressing F3.

Cheers,
Jon

jonenst:
You could have something like
0 : no errors
1 : some errors
2: no tests run
etc...

robx:
0/1 sounds good, will change

mrjbq7:
Merged, thanks!

(It's a shame you have to call ``exit`` manually, so perhaps some future improvement we can make there).

PR #1817, opened 2017-05-26
`Save As` Operation for the `byte-array`s

AlexIljin:
As I mentioned in the mailing list, I've added the possibility to right-click on a `byte-array` and - with the help of the standard Save As dialog - store its contents into a file. This PR also does a bit of metadata maintenance in the entire `file-picker` subtree by adding authors files and copyright headers.

To enable the new context menu operation, do `USE: file-picker.operations`.

### DIFF
--- a/extra/file-picker/operations/operations.factor
+++ b/extra/file-picker/operations/operations.factor
@@ -1,0 +1,13 @@
+! Copyright (C) 2017 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: byte-arrays file-picker io io.encodings.binary io.files
+kernel locals ui.commands ui.operations ;
+IN: file-picker.operations
+
+:: save-as ( data -- )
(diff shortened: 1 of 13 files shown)


extra/file-picker/operations/operations.factor
```diff
@@ -0,0 +1,13 @@
+! Copyright (C) 2017 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: byte-arrays file-picker io io.encodings.binary io.files
+kernel locals ui.commands ui.operations ;
+IN: file-picker.operations
+
+:: save-as ( data -- )
+    "" save-file-dialog [ binary [ data write ] with-file-writer ] when* ;
```
mrjbq7: Instead, this is probably better using ``set-file-contents``:

```factor
data binary set-file-contents
```

mrjbq7:
Merged, thanks!

(Also, I made the quick change to use ``set-file-contents``).

PR #1819, opened 2017-05-29
file-picker.operations: fix invalid parameter order in save-as

AlexIljin:
`set-file-contents ( seq path encoding -- )`


mrjbq7:
Haha, whoops.  Serves me right doing merges before coffee.  Thanks for fixing that.  Merged!

PR #1823, opened 2017-06-01
cocoa: Allow ?-> syntax for methods that might not exist.

erg:
If a method doesn't exist we need to provide a signature so the stack is balanced. This should also allow deploying from macOS versions that do not contain methods to ones that do. This is an alternative to asking a class if it provides a selector.

### DIFF
--- a/basis/cocoa/cocoa.factor
+++ b/basis/cocoa/cocoa.factor
@@ -14,6 +14,8 @@ SYMBOL: sent-messages
 
 SYNTAX: -> scan-token dup remember-send suffix! \ send suffix! ;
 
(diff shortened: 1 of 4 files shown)


erg:
Sure! I started trying to do it the retina way but it would involve looking up each selector separately and a lot of boilerplate.

erg:
Merge button totally doesn't work with our git setup. Lol.

mrjbq7:
Merge button is big no-no! Not only is this a copy haha but also it makes a merge commit bad git foo.

erg:
I actually did a rebase + merge (new github feature!) so there's no merge commit. But the two patches were still MIA.

Issue #1828, opened 2017-06-03
32-bit - Y U Crash?


AlexIljin:
Hello!

  I have created the following vocab:

```
USING: arrays checksums checksums.openssl kernel sequences ;
IN: crash-factor

: 1hash ( bytes -- hash ) openssl-sha1 checksum-bytes ;
: 5hash ( bytes -- h h h h h ) 1hash dup 1hash dup 1hash dup 1hash dup 1hash ;
: hashes ( seed -- seed' seq ) 300000 iota [ drop 5hash [ 4array ] dip swap ] map ;
: crash ( -- h h h h h h ) B{ } hashes swap hashes swap hashes swap hashes swap hashes ;
```

To run it, use: `USE: crash-factor crash`.

I'd like someone to tell me why does it crash the 32-bit build of Factor. Is it simply memory exhaustion, or is it something else?

The code was developed to reproduce a crash that was happening with much smaller memory load, but involving reading about 300'000 records from an SQLite DB twice (always good the first time, always crash the second). With this code I see that the Factor.exe process uses about 500Mb of memory at the time of the crash. It's about a quarter of the way to its theoretical limit for the 32-bit build. That's why I suspect GC issues or at least some fixable error of some kind.

I have tried to debug the app, but had little success. Despite having no VS2015 solution file in the Git repository, I managed to create a 32-bit debug build, and crash it with the vocab from under VS2015, but I see nothing useful I can do with it. First it throws a whole lot of `factor::must_start_gc_again` exceptions (which I think is a normal way it operates), then it says "factor.exe has triggered a breakpoint" while pointing at line 590 of the slot_visitor.hpp (again, many times - and I didn't set any breakpoints in VS), and then I see a lot of the following messages:

```
Debug Error!

Program: C:\Programs\Dev\factor.git\Debug\factor.exe

abort() has been called
```

And "The program '[10932] factor.exe' has exited with code 3 (0x3)." at the end.

Maybe I misunderstand, but it looks like debugging fails to be useful in this case, especially with the breakpoint nonsense, and the unclear reason for abort(). I was hoping to see some "OutOfMemory" or "AccessViolation" exception immediately before the termination, but no such luck. Could someone try my vocab and see if it crashes?

I have so far reproduced the crash of the 32-bit Factor build (but not 64-bit Factor) on two PCs: Win XP 32-bit, and Win 8.1 64-bit.

Edit: corrected from "VS 2016" to "VS 2015".

mrjbq7:
[...]

PR #1829, opened 2017-06-03
Nmakefile: add the /largeaddressaware linker option for 32-bit builds

AlexIljin:
This somewhat mitigates issue #1828 by allowing more memory to be allocated for the 32-bit apps on some systems. Has no effect on 32-bit Windows XP.


erg:
Merged, thanks!!

PR #1830, opened 2017-06-05
Doc fixes

AlexIljin:
I continue my documentation grooming efforts with this cosmetic contribution. All double spaces were replaced with single spaces, except where they were used for indentation. In one place a newline character was added.

The `help.lint.spaces` vocab was used to provide source data for the manual work that was performed. If you think that the vocab is not needed, feel free to skip that commit, although I'd like to keep it around and use once in a while.


mrjbq7:
Merged, thanks!

PR #1831, opened 2017-06-05
io.directories: add size slot to directory-entry tuple

AlexIljin:
This adds the file size slot to the basic tuple, and its implementation on Windows. I don't know anything about Unix systems, but I left a TODO-comment if someone wants to contribute to the Unix backend.

I'm creating a file management system, and I want to scan an entire 2 Tb drive worth of files once in a while (currently about 290'000 files). When I traverse directory entries, I always want the file size information, and it seems wasteful to make a second pass over the data set, especially since I know that the data are available from WinAPI during the first pass, and merely not published by Factor.

### DIFF
--- a/basis/io/directories/windows/windows.factor
+++ b/basis/io/directories/windows/windows.factor
@@ -4,7 +4,7 @@ USING: system io.directories alien.strings
 io.pathnames io.backend io.files.windows destructors
 kernel accessors calendar windows windows.errors
 windows.kernel32 alien.c-types sequences splitting
-fry continuations classes.struct ;
+fry continuations classes.struct windows.time ;
 IN: io.directories.windows
 
 M: windows touch-file ( path -- )
@@ -58,7 +58,7 @@ M: windows delete-directory ( path -- )
         ] unless drop f
     ] when ;
 
-TUPLE: windows-directory-entry < directory-entry attributes ;
+TUPLE: windows-directory-entry < directory-entry attributes size ;
 
 C: <windows-directory-entry> windows-directory-entry
 
@@ -67,8 +67,10 @@ C: <windows-directory-entry> windows-directory-entry
     [
         dwFileAttributes>>
         [ win32-file-type ] [ win32-file-attributes ] bi
-    ] bi
-    dupd remove <windows-directory-entry> ; inline
+        dupd remove
+    ]
+    [ [ nFileSizeLow>> ] [ nFileSizeHigh>> ] bi >64bit ] tri
+    <windows-directory-entry> ; inline
 
 M: windows (directory-entries) ( path -- seq )
     "\\" ?tail drop "\\*" append


basis/io/directories/unix/unix.factor
```diff
@@ -67,7 +67,8 @@ M: unix copy-file ( from to -- )
 
 : >directory-entry ( dirent* -- directory-entry )
     [ d_name>> alien>native-string ]
-    [ d_type>> dirent-type>file-type ] bi
+    [ drop f ] ! TODO: set size slot
```
mrjbq7: I don't think ``dirent`` has a size slot, which is unfortunate for linux/mac.

AlexIljin: If so, I could put the size slot in the windows-directory-entry. Would that be better?

mrjbq7: That might be better?  Dunno.

mrjbq7: Maybe add to windows and we'll make a TODO to generalize if makes sense.  I think on linux/mac you need to make another system call (``stat``) to get the file sizes while iterating, which might be worth doing always, but might not...

AlexIljin:
OK, I have redone the PR. Now it contains the extra slot only on the Windows platform. That's good enough for me for now.

catb0t:
On POSIX-likes, the size of a directory as returned by stat(2) is the length of its metadata. To get its size as in the number of bytes used by all its children inodes, you need to `fstat` each dentry individually ([Where does ext4 store directory sizes?](https://unix.stackexchange.com/q/124979/136107)) and sum the sizes. 

If I can make a decent implementation I'll try to create a pull for `unix` unless someone else beats me out.

mrjbq7:
We already support `file-info` that does an fstat if you want. What I was saying is that we might not want to automatically do an fstat on directory iteration.  It's free on windows which is nice though!

AlexIljin:
@catb0t: There is definitely no need to go into subdirectories to find out the size of a directory, that'll just create new overhead and add complexity to the low-level mechanism we are discussing. The only thing needed is the size for the files. For directories the value may be `f` or zero.

On Windows the size for the directories is always zero. Only for files it has a useful value.

mrjbq7:
Merged, thanks!

Unfortunate that your code becomes non-portable if it depends on a ``size`` slot, but maybe you can do something temporary (sort of hackish) like:

```factor
: my-file-size ( directory-entry -- size )
    dup class-of directory-entry eq? [ name>> file-info size>> ] [ size>> ] if ;
```

mrjbq7:
Also, note that maybe ``link-info`` is more correct for some use-cases than ``file-info`` when encountering symbolic links.

mrjbq7:
Also @catb0t we have ``directory-size`` in ``io.directories.search`` that traverses a directory structure and returns "total size of files".  You might play with that if you're thinking about a feature along those lines and see if it works for you.

mrjbq7:
P.S. @AlexIljin another approach (more verbose, but maybe more explicit and cross platform) would be to have code like this:

```factor
IN: my-vocab

GENERIC: my-file-size ( directory-entry -- size )

M: directory-entry my-file-size link-info size>> ;

os windows? [ "my-vocab.windows" require ] when
```

and

```factor
IN: my-vocab.windows

M: windows-directory-entry my-file-size size>> ;
```

AlexIljin:
@mrjbq7: `dup class-of directory-entry eq?`

That's interesting... You are checking if the `directory-entry` is _not_ subclassed, which would mean it's _not_ a unix version, therefore a windows one. Clever. A bit brittle, I'm afraid, in case unix version gets subclassed.

Of course, the problem we are solving is how not to import implementation vocabs (to check directly for `windows-directory-entry?`) and keep the high-level code cross-platform. Which brings me to the wonderful discovery, which would not be possible in many other OO-languages. In my code I have used a slot accessor without importing the class definition (and the implementation module) of its origin. Blows my mind. That's why I like Factor, and maybe it's one of the ways all the time spent studying it will eventually pay off. I'm starting to think differently (wink, Apple, wink-wink).

I mean, I have used that possibility without even breaking my stride, and only when I stopped to think about it now I noticed that it would not have been possible in Delphi, C++ and C#.

In your P.S. I see the code that I would prefer. Thank you for the suggestion!

mrjbq7:
Alternatively (if you need one), you could argue for adding this to the stdlib:

```factor
M: directory-entry size>> link-info size>> ;
```

And then document that calling it involves additional system call on non-Windows platforms.

PR #1832, opened 2017-06-05
Add .gitattributes to enforce EOL policy for factor sources

AlexIljin:
There were times when I committed files with Windows EOL by mistake. This file would have prevented the history rewrites that I had to do.


mrjbq7:
Merged, thanks!

PR #1833, opened 2017-06-07
Improve Inspector performance

AlexIljin:
Given the code `200000 0 <array> ui.tools.inspector:inspector`, this patch improves the time it takes to open the Inspector window from over 6 seconds (5-11 seconds was the range of measurements) down to about 1 second.

The main bottleneck was the run through all the rows in a table to measure the maximum width for the Key column. This was especially useless when inspecting arrays and other sequences, where the Key is always the item index. The time was proportional to the sequence length.

With this patch it takes O(1) to measure the Key column width, and then the fixed value is used. Also, for some reason, the rendering of the mouse-over highlight becomes instantaneous instead of sluggish. Same for scrolling and general redrawing of the table.

I do have some questions left over. To measure the column width I have used the `text-dim` word, but conceptually it would be better to use `row-column-widths`. The latter considers the padding returned by the `cell-dim`, while the former does not. The only reason I didn't go with the `row-column-widths` was that it's in the private section of the `ui.gadgets.tables` vocab, and I wanted to respect the interface boundaries. What shall we do? Shall we make `row-column-widths` a part of the public interface? Shall we make the text padding value (currently zero) externally available in some way, like store it in a `SYMBOL`? Or shall we just assume that it will never be anything other than zero for the text strings? I can implement whichever you think is best.

### DIFF
--- a/basis/ui/gadgets/tables/tables.factor
+++ b/basis/ui/gadgets/tables/tables.factor
@@ -39,6 +39,7 @@ single-click?
 { hook initial: [ drop ] }
 { gap initial: 2 }
 column-widths total-width
+fixed-column-widths
 focus-border-color
 mouse-color
 column-line-color
@@ -101,8 +102,12 @@ M: image-name draw-cell nip draw-image ;
 
 : compute-column-widths ( table -- total widths )
     dup rows>> [ drop 0 { } ] [
-        [ drop gap>> ] [ initial-widths ] [ ] 2tri
-        [ row-column-widths vmax ] with each
+        over fixed-column-widths>> [
+            nip [ gap>> ] dip
+        ] [
+            [ drop gap>> ] [ initial-widths ] [ ] 2tri
+            [ row-column-widths vmax ] with each
+        ] if*
         [ compute-total-width ] keep
     ] if-empty ;
 
--- a/basis/ui/tools/inspector/inspector.factor
+++ b/basis/ui/tools/inspector/inspector.factor
@@ -109,7 +109,8 @@ M: inspector-gadget focusable-child*
     table>> ;
 
 : com-refresh ( inspector -- )
-    model>> notify-connections ;
+    [ model>> notify-connections ] keep
+    [ table>> ] [ model>> ] bi fix-column-widths drop ;
 
 : com-push ( inspector -- obj )
     control-value ;
--- a/basis/ui/tools/inspector/inspector.factor
+++ b/basis/ui/tools/inspector/inspector.factor
@@ -72,17 +72,13 @@ M: object make-slot-descriptions
 M: hashtable make-slot-descriptions
     call-next-method [ key-string>> ] sort-with ;
 
-! If model is a sequence, get its maximum index, measure its width
-! and use that as the first column width (or the first column title
-! width, whichever is greater). This improves performance when
-! inspecting big arrays.
+! If model is an enum, get its maximum index, measure its width
+! rendered in the table, and use that as the first column width (or
+! the first column title width, whichever is greater). This
+! improves performance when inspecting big arrays.
 : first-column-width ( table model -- width )
-    value>> dup sequence? [
-        length 1 - 1array
-    ] [
-        make-mirror keys
-    ] if [ unparse-short ] map
-    over renderer>> column-titles first suffix
+    value>> make-mirror dup enum? [ length 1 - 1array ] [ keys ] if
+    [ unparse-short ] map over renderer>> column-titles first suffix
     row-column-widths supremum ;
 
 : <inspector-table> ( model -- table )
--- a/basis/ui/tools/inspector/inspector.factor
+++ b/basis/ui/tools/inspector/inspector.factor
@@ -1,14 +1,15 @@
 ! Copyright (C) 2006, 2009 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors colors inspector namespaces kernel models fry
-colors.constants models.arrow prettyprint sequences mirrors
-assocs classes io io.styles arrays hashtables math.order sorting
-refs fonts ui.tools.browser ui.commands ui.operations ui.gadgets
+USING: accessors arrays assocs classes colors colors.constants
+combinators fonts fry hashtables inspector io io.styles kernel
+math math.order math.parser mirrors models models.arrow
+namespaces prettyprint refs sequences sorting ui ui.commands
+ui.gadgets ui.gadgets.buttons ui.gadgets.labeled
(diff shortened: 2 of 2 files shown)


mrjbq7:
So it returns 0 for the width of the value column?

AlexIljin:
Yes, but the second column stretches to the window size, so it doesn't matter.

mrjbq7:
I'm okay with having it use ``cell-dim`` from ``ui.gadgets.tables.private``.  It's fairly tied to its implementation and at least in this case if we change how one works we can see that it's used this way in the same codebase.

Another point, I wonder if you can get a general speedup by the approach of always fixing the column widths (by doing what you do with sequences, but doing the general approach of computing the key widths of the slot descriptions and then leaving ``0`` for the value width).

AlexIljin:
@mrjbq7: Yes, I could do a pass over the Key column contents to find out the maximum width, and use that as a fixed value. I guess, theoretically that would halve the time it takes to calculate the initial `pref-dim`, and the Inspector window should show up faster. It's difficult to measure, though. Both `benchmark` and `profile` are pretty much useless, because all the initial calculations (`pref-dim`) are happening in a created window's thread.

This code `[ 200000 0 <array> ui.tools.inspector:inspector ] benchmark` will show a fraction of a second, even though the UI hangs for several seconds.

Side note: I think I'll return to this task on the weekend, got some urgent stuff to do in the meantime.

AlexIljin:
Here are two implementations. One is easier to read, but the other doesn't generate intermediate arrays. Both work. What do you think?

https://gist.github.com/AlexIljin/f453f86594e61741b14c0492084cfbe8
https://gist.github.com/AlexIljin/aac4f5b40c214ec96f4fe12a8caeea5e

mrjbq7:
I kind of like the simpler one.

AlexIljin:
I have updated the PR. I like the simpler code as well, because it's much more readable.

mrjbq7:
One comment, maybe, is that your ``sequence`` check depends on what ``mirrors:make-mirror`` returns.  So, if we end up making a custom mirror for ``string`` (as an example), then this code would be wrong because it would be inconsistent with the ``mirrors`` vocabulary.

mrjbq7:
Also, one problem with this whole approach might be if the object changes underneath, you aren't re-calculating the column widths...

AlexIljin:
@mrjbq7, I think I addressed both of your concerns. What do you think?

mrjbq7:
Hmm, so your approach has the disadvantage of running ``[ unparse-short ] map`` twice, once for your keys in the ``fix-column-widths`` calculation and once again for the ``make-slot-descriptions``.

How about something like this by making ``compute-column-widths`` in ``ui.gadgets.tables.private`` a generic?

```diff
diff --git a/basis/ui/gadgets/tables/tables.factor b/basis/ui/gadgets/tables/tab
les.factor
index 1d1e68b30c..0dc5ba0db5 100644
--- a/basis/ui/gadgets/tables/tables.factor
+++ b/basis/ui/gadgets/tables/tables.factor
@@ -99,7 +99,9 @@ M: image-name draw-cell nip draw-image ;
 : compute-total-width ( gap widths -- total )
     swap [ column-offsets drop ] keep - ;
 
-: compute-column-widths ( table -- total widths )
+GENERIC: compute-column-widths ( table -- total widths )
+
+M: table compute-column-widths
     dup rows>> [ drop 0 { } ] [
         [ drop gap>> ] [ initial-widths ] [ ] 2tri
         [ row-column-widths vmax ] with each

diff --git a/basis/ui/tools/inspector/inspector.factor b/basis/ui/tools/inspecto
r/inspector.factor
index d2af81ede1..a2ec4b146a 100644
--- a/basis/ui/tools/inspector/inspector.factor
+++ b/basis/ui/tools/inspector/inspector.factor
@@ -71,8 +71,22 @@ M: object make-slot-descriptions
 M: hashtable make-slot-descriptions
     call-next-method [ key-string>> ] sort-with ;
 
+TUPLE: inspector-table < table ;
+
+! Improve performance for big arrays or large hashtables by
+! only calculating column width for the longest key.
+M: inspector-table compute-column-widths
+    dup rows>> [ drop 0 { } ] [
+        [ drop gap>> ]
+        [ initial-widths ]
+        [ keys longest "" 2array ] 2tri
+        row-column-widths vmax
+        [ compute-total-width ] keep
+    ] if-empty ;
+
 : <inspector-table> ( model -- table )
-    [ make-slot-descriptions ] <arrow> inspector-renderer <table>
+    [ make-slot-descriptions ] <arrow> inspector-renderer
+    inspector-table new-table
         [ invoke-primary-operation ] >>action
         monospace-font >>font
         line-color >>column-line-color
```

This should be faster than your way for large hashtables:

```200,000 <iota> dup H{ } zip-as ui.tools.inspector:inspector```

But maybe ever-so-slightly slower for your sequence case (but really ``keys longest`` is fast anyway on an ``<enum>``).

**Comment by mrjbq7** [comment]
Committed in e8cef244860d750ce8a797bd3cefe1f9d840dfd8.  Thank you!!

(Any further improvements or suggestions are welcome!)

**Comment by erg** [comment]
Can we fix this for the main UI too?

``200,000 <iota> describe`` takes way too long.

**Comment by mrjbq7** [comment]
That's slow also for other reasons, like printing 400,000 cells...

**Comment by mrjbq7** [comment]
One thing you/we could look into is improving ``unparse-short`` performance:

``
200,000 <iota> [ [ unparse-short ] map length ] profile
```

mrjbq7:
```factor
IN: scratchpad 10,000,000 <byte-array> gc [ [ unparse-short ] { } map-as ] time
Running time: 20.247326939 seconds
```

catb0t:
20s for 10 million items? Isn't that pretty good?

erg:
```
10,000,000 <byte-array> gc [ [ 10 >base ] { } map-as ] time
Running time: 1.078117906 seconds

Additional information was collected.
dispatch-stats.  - Print method dispatch statistics
gc-events.       - Print all garbage collection events
gc-stats.        - Print breakdown of different garbage collection events
gc-summary.      - Print aggregate garbage collection statistics
```

Side note, we have ``dec>`` but not ``>dec``? And maybe it should be ``decimal>`` and ``>decimal``...

mrjbq7:
+1 for deleting ``dec>``.

erg:
Doing it!

mrjbq7:
Oh right, we needed ``dec>`` for parsing decimal strings with no radix char.

mrjbq7:
Although ``dec>`` is only used in one place in codebase (``html.entities``).

PR #1838, opened 2017-06-22
Supremum infimum

AlexIljin:
As discussed in the mailing list, here they are: `?infimum` and `?supremum`, complete with tests and documentation.

I'm not sure about the last commit, because it introduces dependency from `basis` on `extra`, so feel free to skip it.

### DIFF
--- a/extra/sequences/extras/extras.factor
+++ b/extra/sequences/extras/extras.factor
@@ -559,6 +559,16 @@ PRIVATE>
 : infimum-by* ( ... seq quot: ( ... elt -- ... x ) -- ... i elt )
     [ before? ] select-by* ; inline
 
+: ?supremum ( seq/f -- elt/f )
+    [ f ] [
+        [ ] [ 2dup and [ max ] [ dupd ? ] if ] map-reduce
+    ] if-empty ;
+
+: ?infimum ( seq/f -- elt/f )
+    [ f ] [
+        [ ] [ 2dup and [ min ] [ dupd ? ] if ] map-reduce
+    ] if-empty ;
+
 : change-last ( seq quot -- )
     [ drop length 1 - ] [ change-nth ] 2bi ; inline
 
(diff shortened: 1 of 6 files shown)


extra/sequences/extras/extras.factor
```diff
@@ -559,6 +559,18 @@ PRIVATE>
 : infimum-by* ( ... seq quot: ( ... elt -- ... x ) -- ... i elt )
     [ before? ] select-by* ; inline
 
+: ?supremum ( seq/f -- elt/f )
```
mrjbq7: You don't need the ``dup [ ... ] when`` because ``f`` is defined as an empty sequence.

So this would work:

```factor
: ?supremum ( seq -- elt/f )
    [ f ] [ [ ] [ 2dup and [ max ] [ dupd ? ] if ] map-reduce ] if-empty ;
```

catb0t:
These words, and especially the PR title, really sound like you're trying to cast a spell. ***Abracadabra, supremum infimum solitus monoculous!!***

AlexIljin:
@mrjbq7: good point, I have updated the PR.

@catb0t: : ))

mrjbq7:
Cherry-picked a9cb06705c85bb1d6f469c9937ad0e71e031e7e1, but not ready to have ``basis`` depend on ``extras`` yet.

AlexIljin:
235c4fb is also in extra, you could've picked it too.

bjourne:
picked!

mrjbq7:
I'm mostly fine with merging 235c4fb3e2a3fd37a240a90a109b74b0aa603b60 but didn't because of two reasons, one I thought ``best-vreg`` was a nice documenting name for what that was doing, and two I wasn't sure I wanted to hide potential bugs (putting ``f`` in the array that is being supremum'd and not noticing because of the new behavior of ignoring ``f`` in ``?supremum``).

jonenst:
>  two I wasn't sure I wanted to hide potential bugs (putting f in the array that is being supremum'd and not noticing because of the new behavior of ignoring f in ?supremum).

+1. I think this should be reverted. The commit implies that the use of ?supremum is equivalent to the previous code, but it is not.

mrjbq7:
@jonenst okay, well thats good enough for me then, since you concur.  I'll revert.

bjourne:
Alright. I thought you had just forgotten that commit.

mrjbq7:
No worries!!

PR #1846, opened 2017-07-16
checksums: add ripemd-160 (based on the md5 implementation)

jonenst:
Hi, I added checksums.ripemd . The code is pretty verbose because I tried to keep the same structure as md5. I added it because I want to use it in https://rosettacode.org/wiki/Bitcoin/public_point_to_address and https://rosettacode.org/wiki/RIPEMD-160


mrjbq7:
Merged, thanks!!

PR #1850, opened 2017-08-17
windows.dragdrop-listener is revived

AlexIljin:
Hello!

  I have brought that vocab from the unmaintained and added some documentation.
  It works now.


mrjbq7:
Awesome!

Maybe someday we can make it cross-platform, split out the dragdrop support from the listener wrapping, and maybe document some examples!  But so glad to see unmaintained stuff become maintained!

PR #1865, opened 2017-10-13
ui.tools.listener: update interactor's line-height in set-listener-font

AlexIljin:
If you increase the font size in the listener without this fix, and then enter multi-line text in the listener's interactor, the lines of the (bigger) text will overlap, because their height is considered to be the old (smaller) size.


mrjbq7:
Good idea, merged, thanks!!

Issue #1742, opened 2016-11-07
at and of, nth and ???


bjourne:
Something I've felt missing for a while is something like `of` but for sequences. Often in your code, you have the element index last and the sequence itself second-to-last. For example, if you do something with an array and then want to pick the fifth element:

```factor
100 sample ... [ sq ] map ... 5 swap nth
```
So if you had something like `ath` (bad name!) you could write `5 ath` and avoid the swap. Grepping through the code `git grep -e "\bswap nth\b"` shows that the `swap nth` idiom is very common.

mrjbq7:
It does look like (in a cursory inspection) that something like 25% of ``nth`` uses are ``swap nth``.  It would be good to provide such a word, but also maybe we could imagine some other ways to fix the general problem of mis-ordered stack arguments.

nicolas-p:
I use `swap member?` all the time.

catb0t:
Just looking for uses of `swap` in my own code, it seems that all, with a few exceptions, are related to `swap nth`, `swap member?` and `swap at`, which is not surprising to me -- usually I take pleasure in reFactoring to not use `dup`, `swap`, etc and so I'm left with the unfortunate and forgettable argument ordering to `nth` and friends.

PR #1869, opened 2017-10-17
io.stream.limited: swap limit-stream parameters

AlexIljin:
This order of the `limit-stream` parameters (i.e. `limit` then `stream`) seems more logical and simplifies code.

### DIFF
--- a/basis/io/streams/limited/limited-docs.factor
+++ b/basis/io/streams/limited/limited-docs.factor
@@ -5,14 +5,14 @@ IN: io.streams.limited
 
 HELP: <limited-stream>
 { $values
-     { "stream" "an input stream" } { "limit" integer }
+     { "limit" integer } { "stream" "an input stream" }
      { "stream'" "an input stream" }
 }
 { $description "Constructs a new " { $link limited-stream } " from an existing stream. User code should use " { $link limit-stream } " or " { $link limited-input } "." } ;
 
 HELP: limit-stream
 { $values
-     { "stream" "an input stream" } { "limit" integer }
+     { "limit" integer } { "stream" "an input stream" }
      { "stream'" "a stream" }
 }
 { $description "Changes a decoder's stream to be a limited stream, or wraps " { $snippet "stream" } " in a " { $link limited-stream } "." }
@@ -21,7 +21,7 @@ HELP: limit-stream
     { $example
         "USING: accessors continuations io io.streams.limited"
         "io.streams.string kernel prettyprint ;"
-        "\"123456\" <string-reader> 3 limit-stream"
+        3 "\"123456\" <string-reader> limit-stream"
         "100 swap stream-read ."
         "\"123\""
     }
--- a/basis/io/streams/limited/limited-tests.factor
+++ b/basis/io/streams/limited/limited-tests.factor
@@ -86,21 +86,21 @@ IN: io.streams.limited.tests
 
 { t }
 [
-    "abc" <string-reader> 3 limit-stream unlimit-stream
+    3 "abc" <string-reader> limit-stream unlimit-stream
     "abc" <string-reader> =
 ] unit-test
 
 { t }
 [
-    "abc" <string-reader> 3 limit-stream unlimit-stream
+    3 "abc" <string-reader> limit-stream unlimit-stream
     "abc" <string-reader> =
 ] unit-test
 
 { t }
 [
     [
-        "resource:LICENSE.txt" utf8 <file-reader> &dispose
-        3 limit-stream unlimit-stream
+        3 "resource:LICENSE.txt" utf8 <file-reader> &dispose
+        limit-stream unlimit-stream
         "resource:LICENSE.txt" utf8 <file-reader> &dispose
         [ decoder? ] both?
     ] with-destructors
--- a/basis/io/streams/limited/limited.factor
+++ b/basis/io/streams/limited/limited.factor
@@ -9,34 +9,33 @@ IN: io.streams.limited
 TUPLE: limited-stream stream count limit current start stop ;
 INSTANCE: limited-stream input-stream
 
-: <limited-stream> ( stream limit -- stream' )
+: <limited-stream> ( limit stream -- stream' )
     limited-stream new
-        swap >>limit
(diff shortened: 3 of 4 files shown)


AlexIljin:
Here's another example that works better with this parameter order:
```
: overwrite-with-random-bytes ( file-name -- )
    [ file-info size>> ] keep binary [
        <random-stream> limit-stream
        0 seek-absolute output-stream get
        [ stream-seek ] keep stream-copy
    ] with-file-appender ;
```

AlexIljin:
The long-winded hard-to-see examples are in the help and in the tests, i.e. artificial. The real use cases, I would argue, are in `M: decoder limit-stream`, `limited-input ( limit -- )`, `with-limited-input ( limit quot -- )`, `overwrite-with-random-bytes ( file-name -- )`, and `random-file ( n path -- )`, all of which got simpler.

There was also some code in unmaintained: `data>> [ length ] [ binary <byte-reader> ] bi limit-stream`. It works equally well with any parameter order, but I would argue that the name `limit-stream` serves as a hint of what the parameters are, namely `limit`, then `stream`.

There are no more uses of the word in all the codebase, so this represents an exhaustive argument, in my opinion.

Of course, there may (or may not) be user code that'll be broken by this change.

jonenst:
> Of course, there may (or may not) be user code that'll be broken by this change.

Yeah, not sure it's worth it.

bjourne:
I also prefer the current order. To me it makes sense if the "fatter" argument comes before the lighter ones. And the stream itself is fatter than the limit to it, which is just a number. But I hope you don't get discouraged Alex! Your other pr:s have been really great.

AlexIljin:
Thanks for the encouragement, @bjourne, I really appreciate it!
I decided to keep the modified order in my personal Factor branch, because it makes sense to me.
As to the "fatter, then lighter" argument, it also makes a lot of sense, actually. It really does.
I wonder why it's so rarely the case. `nth` and `set-nth` immediately jump to mind...

AlexIljin:
`at` and `set-at` closely follow...

bjourne:
It could be inertia. For a long time Factor could only do generic dispatch on the topmost stack item. In my own code, I've often wished for `nth` and `set-nth` with swapped parameter order: #1742

erg:
There's also the issue of ``dropping`` vs ``preserving`` in stack effects. The dropping version can only modify the existing object, while the preserving one can either return the existing object or a new one. (Which? You'd need an effects system, or conventions or something to fix it properly.)

```
: nth ( n seq -- obj ) ... ; ! dropping (drops the seq)
: nth ( seq n -- seq obj ) ... ; ! preserving

: set-at ( obj key assoc -- ) ... ; ! dropping
: set-at ( assoc obj key -- assoc ) ... ; ! preserving, with same or new object?
```

With ``limit-stream``, we have the preserving one right now. The ``dropping`` version doesn't make sense here unless since ``limit-stream`` wraps an existing stream with a ``limited-stream``.

```
: limit-stream ( limit stream -- ) .. ; ! dropping
: limit-stream ( stream limit -- stream' ) ... ; ! preserving, with new object
```

Maybe we could have a convention for naming these, maybe ``nth`` vs ``nth*`` if we did a massive renaming...

AlexIljin:
It seems that currently `nth*` means "nth from the end", as opposed to `nth` meaning "nth from beginning".
That convention is followed in words like `head*`, `tail*` and their `slice` variants.
In other contexts the `*` suffix means the method to be overridden by custom implementations, as opposed to the normal entry point.

I think the preserving versions could use the same apostrophe we currently use in the stack effects:
```
: limit-stream ( limit stream -- ) .. ; ! dropping
: limit-stream' ( stream limit -- stream' ) ... ; ! preserving, with new object
```
For specifying modification of an existing object we normally use the exclamation point: `append` vs. `append!`.
So, if it were possible to somehow make a stream limited without creating a wrapper object, we could do something like this:
```
: limit-stream ( stream limit -- ) ... ; ! dropping
: limit-stream' ( stream limit -- stream' ) ... ; ! preserving, with new object
: limit-stream! ( stream limit -- ) ... ; ! dropping, modifying an existing object
: limit-stream!' ( stream limit -- stream' ) ... ; ! preserving, with same object
```

Another option we have in practice is to suffix "-1" (as in "raised to the power of -1") to words that are somehow inverse of their regular variants. Maybe `tail*` should be `tail-1` or even `1-tail`.

Or maybe `1/tail`? Just thinking out loud here.

Maybe preserving with new objects could be designated like this:
```
: +limit-stream' ( stream limit -- stream' ) ... ; ! preserving, with new object
```
Or simply `+limit`? I think this could be nicely readable in the code.
```
: nth ( n seq -- obj ) ... ; ! dropping (drops the seq)
: nth' ( seq n -- seq obj ) ... ; ! preserving
: +nth ( seq n -- seq obj ) ... ; ! preserving

: set-at ( obj key assoc -- ) ... ; ! dropping
: set-at' ( assoc obj key -- assoc' ) ... ; ! preserving, with same or new object?
: +set-at ( assoc obj key -- assoc' ) ... ; ! preserving, with same or new object
: +set-at! ( assoc obj key -- assoc' ) ... ; ! preserving, with same object
```

AlexIljin:
The conflict here to some extent is in the two use cases that we create while programming. In the one case we want data to "flow", so we have a "big" thing on the stack, and we add smaller things to operate on it. That's when we want "nth" to be `( seq n -- obj )` and "set-nth" to be `( seq obj n -- )`, i.e. "fatter, then lighter".
In the other case we have some sort of a global symbol or a named variable, from/to which we want to push data from time to time. Then we work to create a small thing (like an index) in a working context, then we want to summon the "big" thing immediately before the "nth" call: `<some index computations> my-seq get-global nth`.

I don't see it possible that one use case is always convenient to reduce to the other. Neither do I think it's possible or convenient to get rid of one of them. So, to be a practical language, I think we need two variants of `nth` and other fundamental words.

As to the preserving vs. dropping - that's something I always assumed could be resolved on the outer level. Unless dropping is extra work, we should drop, and the caller can always do `keep` or something if s/he needs it.

PR #1870, opened 2017-10-17
Doc fixes

AlexIljin:
A few improvements.

(GitHub shows this as an old PR, but I've just pushed a bunch of stuff here, so it's supposed to be new. Not sure what went wrong, maybe the old PR was never closed when it was merged. Was it merged?)

### DIFF
--- a/basis/concurrency/distributed/distributed-docs.factor
+++ b/basis/concurrency/distributed/distributed-docs.factor
@@ -43,7 +43,9 @@ $nl
 "response to a distributed message. When an instance of " { $link thread } " "
 "is marshalled, it is converted into an instance of " { $link remote-thread }
 ". The receiver of this can use it as the target of a " { $link send }
-", " { $link send-synchronous } " or " { $link reply-synchronous } " call." ;
+", " { $link send-synchronous } " or " { $link reply-synchronous } " call."
+$nl
+"Note: " { $link send-synchronous } " can only work if " { $link local-node } " is assigned (use " { $link start-node } "), because there must be a server for the remote instance to send its reply to." ;
 
 ARTICLE: "concurrency.distributed" "Distributed message passing"
 "The " { $vocab-link "concurrency.distributed" } " implements transparent distributed message passing, inspired by Erlang and Termite." $nl
--- a/basis/concurrency/distributed/distributed-docs.factor
+++ b/basis/concurrency/distributed/distributed-docs.factor
@@ -14,13 +14,13 @@ ARTICLE: "concurrency.distributed.example" "Distributed Concurrency Example"
 "The code to run the server is:"
 { $code
   "USING: io.servers ;"
-  "9000 local-server <node-server> start-server drop"
+  "9000 local-server start-node"
 }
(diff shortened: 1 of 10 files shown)


core/syntax/syntax-docs.factor
```diff
@@ -671,16 +671,16 @@ HELP: NAN:
 HELP: GENERIC:
 { $syntax "GENERIC: word ( stack -- effect )" }
 { $values { "word" "a new word to define" } }
-{ $description "Defines a new generic word in the current vocabulary. Initially, it contains no methods, and thus will throw a " { $link no-method } " error when called." } ;
+{ $description "Defines a new generic word in the current vocabulary. The word dispatches on the topmost stack element. Initially it contains no methods, and thus will throw a " { $link no-method } " error when called." } ;
```
mrjbq7: What do you mean by "dispatching on the topmost stack element"?

AlexIljin: I mean that it takes the top element on the stack, tests its class and calls the appropriate method based on the class. You know, [method dispatch](https://en.wikipedia.org/wiki/Dynamic_dispatch).

basis/concurrency/distributed/distributed-docs.factor
```diff
@@ -14,22 +14,22 @@ ARTICLE: "concurrency.distributed.example" "Distributed Concurrency Example"
...
-  "9000 local-server <node-server> start-server drop"
+  "9000 local-server start-node"
 }
 "The code to start the thread is:"
 { $code
     "USING: concurrency.messaging threads ;"
     ": log-message ( -- ) receive . flush log-message ;"
-    "[ log-message ] \"logger\" spawn dup name>> register-remote-thread"
+    "[ log-message ] \"logger\" [ spawn ] keep register-remote-thread"
 }
```
bjourne: Isn't this example wrong? `register-remote-thread` has stack effect `( thread name -- )` but you pass it two thread instances.

bjourne:
Except for the nit I had about the example in distributed-docs.factor, the PR looks good to me!

basis/concurrency/distributed/distributed-docs.factor
AlexIljin: Umm... See the `keep` call? It keeps the "logger" string, doesn't it? Which is the thread's name... Am I right?

mrjbq7:
Merged, thanks!!

basis/concurrency/distributed/distributed-docs.factor
bjourne: You are right. My factor skills are getting rusty.

jonenst: I like
```factor
[ log-message ] "logger" [ spawn ] [ register-remote-thread ] bi
```
for symmetry but that's just style :)

AlexIljin: I can understand your argument, but most times I prefer shorter text, even if it saves only a couple of characters. Maybe I'll change my preference over time.

Issue #1087, opened 2014-07-04
make python binding compatible with python3


erg:
On Arch linux, the library is called `libpython3.so` instead of `libpython3.0.so`. This will be patched shortly using `find-library*`

However, there is another problem -- in Python 3, all of the `PyString_*` etc are renamed to `PyBytes_`, `PyInt_` was renamed to `PyLong_`. So, if we want to support Python 2 and Python 3, we should break out the functions into two vocabularies and have `extra/python2` and `extra/python3` imo.

http://stackoverflow.com/questions/8229597/embedding-python

mrjbq7:
Updated the Python bindings to Python 3.

PR #1871, opened 2017-10-22
compatible with python3

sam0410:
#1087


bjourne:
In this pr, you have essentially duplicated most of the python vocabs and made new ones in python3. I don't agree with that approach. Instead, I think there should be a vocab tree called `python` in which common code is collected. Then there should be two vocabs `python2` and `python3` containing what is different in each api.

PR #1872, opened 2017-10-22
use lists vocab in factor handbook

twopir:
Without the `lists` vocabulary being used, the help page causes a traceback.

`help-lint-all` caught this error, for what it's worth.


mrjbq7:
Merged, thanks!!

PR #1873, opened 2017-10-22
Unicode Control Pictures

twopir:
I suppose this is best illustrated with an example:
```
IN: scratchpad 48 <iota> >string control-pictures print
␀␁␂␃␄␅␆␇␈␉␊␋␌␍␎␏␐␑␒␓␔␕␖␗␘␙␚␛␜␝␞␟ !"#$%&'()*+,-./
```


mrjbq7:
That's cool.  Merged, thanks!!

jonenst:
Maybe we can merge $url and this new $url-to ?
$url only uses the first element of the array right now, so it can be changed in a backwards compatible manner to use the first two elements (url and optinal label).

something like that ?
```factor
: label-url ( children -- label url )
    [ dup length 2 >= [ second ] [ first ] if ] [ first >url ] bi ;
: $url2 ( children -- )
    label-url [
        dup present href associate url-style get assoc-union
        [ write-object ] with-style
    ] ($span) ;
```

mrjbq7:
Yes! I noticed that too and made this commit:

https://github.com/factor/factor/commit/1006b4afc4c5ec89fcf6a39f348add4067635147

jonenst:
Cool !
`over [ nip dup ] unless`  could be `[ or ] keep` now that I think of it.

PR #1898, opened 2018-01-04
extra/cap: fix a bug where screenshots were incorrectly downscaled

### LINKED ISSUE #1888: cap vocabulary has too small of resolution
```factor
USE: cap
ui-windows get-global random second screenshot.
```

capture

catb0t:
closes #1888, screenshots are the native resolution now
cap

### DIFF
--- a/extra/cap/cap.factor
+++ b/extra/cap/cap.factor
@@ -6,6 +6,8 @@ models namespaces opengl opengl.gl sequences ui ui.gadgets
 ui.gadgets.worlds ;
 IN: cap
 
+<PRIVATE
+
 : screenshot-array ( world -- byte-array )
     dim>> [ first 4 * ] [ second ] bi
     [ gl-scale ] bi@ * >fixnum <byte-array> ;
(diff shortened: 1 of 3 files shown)


mrjbq7:
Hmm, do we want to set ``2x`` when ``gl-scale-factor`` is 2.0?

catb0t:
i'm no longer sure this code is optimal, how do you like `when* ] when*`?

extra/cap/cap.factor
```diff
@@ -24,7 +24,10 @@ IN: cap
     [ glReadPixels ] keep ;
 
 : screenshot ( window -- bitmap )
-    [ <image> t >>2x? ] dip
+    [ <image>
+        gl-scale-factor get-global
+        [ >integer 2 = [ >>2x? ] when* ] when*
```
mrjbq7: if you wanted it a little cleaner, this could probably be:

```factor
gl-scale-factor get-global [ 2.0 = >>2x? ] when*
```

catb0t: oh yeah, of course

catb0t: will `gl-scale-factor` always be `float`?

catb0t:
`screenshot` was cleaned up and two of the words are moved to cap.private, as well as some documentation for this vocab. i was considering a separate PR but it seems trivial enough to include here, i can revert though

mrjbq7:
Merged, thanks!!

PR #1901, opened 2018-01-11
concurrency.distributed: with-connection

AlexIljin:
Please, have a look at this implementation. It allows a client to send multiple messages to the server in a single socket connection (as opposed to opening a connection per message). The implementation sends a special symbol `keep-connection` as the first message to make sever enter a receiving loop until the `terminate-connection` symbol is received. All messages between the two symbols are forwarded to the destination thread. Both client and server need to support the new mode of operation: the client remembers its connection stream in the new `connection` slot of the `remote-thread` tuple; the server's thread (spawned to handle the client connection) optionally enters a `loop`, which has the stream in its stack until the `loop` is terminated.

**Interoperability:** the new mode of operation is enabled by using the new `with-connection` combinator at the client end. If the combinator is not used, the implementation works as before (with per-message connections). Old client implementations can work with new server implementations without issues (the server implementation is fully backwards-compatible). New clients can work with old servers as before if the `with-connection` combinator is not used. If new clients use `with-connection` while connecting to old servers, the server's target thread will receive the `keep-connection` symbol **instead of** the intended message, and the socket connection will be terminated by the old server before the intended message can be sent by the client.
There is no possibility to overcome the interoperability limitations within the `concurrency.distributed` vocab itself, because there is no protocol versioning mechanism implemented in the vocab.

If you have any suggestions on how this implementation can be improved, I'd like to hear your feedback.

A further improvement to the vocab could be allowing *bidirectional* communication between a client and a server over the long-lived socket connection. For example, currently to do `send-synchronous`/`reply-synchronous` we need to open a server socket at the client side, which is completely unnecessary from the TCP viewpoint, as any established connection is inherently a `duplex-stream`. If you can suggest a way to implement that, or even contribute code, please do.

Another improvement could be replacing the initial `keep-connection` symbol with a tuple describing the required minimum version of the server implementation along with the desired connection options. I have little experience designing such protocols, so I would appreciate it if you point me to the relevant reading material.

PS: this implementation perfectly solves my particular use case, which is to start a client worker process, and send a bunch of data back to the server asynchronously. My test case sends almost 63500 messages over the single socket connection without breaking a sweat, whereas the previous implementation choked on a fraction of that number due to port number exhaustion and was completely CPU-bound, having to bring up and tear down TCP connections per each message. This means two things. One - it works, which is good; two - without your help I may leave this alone and not take it any further, which is not so good. So, if you have any interest in this topic, please voice it.

Also, I only tested it on Windows, so YMMV, although the code is pretty high-level so I don't expect any issues on other platforms.

### DIFF
--- a/basis/concurrency/distributed/distributed.factor
+++ b/basis/concurrency/distributed/distributed.factor
@@ -13,6 +13,17 @@ IN: concurrency.distributed
 : registered-remote-threads ( -- hash )
    \ registered-remote-threads get-global ;
 
+: thread-connections ( -- hash )
+    \ thread-connections get-global ;
+
+: get-thd-conn ( thread -- connection/f )
+    thread-connections at ;
+
+: set-thd-conn ( thread connection/f -- )
+    [ swap thread-connections set-at ] [
+        thread-connections delete-at
+    ] if* ;
+
 PRIVATE>
 
 : register-remote-thread ( thread name -- )
@@ -40,20 +51,19 @@ SYMBOL: local-node
 : start-node ( addrspec -- )
     <node-server> start-server local-node set-global ;
 
-TUPLE: remote-thread node id connection ;
+TUPLE: remote-thread node id ;
 
-: <remote-thread> ( node id -- remote-thread )
-    f remote-thread boa ;
+C: <remote-thread> remote-thread
 
 TUPLE: connection remote stream local ;
 
 C: <connection> connection
 
 : connect ( remote-thread -- )
-    dup node>> dup binary <client> <connection> >>connection drop ;
+    dup node>> dup binary <client> <connection> set-thd-conn ;
 
 : disconnect ( remote-thread -- )
-    dup connection>> [ stream>> dispose ] when* f >>connection drop ;
+    dup get-thd-conn [ stream>> dispose ] when* f set-thd-conn ;
 
 : with-connection ( remote-thread quot -- )
     '[ connect @ ] over [ disconnect ] curry [ ] cleanup ; inline
@@ -65,7 +75,7 @@ C: <connection> connection
     stream>> [ serialize flush ] with-stream* ;
 
 M: remote-thread send ( message thread -- )
-    [ id>> 2array ] [ node>> ] [ connection>> ] tri
+    [ id>> 2array ] [ node>> ] [ get-thd-conn ] tri
     [ nip send-to-connection ] [ send-remote-message ] if* ;
 
 M: thread (serialize) ( obj -- )
@@ -76,4 +86,5 @@ M: thread (serialize) ( obj -- )
 
 [
     H{ } clone \ registered-remote-threads set-global
+    H{ } clone \ thread-connections set-global
 ] "remote-thread-registry" add-startup-hook
--- a/basis/concurrency/distributed/distributed.factor
+++ b/basis/concurrency/distributed/distributed.factor
@@ -1,4 +1,5 @@
 ! Copyright (C) 2005 Chris Double. All Rights Reserved.
+! Copyright (C) 2018 Alexander Ilin.
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors arrays assocs concurrency.messaging
 continuations destructors fry init io io.encodings.binary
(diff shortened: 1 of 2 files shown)


basis/concurrency/distributed/distributed.factor
```diff
@@ -37,22 +59,46 @@ SYMBOL: local-node
 : start-node ( addrspec -- )
     <node-server> start-server local-node set-global ;
 
-TUPLE: remote-thread node id ;
+TUPLE: remote-thread node id connection ;
```
mrjbq7: I'm concerned that storing the connection here will cause it to be serialized and sent across the connection, see for example ``M\ thread (serialize)``.

basis/concurrency/distributed/distributed.factor
```diff
@@ -37,22 +59,46 @@ SYMBOL: local-node
...
+        [ terminate-connection over send ] dip stream>> dispose
+    ] when* f >>connection drop ;
+
+: with-connection ( remote-thread quot -- )
+    '[ connect @ ] over [ disconnect ] curry [ ] cleanup ; inline
 
 : send-remote-message ( message node -- )
     binary [ serialize ] with-client ;
 
+: with-stream* ( duplex-stream quot -- )
```
mrjbq7: I think this is duplicative of ``io.streams.duplex:with-stream*``...

mrjbq7:
If we're using TCP streams, I'm not sure we really need ``keep-connection`` and ``terminate-connection``.  The client can just not close the connection if it wants to keep it open.  If your application needs some kind of heartbeat message then maybe thats something that could be layered on top, but why not just have the server not close, and loop, and the old client closes and the new client doesn't if it has more messages to send?

basis/concurrency/distributed/distributed.factor
AlexIljin: Wow, you are right, this is completely redundant. I'll remove it.

AlexIljin:
If the server loops, how will it know when to stop the loop and close the connection? The old implementation deserializes a single message in a `with-stream`, then the connection is terminated from the server side. Are you saying the server can simply read until EOF or something?

mrjbq7:
If the client does ``[ send ] with-client`` then it will close the connection and the server's TCP connection would read EOF (``f`` presumably) and then exit.

basis/concurrency/distributed/distributed.factor
AlexIljin: What would you suggest? A separate global assoc to link `remote-thread`s with their `connection`s?

AlexIljin:
I'll try to implement that. The server will always `loop` until `f` is `receive`d, and then close the connection. If that works, then it'll be simpler for sure, thanks for the idea!

AlexIljin:
OK, I have updated the PR according to the comments above: dropped the redundant `with-streams*` implementation, got rid of the extra symbols and made the server always loop until the connection is closed. The code got simpler.
Also, I have removed the `connection` slot from `remote-thread`, and created a global assoc instead. The assoc stores the { thread connection } pairs until the connection is terminated.

Please, review and merge.

erg:
Nice work, merged!

mrjbq7:
I still see this:

```factor
TUPLE: remote-thread node id connection ;
```

Is this the latest code where "I have removed the connection slot from remote-thread"?

mrjbq7:
I think @erg merged too quickly?

mrjbq7:
Those commits are still from 12 days ago.

AlexIljin:
I'm terribly sorry! I made a fixup to the old commit, so as to keep the PR short. Unfortunately, I now see that only two out of three commits made it into the fixup. I'll reopen the PR with the additional commit...

mrjbq7:
No worries!

AlexIljin:
Now it's all in.

AlexIljin:
You can accept the extra commit to be "fixup" of 6f908432d184 if you want.

mrjbq7:
Okay, merged.

I also changed slightly, removing ``get-thd-conn`` and ``set-thd-conn`` since I tend to prefer full words when they are short ("thread" instead of "thd") and I thought just using the assoc was a little cleaner.

Issue #1903, opened 2018-01-13
Rename m. to m.m?


catb0t:
I'm working on giving `math.matrices` better documentation and cleaning up some of its names.

I wonder if [`m.`](http://docs.factorcode.org:8080/content/word-m.,math.matrices.html) (dot product of two matrices) should become `m.m`?

We have:
- [`v.m`](http://docs.factorcode.org:8080/content/word-v.m,math.matrices.html) and [`m.v`](http://docs.factorcode.org:8080/content/word-m.v,math.matrices.html) in `math.matrices` which compute the dot product of a matrix and a vector or vice versa
- `.`, `callstack.`, `infer.`, `simple-table.`, `leaks.` and so on which output to a stream some data
- `m-`, `m+`, `m~`, `m*`, `m/` which perform the mathematical operation on two matrices

`m.` would to me signal that some `m` is going to be prettyprinted.

Here are the vocabularies which have to change:

```
extra/project-euler/common/common
extra/rosetta-code/pythagorean-triples/pythagorean-triples
extra/rosetta-code/conjugate-transpose/conjugate-transpose
extra/koszul/koszul
extra/game/debug/tests/tests
extra/gpu/util/wasd/wasd
extra/benchmark/pidigits/pidigits
extra/benchmark/3d-matrix-scalar/3d-matrix-scalar
```

Note they are all in `resource:extra`, so this would not be particularly impactful. (13 times in 8 files)

AlexIljin:
I'm 100% behind the idea of renaming `m.` to `m.m` for consistency.
And then also for renaming `m.m` further to `m*m`, `v.m` to `v*m` and `m.v` to `m*v`.

mrjbq7:
Or maybe ``vdot`` and ``mdot``...

AlexIljin:
I agree that `v.` is strongly associated with `.`, which is print command in our language.
When I wrote my previous suggestion I thought we are doing simple matrix product (`m*m` or `m*`).
But now that I know there is the "dot product", which is a different mathematical operation, I'm thinking maybe we could use something like `v.*` for "vector dot product" and `m.*` for "matrix dot product".

bjourne:
But where do you end? :) Should `v+` be renamed `v+v` and `v-` `v-v` too?

catb0t:
[...]

PR #1908, opened 2018-01-20
math.matrices: docs, various cleanups and fixes

catb0t:
this is not ready for merge, but i'm soliciting comments anyway.

this doesn't contain #1903 (rename `m.`) yet, either, because i'm not sure where to go with that thread.

these words aren't used very much or at all in `core` and `basis`, so these changes aren't very breaking IIR.


basis/bootstrap/image/primitives/primitives-docs.factor
```diff
@@ -19,7 +19,7 @@ HELP: primitive-quot
 { $description "Creates the defining quotation for the primitive. If 'vm-func' is a string, then it is prefixed with 'primitive_' and a quotation calling that C++ function is generated." } ;
 
 ARTICLE: "bootstrap.image.primitives" "Bootstrap primitives"
-"This vocab contains utilities for declaring primitives to be added to the bootstrap image. It is used by " { $vocab-link "bootstrap.primitives" }
```
mrjbq7: I think it was (kind of) right before, it's used by the ``core/bootstrap/primitives.factor``.

```
./core/bootstrap/primitives.factor:USING: arrays assocs bootstrap.image.primitives
```

basis/math/matrices/matrices.factor
```diff
@@ -6,25 +6,49 @@ sequences.private fry math.statistics grouping
 combinators.short-circuit math.ranges combinators.smart ;
 IN: math.matrices
 
+DEFER: eye
```
mrjbq7: Why do you defer it here, but not really use it until much lower?

basis/math/matrices/matrices.factor
```diff
@@ -6,25 +6,49 @@ sequences.private fry math.statistics grouping
...
+
+ERROR: negative-power-matrix m n ;
+
 ! Matrices
-: make-matrix ( m n quot -- matrix )
+: make-matrix ( m n quot: ( ... -- elt ) -- matrix )
     '[ _ _ replicate ] replicate ; inline
 
+: make-matrix-with-indices ( m n quot -- matrix )
+    [ [ <iota> ] bi@ ] dip '[ @ ] cartesian-map ; inline
```
mrjbq7: can't ``'[ @ ]`` be skipped?

```factor
: make-matrix-with-indices ( m n quot -- matrix )
    [ [ <iota> ] bi@ ] dip cartesian-map ; inline
```

mrjbq7: Also, maybe put the stack effects ``( ... m n quot: ( ... m' n' -- ... elt ) -- ... matrix )``

basis/math/matrices/matrices.factor
```diff
@@ -6,25 +6,49 @@ sequences.private fry math.statistics grouping
 combinators.short-circuit math.ranges combinators.smart ;
 IN: math.matrices
 
+DEFER: eye
+<PRIVATE
+: keep-order ( ..a x quot: ( ..a x -- ..b ) -- ..b x )
+    [ dup ] dip call ; inline
```
mrjbq7: not a big fan because the quotation is called not on ``( ..a x -- )`` but on ``( ..a x x -- )`` since you duplicate the X before calling it?  Also the output is `` -- x ..b )``.  Better maybe to use ``keep swap`` or something.

basis/math/matrices/matrices.factor
```diff
@@ -6,25 +6,49 @@ sequences.private fry math.statistics grouping
...
 
-: eye ( m n k -- matrix )
-    [ [ <iota> ] bi@ ] dip neg '[ _ + = 1 0 ? ] cartesian-map ;
+: eye ( m n k z -- matrix )
+    [ [ <iota> ] bi@ ] 2dip
+    '[ _ neg + = _ 0 ? ]
+    cartesian-map ;
+
+! if m = n and k = 0 then identity-matrix is (possibly) faster
+: simple-eye ( m n k -- matrix )
```
mrjbq7: This might be cleaner with locals:

```factor
:: simple-eye ( m n k -- matrix )
    m n = k 0 = and [ n identity-matrix ] [ 1 eye ] if ;
```

basis/math/matrices/matrices.factor
```diff
@@ -197,11 +209,11 @@ ERROR: negative-power-matrix m n ;
 : cols ( seq matrix -- cols )
     '[ _ col ] map ; inline
 
-: set-index ( object pair matrix -- )
+: matrix-set-nth ( n pair matrix -- )
```
mrjbq7: I can see calling the element being put in the matrix ``n``, but maybe ``elt`` for consistency?  I dunno.

mrjbq7: Since technically, it doesn't have to be a number maybe.

mrjbq7:
Nice docs!

basis/bootstrap/image/primitives/primitives-docs.factor
catb0t: `bootstrap.primitives` is not a vocabulary that can be linked to though:
```factor
Factor 0.98 x86.64 (1852, heads/master-e0a93a6ea8, Thu Dec 21 07:25:16 2017)
[Clang (GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final))] on linux
IN: scratchpad USE: bootstrap.primitives
1: USE: bootstrap.primitives
                            ^
Vocabulary does not exist
name "bootstrap.primitives"
```
`help-lint` gives something similar which is why i did that.

`core/bootstrap/primitives.factor` is a file that exists but it must be in error, shouldn't it be called `core/bootstrap/primitives/primitives.factor`?

or maybe the docs meant `bootstrap.image.primitives` (which is a vocab)?

mrjbq7: So, ``core/bootstrap/primitives.factor`` is a script used in the bootstrap process, not meant to be a vocab you can import, which leads to the error using ``$vocab-link``.  I think ``bootstrap.image.primitives`` is only used by ``core/bootstrap/primitives.factor`` in the entire project.

catb0t: well, can we put the script `core/bootstrap/primitives.factor` in a `primitives` folder so that `help-lint` passes? or should i just ignore the linter?

mrjbq7: The quickest way is to make it a ``$snippet`` instead of a ``$vocab-link``.

Maybe longer-term we could make ``bootstrap.primitives`` it's own vocab and move the top-level script code into a ``MAIN:`` word.

catb0t: that makes sense, i'll put this in the next set of commits i make on this pull

mrjbq7: It's also funny that the "script" for ``bootstrap/primitives.factor`` actually does define itself as a vocab when you run it and puts some words in the vocab that it uses in the script part.  Anyway, good cleanup suggestion for later! Thanks!

mrjbq7:
Let me know when you're ready for another review or to merge the PR.

catb0t:
When I'm done writing the code and I start cleaning things up, I'll get rid of some of the merge commits from my other dev branches (or you can just squash when you merge).

mrjbq7:
I appreciate the cleanup!

Looks like you have some help.lint.pedantic stuff in this PR?

catb0t:
Yep, tracking those files on this branch was accidental (I needed the merges so that I can use `vocab-pedant` on `math.matrices`) but I'm going to leave them in until the final cleanup commits, so I can keep using the docs coverage tool.

catb0t:
Eh, this is more messy than it's worth now. I'm going to re-open this pull request with fewer (hopefully 1 or 2) commits for the same changes soon.

PR #1912, opened 2018-01-22
Boolean expr

AlexIljin:
Resurrect `boolean-expr` from the unmaintained, add documentation.

### DIFF
--- a/extra/boolean-expr/authors.txt
+++ b/extra/boolean-expr/authors.txt
@@ -1,0 +1,1 @@
+Slava Pestov
--- a/extra/boolean-expr/boolean-expr-docs.factor
+++ b/extra/boolean-expr/boolean-expr-docs.factor
@@ -9,23 +9,21 @@ ARTICLE: "boolean-expr" "Boolean expressions"
 "The " { $vocab-link "boolean-expr" } " vocab demonstrates the use of Unicode symbols in source files and multi-method dispatch."
 ;
 
-HELP: cnf
+HELP: dnf
 { $values
     { "expr" □ }
-    { "cnf" array }
+    { "dnf" array }
 }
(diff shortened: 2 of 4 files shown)


mrjbq7:
That's cool.  Saw your comment on cnf, do you think it's busted or..?

AlexIljin:
It's either incorrectly implemented in the first place (inverted logic or something like that) or I don't understand its output.

AlexIljin:
Judging by this code:
```
METHOD: satisfiable? { □ }
    cnf [ (satisfiable?) ] any? ;
```
I think `cnf` should be renamed `dnf`. Otherwise instead of `any?` there would have been `all?`.
What do you think?

AlexIljin:
Yeah, I'm pretty confident now. Both its output and the `satisfiable?` implementation make sense only if `cnf` produces DNF. I have renamed it accordingly and updated the documentation.

mrjbq7:
Cool. Want to have a go at making a working CNF?

AlexIljin:
No, thank you! : ))

AlexIljin:
I worked a little bit on reviving `bubble-chamber`, but got stuck at the multi-method dispatch syntax.

erg:
They're in ``unmaintained/`` for a reason ;)

AlexIljin:
Multi-methods are in extra, and there was a custom (better?) syntax vocab on top of that in unmaintained (`multi-method-syntax`), but I have no experience working with the parser.

mrjbq7:
Merged, thanks!

mrjbq7:
Since you mentioned it, I resurrected ``bubble-chamber`` and ``frame-buffer`` which it depended on from unmaintained.

It draws stuff but seems to have an issue on retina displays.

PR #1913, opened 2018-01-22
Assorted updates

AlexIljin:
I went through my changes made in 2017, and here are the ones worth sharing.


mrjbq7:
Merged, thanks!!

PR #1914, opened 2018-01-22
Sodium FFI

AlexIljin:
[Sodium](https://github.com/jedisct1/libsodium) is a well-regarded modern crypto library. I have started creating bindings for it, and those I added were tested. Unfortunately, the work is incomplete.

Are you interested?

### DIFF
--- a/extra/checksums/sodium/authors.txt
+++ b/extra/checksums/sodium/authors.txt
@@ -1,0 +1,1 @@
+Alexander Ilin
--- a/extra/checksums/sodium/sodium-tests.factor
+++ b/extra/checksums/sodium/sodium-tests.factor
@@ -1,0 +1,21 @@
+! Copyright (C) 2018 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: tools.test checksums.sodium ;
+IN: checksums.sodium.tests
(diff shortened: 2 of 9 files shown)


mrjbq7:
Sure, what work did you mean was unfinished?

Does the stuff implemented work?  And there's just more of the library to expose?

Can you add some test cases in ``checksums.sodium`` just to make sure kicking the tires works?

AlexIljin:
@mrjbq7: the implemented stuff does work, but there are many more functions to be added before full library is accessible. I was adding the functions in logical groups (memory management, hashing, etc.), so each group should be complete, but that was a couple of minor releases ago, and some groups may have grown a little since.

I can add some tests, but wouldn't that add the libsodium dependence to the build machine with its automated testing?

mrjbq7:
You can add a ``tags.txt`` with "not tested" as one of the tags.

http://docs.factorcode.org:8080/content/word-don't-test__que__,vocabs.metadata.html

Then it won't be automatically tested, but can still be tested with ``"checksums.sodium" test``.

AlexIljin:
OK, I'll do that. I wasn't sure of the protocol for adding tests like that, thanks for the tip.

PR #1917, opened 2018-01-24
Fix the IN: for test vocabs to include the ".tests" suffix

AlexIljin:
Change some existing suffixes from "-tests" or ".test" to ".tests".

Inspired by @bjourne's commit 2424a77507408d382eddd3ca6395b4b0664ac98c.

Most of the work was done by the following script, but manual adjustments were also necessary.
```
! Copyright (C) 2018 Alexander Ilin.
USING: io.directories.search io.encodings.utf8 io.files kernel
sequences ;
IN: tests-fix-in

: ensure-tail ( str tail -- str' )
    2dup tail? [ drop ] [ append ] if ;

: fix-tests-IN: ( lines -- lines' )
    dup [ "IN: " head? ] find [ ! lines i str
        over [ ".tests" ensure-tail ] change-nth
    ] [ drop ] if ;

: fix-tests ( -- )
    "." "-tests.factor" find-files-by-extension
    [ utf8 [ fix-tests-IN: ] change-file-lines ] each ;

MAIN: fix-tests
```

### DIFF
--- a/basis/concurrency/count-downs/count-downs-tests.factor
+++ b/basis/concurrency/count-downs/count-downs-tests.factor
@@ -1,5 +1,5 @@
 USING: concurrency.count-downs threads kernel tools.test ;
-IN: concurrency.count-downs.tests`
+IN: concurrency.count-downs.tests
 
 { } [ 0 <count-down> await ] unit-test
 
--- a/basis/interval-maps/interval-maps-tests.factor
+++ b/basis/interval-maps/interval-maps-tests.factor
@@ -1,5 +1,5 @@
 USING: kernel namespaces interval-maps tools.test ;
-IN: interval-maps.test
+IN: interval-maps.tests
 
 SYMBOL: test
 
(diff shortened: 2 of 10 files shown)


bjourne:
In my commit, the name of the vocab was also misspelled. :P We have discussed what to do with vocabs unit tests and whether we should put them in `.tests` directories or not. I like that convention because it is consistent. But it is only required if you define words in the unit test files. 

I think your pr will break many tests though. E.g, if you change `IN: calendar` to `IN: calendar.tests`, you must also add a corresponding `USING: calendar`.

AlexIljin:
@bjourne: whoa, let's not do this then. At least I learned something new! : ))

AlexIljin:
I'll update the PR to only update the non-standard suffixes like "-tests" or ".test" for consistency.

mrjbq7:
As an aside, it would be good to not have IN: declarations if no words are defined as part of the tests. And I’ve always liked the idea of not requiring IN: when it’s “obvious” (foo/foo.factor) or the system can know what it should be (-docs, -tests, etc). 

Long term we should definitely clean this up!

AlexIljin:
I've updated the PR, now it should introduce no issues that weren't there in the first place.

mrjbq7:
Merged,  thanks!

PR #1926, opened 2018-02-01
fix syntax typo in shaders docs

catb0t:
tiny little `load-all` blocker :)


bjourne:
Ops, that was embarrasing! Thanks for spotting it.

PR #1927, opened 2018-02-02
v0.3 help.lint.coverage

catb0t:
Especially in the case of large vocabularies like `math.matrices`, it is difficult for me to keep a memory of all the documentation I have written while editing another part of the enormous documentation file.

Also, I always say, "ah, I will write the documentation later" and then forget about it, and never write it.

Therefore, I need (as we all do) a pedant looking over my shoulder to nag me about documentation I forgot to write >_>

Missing features:

1. fixes for some of the simple ways to trick the pedant, like a way to check non-empty `$examples` doesn't simply say `{ $examples { $example "" } }` or `{ $examples { } }`
1. a way to ignore certain words

A more mature version of this vocabulary will have all these features, but for now I think this is useful in the standard library on its own, to quickly nag you about missing documentation sections -- when you would like to include such sections.

Also, if you run `vocab-pedant` on the Factor distribution's core vocabs you will get a bunch of errors because words like `map` and `dup` are missing sections like `$examples`. I am not requesting you pull this code so that we are all incentivised to add missing sections to every word in the Factor distribution, because that would be ridiculous. 

We all learned to write Factor just fine without an `$examples` section in `map`'s help :) 

This is intended only for complex end-user and domain-specific code which you want to ensure has a very complete documentation.


catb0t:
I forgot to mention, the license is CC0 public domain so it can be BSD when it's merged https://github.com/catb0t/pedant-linter

catb0t:
Incidentally I gave `help.lint.checker` the `tools` tag it was missing, too

bjourne:
Great work! But I think you should put it in the `extra/` hierarchy.

catb0t:
Moved to `resource:extra/help/lint/pedantic` now, it makes more sense :)

catb0t:
The commits are squashed and stand-alone now, it's ready for merge whenever you are

mrjbq7:
Merged, but some suggestions:

1) make words like this so easier to use:

```factor
: vocab-help-coverage. ( vocab-spec -- )
    <vocab-help-coverage> print-coverage;
```

or maybe some variation of:

```factor
GENERIC: help-coverage. ( object -- )

M: word help-coverage. <word-help-coverage> print-coverage ;

M: object help-coverage <vocab-help-coverage> print-coverage ;
```

2) Print the word out so you can right-click and edit, something like:

```factor
"[" write "math" dup lookup-vocab write-object "] " write
\ zero? [ name>> ] keep write-object
": should define help section $examples" print
```

3) The "should define help sections" is a little verbose, maybe just "needs $contract, $example".

catb0t:
Those are in the TODO now, thank you! shall we leave this open until I push those changes for you to remerge?

mrjbq7:
Or even,

```factor
GENERIC: help-coverage ( object -- coverage )

M: word help-coverage <word-help-coverage> ;

M: object help-coverage <vocab-help-coverage> ;

: help-coverage. ( object -- ) help-coverage print-coverage
```

Something convenient, doesn't matter too much to me.

You can keep this PR open or use the same branch or make another PR, I can cherry-pick pretty easily.

mrjbq7:
Seeing this error on the build bot:

```
Help lint failures
resource:extra/help/lint/coverage/coverage-docs.factor
Details:
==== resource:extra/help/lint/coverage/coverage-docs.factor

resource:extra/help/lint/coverage/coverage-docs.factor: 39

Asset: T{ link f print-coverage }

Assertion failed
got    "[help.lint.coverage] <word-help-coverage>: full help coverage"
expect "Word '<word-help-coverage>' has 100% help coverage"
```

Reproduce with:

```factor
IN: scratchpad "help.lint.coverage" help-lint
```

catb0t:
I've implemented your suggestions, and I rewrote history so it's (hopefully) easier to merge on the real git repository, not this one.

Unfortunately, there are now some obscure and inexplicable heisenbugs when writing `word-help-coverage.` causing the output to be too long for the UI listener to display, causing a word error. These will take more time to debug, so I'll do that when I polish this new version for another merge.

mrjbq7:
Looks like rewriting history is going to make it harder to merge since I already merged your 0.1, 0.2, and 0.3 commitsA:

extra/help/lint/coverage/coverage.factor
```diff
@@ -0,0 +1,146 @@
...
+PRIVATE>
+
+GENERIC: <word-help-coverage> ( word -- coverage )
+M: word <word-help-coverage>
+    dup [ missing-sections ] [ empty-examples? ] bi
+    2dup 2array { { } f } =
+    word-help-coverage boa ; inline
+
+M: string <word-help-coverage>
+    loaded-vocab-names resolve-name-in call-next-method ; inline
```
mrjbq7: Not a big fan of this approach to looking up words from strings, I think maybe better to either not support string resolution of names, or do something like:

```factor
: find-word ( name -- word/f )
    dup dictionary get values [ lookup-word ] with map sift
    dup length {
        { 0 [ 2drop f ] }
        { 1 [ first nip ] }
        [ drop <ambiguous-use-error> throw-restarts ]
    } case ;
```

So if you try, for example, ``"search" find-word`` you get a restart with the three possibilities in binary-search, peg.search, and, vocabs.parser.

extra/help/lint/coverage/coverage.factor
```diff
@@ -0,0 +1,146 @@
...
+
+: <vocab-help-coverage> ( vocab-spec -- coverage )
+    [ auto-use? off vocab-words natural-sort [ <word-help-coverage> ] map ] with-scope ;
```
mrjbq7: If you use my ``find-word`` above, then you can get rid of these ``with-scope`` stuff because there is no eval taking place.

mrjbq7:
I haven't tried the code, but not a big fan of heisenbugs which seems to me might be a result of the ``eval`` taking place.  Maybe you can try making this a commit on top of the three I already merged, or just fix it to where you're happy and I can do a manual merge, no worries.

extra/help/lint/coverage/coverage.factor
bjourne: There is also a `find-word`  already existing in `fuel.help`

extra/help/lint/coverage/coverage.factor
catb0t: I take your point, but the `with-scope` here is because of `auto-use? off` and `group-articles vocab-articles set`, not because of `eval`. The `eval` would have the same behaviour regardless of the new namespace scope, because I am mimicking the definition of `help-lint` which looks like 

```factor
: help-lint ( prefix -- )
    [
        auto-use? off group-articles vocab-articles set
        loaded-child-vocab-names [ check-vocab ] each
    ] with-scope ;
```

extra/help/lint/coverage/coverage.factor
catb0t: Hmm... the use of `loaded-vocab-names resolve-name-in` was that the string would be looked up in the exact same context where it was written, and `fuel.help:find-word` seems good for that purpose, so I'll use that

catb0t:
fixes are in #1933

extra/help/lint/coverage/coverage.factor
mrjbq7: Looks like when the test runner runs ``help.lint.coverage`` tests, the ``"eye" find-word`` fails because ``search`` (called by ``fuel.help:find-word``) throws an error that ``manifest`` is ``f``...

```
==== resource:extra/help/lint/coverage/coverage-tests.factor

resource:extra/help/lint/coverage/coverage-tests.factor: 37

Unit Test: { { eye } [ "eye" find-word ] }

Generic word qualified-vocabs>> does not define a method for the POSTPONE: f class.
Dispatching on object: f
```

Since the listener has ``manifest`` set to the ``scratchpad`` vocabulary (causing the tests to pass when run manually), you can reproduce this way:

```factor
IN: scratchpad f manifest [ "help.lint.coverage" test ] with-variable
```

For the moment, I put a ``find-word`` that is more like the one I suggested, so the tests pass, but feel free to either just call ``words-named ?first`` or change it in some other way.  I only patched it to fix the build not to overwrite your ideas.

PR #1933, opened 2018-02-05
refactor help.lint.coverage with complete docs and tests

catb0t:
follow-on to #1927 
this will un-break the build and provide a more clear API with better tests too.

### DIFF
--- a/extra/help/lint/coverage/coverage.factor
+++ b/extra/help/lint/coverage/coverage.factor
@@ -1,8 +1,9 @@
 USING: accessors arrays classes classes.error combinators
 combinators.short-circuit continuations english eval formatting
-fry generic help help.lint help.lint.checks help.markup io
-kernel math namespaces parser prettyprint sequences sets sorting
-splitting strings summary vocabs words ;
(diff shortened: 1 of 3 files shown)


extra/help/lint/coverage/coverage.factor
```diff
@@ -78,30 +109,14 @@ M: word-help-coverage summary
...
-    dup
-    [ missing-sections ]
-    [ empty-examples? ] bi
+    dup [ missing-sections ] [ empty-examples? ] bi
     2dup 2array { { } f } =
     word-help-coverage boa ; inline
 
 M: string <word-help-coverage>
-    loaded-vocab-names resolve-name-in <word-help-coverage> ; inline
+    find-word <word-help-coverage> ; inline
```
mrjbq7: Don't you think this doesn't work well for ``"search" find-word``?

My version did a restart when there were several potential options rather than returning only the first.  Also, we probably want it to work for qualified strings like ``"sequences:nth" find-word``.

mrjbq7:
Merged this PR.

Minor comment about use of ``find-word``.  Kind of liked the features of the one I suggested, only because of not just resolve "the first".

PR #1934, opened 2018-02-05
add $related-subsections to minimise docs duplication

catb0t:
very often there are some words in a `{ $subsectoions a b c d }` and then later on in the docs `{ a b c d } related-words`, but it is duplicate-y and you have to change the list in 2 places every time you add a word. :( 

I have found this incredibly useful for documenting vocabs with lots of similar words like math.matrices' new documentation (https://github.com/catb0t/factor/commit/662e6f855ea6a5f5b86632e70a6e0472b80ffd1f); I've used this in a couple other places and I think it would fit in the distribution

### DIFF
--- a/basis/help/help-docs.factor
+++ b/basis/help/help-docs.factor
@@ -195,7 +195,14 @@ HELP: $subsections
     { $markup-example { $subsections "sequences" nth each } }
 } ;
 
-{ $subsection $subsections $link } related-words
+HELP: $related-subsections 
+{ $values { "children" "a " { $link sequence } " of one or more " { $link topic } "s" } }
+{ $description "A version of " { $link $subsections } " which also calls " { $link related-words } " on " { $snippet "children" } ", relating its contents together." }
+{ $examples
+    { $markup-example { $related-subsections keep + } }
+} ; 
+
+{ $subsection $subsections $related-subsections $link $vocab-subsection } related-words
 
 HELP: $vocab-subsection
 { $values { "element" "a markup element of the form " { $snippet "{ title vocab }" } } }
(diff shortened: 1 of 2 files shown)


mrjbq7:
Merged, could be a neat feature.  Let a thousand flowers bloom!

bjourne:
Does this pr really work? If you have a declaration `{ ... } related-words` then that is evaluated when the doc file is parsed. But the `$related-subsections` is only evaluated when the help for the word containing the markup is rendered.

mrjbq7:
Hmm, yes well @bjourne I should have noticed that also, dang!

{ $related-subsections a b c } is just an array with the first token the word object, but it's not executed until ``print-element`` or whatever is called.

So, side-effect markup isn't probably a good idea.  Reverting for now, maybe another approach could work.

catb0t:
Whoops, I should have considered when it would actually be called for its side effects. 

I initially experimented with 
```factor
{ $subsections a b c d } [ 1 tail related-words ] keep
``` 
and the like, but this doesn't work inside `HELP:` or `ARTICLE:` for obvious reasons (there's no stack to manipulate)

I also tried to write hacky parsing words like (rough recollection) 
```factor
SYNTAX: $related-subsections{ 
  \ } [ >array ] parse-literal 
  [ last related-words ] 
  [ \ $subsections prefix ] bi ; 
``` 
so you can write `$related-subsections{ a b c d }` (ugly) or 
```factor
SYNTAX: $related-subsections 
  dup pop* ! discard {
  \ } [ >array ] parse-literal 
  [ last related-words ] 
  [ \ $subsections prefix ] bi ;
```
so you can write `{ $related-subsections a b c d }` but it didn't really work, for various reasons and things i'm still investigating.

PR #1944, opened 2018-02-14
core/combinators docs: fix unclearness in case help

catb0t:
initially i wanted to correct that `case` throws `no-case` not `no-cond` but i ended up cleaning up some other parts of the word's help too.

### DIFF
--- a/core/combinators/combinators-docs.factor
+++ b/core/combinators/combinators-docs.factor
@@ -343,16 +343,17 @@ HELP: no-cond
 { $error-description "Thrown by " { $link cond } " if none of the test quotations yield a true value. Some uses of " { $link cond } " include a default case where the test quotation is " { $snippet "[ t ]" } "; such a " { $link cond } " form will never throw this error." } ;
 
 HELP: case
-{ $values { "obj" object } { "assoc" "a sequence of object/word, quotation pairs, with an optional quotation at the end" } }
+{ $values { "obj" object } { "assoc" "an assoc of object/quotation pairs, with an optional quotation at the end" } }
 { $description
-    "Compares " { $snippet "obj" } " against the first element of every pair, first evaluating the first element if it is a word. If some pair matches, removes " { $snippet "obj" } " from the stack and calls the second element of that pair, which must be a quotation."
+    "Compares " { $snippet "obj" } " against the first element of every " { $link pair } ", evaluating the first element if it is a " { $link callable } ". If a pair matches, " { $snippet "obj" } " is removed from the stack and the second element of that pair (which must be a " { $link quotation } ") is " { $link call } "ed."
     $nl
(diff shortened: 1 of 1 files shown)


core/combinators/combinators-docs.factor
```diff
@@ -343,16 +343,17 @@ HELP: no-cond
 { $error-description "Thrown by " { $link cond } " if none of the test quotations yield a true value. Some uses of " { $link cond } " include a default case where the test quotation is " { $snippet "[ t ]" } "; such a " { $link cond } " form will never throw this error." } ;
 
 HELP: case
-{ $values { "obj" object } { "assoc" "a sequence of object/word, quotation pairs, with an optional quotation at the end" } }
+{ $values { "obj" object } { "assoc" "an assoc of object/quotation pairs, with an optional quotation at the end" } }
```
mrjbq7: It's a sequence, not an assoc, because the "optional quotation" isn't a key/value pair at the end:

```factor
15 {
    { 1 [ "one" ] }
    { 2 [ "two" ] }
    [ number>string ]
} case
```

core/combinators/combinators-docs.factor
```diff
@@ -343,16 +343,17 @@ HELP: no-cond
...
+    "Compares " { $snippet "obj" } " against the first element of every " { $link pair } ", evaluating the first element if it is a " { $link callable } ". If a pair matches, " { $snippet "obj" } " is removed from the stack and the second element of that pair (which must be a " { $link quotation } ") is " { $link call } "ed."
     $nl
-    "If there is no case matching " { $snippet "obj" } ", the default case is taken. If the last element of " { $snippet "assoc" } " is a quotation, the quotation is called with " { $snippet "obj" } " on the stack. Otherwise, a " { $link no-cond } " error is raised."
+    "If there is no case matching " { $snippet "obj" } ", the default case is taken. If the last element of " { $snippet "assoc" } " is a quotation, the quotation is called with " { $snippet "obj" } " on the stack. Otherwise, a " { $link no-case } " error is raised."
     $nl
-    "The following two phrases are equivalent:"
+    "In general, the following two phrases are equivalent:"
```
mrjbq7: Is it not exactly equivalent? Just generally so?

catb0t: Hmm,  you're right. I write "in general" because it lines up with the other combinator examples documentation that has a specific case and then a general case, like bi's documentation

> If [ p ] and [ q ] have stack effect ( x -- ), then the following two lines are equivalent:
> [ p ] [ q ] bi
> dup p q
> If [ p ] and [ q ] have stack effect ( x -- y ), then the following two lines are equivalent:
> [ p ] [ q ] bi
> dup p swap q
> In general, the following two lines are equivalent:
> [ p ] [ q ] bi
> [ p ] keep q

I'll fix this tomorrow morning.

catb0t:
Those changes were reverted.

mrjbq7:
Merged, thanks!!

I squashed into a single commit 08b55b6a625b9b2b3b93a2950150080d7450a1d3.

PR #1945, opened 2018-02-14
db.tuples improvement

AlexIljin:
Hey, guys! I'm very excited about these changes. Was pretty surprised that this functionality wasn't there in the first place, and I'm glad I was able to contribute it with only half a day's worth of effort.

Basically, in this PR I add two new words to `db.tuples`: `each-tuple` and `map-tuples`. These words allow one to stream and handle really big amounts of data one tuple at a time without creating the intermediate array in memory. The old `query-tuples` word is now implemented using the new `query-tuples-each`, so there is no code duplication.

A lot of words had to be declared `inline` because of the quotations now being passed through the stack. I hope that's not a big issue.

The new words have been tested by me with an SQLite DB, but I didn't have time yet to add new unit-tests. I want you to take a look and make some comments before I finalize this PR by adding the tests. If you think documentation needs improvement or some words may have better names, please suggest.

### DIFF
--- a/basis/db/tuples/tuples-docs.factor
+++ b/basis/db/tuples/tuples-docs.factor
@@ -122,13 +122,23 @@ HELP: update-tuple
      { "tuple" tuple } }
 { $description "Updates a tuple that has already been inserted into a database. The tuple must have a primary key that has been set by " { $link insert-tuple } " or that is user-defined." } ;
 
+HELP: update-tuples
+{ $values
+     { "query/tuple" tuple }
+     { "quot" { $quotation ( tuple -- tuple'/f ) } } }
+{ $description "An SQL query is constructed from the slots of the exemplar tuple that are not " { $link f } ". The " { $snippet "quot" } " is applied to each tuple from the database that matches the query, and the changed tuple is stored back to the database. If the " { $snippet "quot" } " returns " { $link f } ", the tuple is dropped, and its data remains unmodified in the database."
+$nl
+"The word is equivalent to the following code:"
+{ $code "query/tuple select-tuples quot map sift [ update-tuple ] each" }
+"The difference is that " { $snippet "update-tuples" } " handles query results one by one, thus avoiding the overhead of allocating the intermediate array of tuples, which " { $link select-tuples } " would do. This is important when processing large amounts of data in limited memory." } ;
+
 HELP: delete-tuples
 { $values
      { "tuple" tuple } }
 { $description "Uses the " { $snippet "tuple" } " as an exemplar object and deletes any objects that have the same slots set. If a slot is not " { $link f } ", then it is used to generate an SQL statement that deletes tuples." }
 { $warning "This word will delete your data." } ;
 
-{ insert-tuple update-tuple delete-tuples } related-words
+{ insert-tuple update-tuple update-tuples delete-tuples } related-words
 
 HELP: each-tuple
 { $values
@@ -183,8 +193,11 @@ ARTICLE: "db-tuples-words" "High-level tuple/database words"
 { $subsections drop-table }
 "Inserting a tuple:"
 { $subsections insert-tuple }
-"Updating a tuple:"
-{ $subsections update-tuple }
+"Updating tuples:"
+{ $subsections
+    update-tuple
+    update-tuples
+}
 "Deleting tuples:"
 { $subsections delete-tuples }
 "Querying tuples:"
--- a/basis/db/tuples/tuples-docs.factor
+++ b/basis/db/tuples/tuples-docs.factor
@@ -130,6 +130,12 @@ HELP: delete-tuples
 
 { insert-tuple update-tuple delete-tuples } related-words
 
+HELP: each-tuple
+{ $values
+     { "query/tuple" tuple }
+     { "quot" { $quotation ( tuple -- ) } } }
+{ $description "An SQL query is constructed from the slots of the exemplar tuple that are not " { $link f } ". The " { $snippet "quot" } " is applied to each tuple from the database that matches the query constructed from the exemplar tuple." } ;
+
 HELP: select-tuple
 { $values
      { "query/tuple" tuple }
@@ -148,7 +154,7 @@ HELP: count-tuples
      { "n" integer } }
 { $description "Returns the number of items that would be returned if the query were a select query. Counting the tuples with this word is more efficient than calling " { $link length } " on the result of " { $link select-tuples } "." } ;
 
-{ select-tuple select-tuples count-tuples } related-words
+{ each-tuple select-tuple select-tuples count-tuples } related-words
 
 
 
@@ -183,6 +189,7 @@ ARTICLE: "db-tuples-words" "High-level tuple/database words"
 { $subsections delete-tuples }
 "Querying tuples:"
 { $subsections
+    each-tuple
     select-tuple
     select-tuples
     count-tuples
--- a/basis/db/tuples/tuples-docs.factor
+++ b/basis/db/tuples/tuples-docs.factor
@@ -134,13 +134,13 @@ HELP: select-tuple
 { $values
      { "query/tuple" tuple }
      { "tuple/f" { $maybe tuple } } }
-{ $description "A SQL query is constructed from the slots of the exemplar tuple that are not " { $link f } ". Returns a single tuple from the database if it matches the query constructed from the exemplar tuple." } ;
+{ $description "An SQL query is constructed from the slots of the exemplar tuple that are not " { $link f } ". Returns a single tuple from the database if it matches the query constructed from the exemplar tuple." } ;
 
 HELP: select-tuples
 { $values
      { "query/tuple" tuple }
      { "tuples" "an array of tuples" } }
-{ $description "A SQL query is constructed from the slots of the exemplar tuple that are not " { $link f } ". Returns an array of multiple tuples from the database that match the query constructed from the exemplar tuple." } ;
+{ $description "An SQL query is constructed from the slots of the exemplar tuple that are not " { $link f } ". Returns an array of multiple tuples from the database that match the query constructed from the exemplar tuple." } ;
 
 HELP: count-tuples
 { $values
--- a/basis/db/tuples/tuples-docs.factor
+++ b/basis/db/tuples/tuples-docs.factor
@@ -1,4 +1,5 @@
 ! Copyright (C) 2008 Doug Coleman.
+! Copyright (C) 2018 Alexander Ilin.
 ! See http://factorcode.org/license.txt for BSD license.
 USING: classes help.markup help.syntax io.streams.string kernel
 quotations sequences strings math db.types db.tuples.private db ;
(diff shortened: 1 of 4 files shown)


catb0t:
maybe they should be called `map-tuple` and `each-tuple`, for consistency of plurality -- or maybe it doesn't matter.

basis/db/tuples/tuples.factor
```diff
@@ -63,6 +67,10 @@ GENERIC: eval-generator ( singleton -- object )
     [ <insert-user-assigned-statement> ] cache
     [ bind-tuple ] keep execute-statement ;
 
+: do-each-tuple ( exemplar-tuple statement quot -- tuples )
```
mrjbq7: Might be nice to show ``quot`` stack effect, ``quot: ( tuple -- )``.

basis/db/tuples/tuples-docs.factor
```diff
@@ -122,33 +123,49 @@ HELP: update-tuple
...
 { $description "Uses the " { $snippet "tuple" } " as an exemplar object and deletes any objects that have the same slots set. If a slot is not " { $link f } ", then it is used to generate an SQL statement that deletes tuples." }
 { $warning "This word will delete your data." } ;
 
-{ insert-tuple update-tuple delete-tuples } related-words
+{ insert-tuple update-tuple map-tuples delete-tuples } related-words
+
+HELP: each-tuple
+{ $values
+     { "query/tuple" tuple }
+     { "quot" quotation } }
```
mrjbq7: You can put quotation stack effects in help.

``{ "quot" { $quotation ( tuple -- ) } }``

In a future version of Factor, we wouldn't need to redundantly describe what the system already knows if the actual stack effects are specified properly in the word definition.

basis/db/types/types-docs.factor
```diff
@@ -97,12 +97,12 @@ HELP: user-assigned-id-spec?
 
 HELP: bind#
 { $values
-     { "spec" "a SQL spec" } { "obj" object } }
+     { "spec" "an SQL spec" } { "obj" object } }
```
mrjbq7: If this is actually an ``sql-spec`` object, then we could say that instead of the text "an SQL spec".

But I haven't looked closely at it in awhile to remember if it is or if allows other things to be "SQL spec-like" to be passed in.

mrjbq7:
Nice additions!  

Gave a few comments.

Cleanups in ``db`` are welcome!

bjourne:
I think you should change the name of `map-tuples` to something else. `update-tuples` perhaps. Reason being that map functions "maps" from one sequence or set to another.

AlexIljin:
@bjourne: I kind of agree that `update-tuples` would be better name for `map-tuples`, and my initial name for it was `change-tuples`. But the `change-` can be mistaken for the accessor word convention, and `update-tuples` may be a bit too similar name to `update-tuple`. I mean it's too similar for the difference in implementations. One word takes an existing tuple and writes it to the database, the other takes a query tuple and updates all matching tuples in the database.

Or do you think it's all the more reason for the name `update-tuples`? To me `update-tuples` sounds like an `update-tuple` over a sequence, not a new DB query. What do you say?

basis/db/tuples/tuples-docs.factor
AlexIljin: Believe it or not, I did try to include the stack effect into the help text, but failed to find the correct syntax. Thank you for the tip!

basis/db/tuples/tuples.factor
AlexIljin: Will do, thanks!

AlexIljin:
Here's some bizarre compiler behavior.
If you change the stack effect declaration from
`: query-tuples-each ( exemplar-tuple statement quot: ( tuple -- ) -- )`
to
`: query-tuples-each ( exemplar-tuple statement quot: ( tuple -- ) -- seq )`
and reload the vocab, there will be no compilation error.

Can someone explain this to me? Is this because of too much `inline`ing?

mrjbq7:
The stack effect of inline words isn't really checked, the stack effects of the non-inline words that use them is checked.

mrjbq7:
However, "row variadic" stack effects are checked:

See, for example my commit efcd5ab98dad7dbd86c81befe869fadf3cef9cc9.

If I remove the ``..a`` and ``..b`` it compiles fine.  But if I specified ``...`` instead then it would fail because the input and output stack effects can be slightly different (thus, ``a`` to ``b``).

bjourne:
@AlexIljin In normal languages, you can often deduce from the context whether a function mutates a list or a returns a new one. F.e if the code reads `list2 = map_tuples(list);`, you can be reasonably sure that the word doesn't mutate `list` but instead creates a new list.  If it instead reads `map_tuples(list);` you would guess that it does mutate the list.

In Factor, it is not as easy to look at the context to determine what the code does and therefore naming becomes much more important. F.e if the code read `some-word map-tuples length`, you would guess that `map-tuples` returns a new sequence and its length is taken. But if it read `some-word update-tuples length`, you would guess it is mutating the list. 

If neither `update-`, nor `change-` works, perhaps you can try `modify-tuples`, `apply-to-tuples` or `mutate-tuples`?

catb0t:
looks like it was changed to `update-tuples` which is quite nice.

AlexIljin:
I have added the unit-tests. From my point of view this PR is ready.

bjourne:
Looks good! I assume that you have confirmed that no part of the process between the db and factor materializes the full result set? E.g with your pr it should be possible to make a query that updates billions of rows without ever running out of memory? Sorry for the stupid question. :)

AlexIljin:
@bjourne: Not a stupid question at all! I haven't actually verified the memory requirements so far. Everything I did was based on assumptions. I assume I know how SQLite would handle the query and return requested rows, and I assume that by removing the `collector` call from the Factor code path I have no more array allocation for the full result. But I like the idea to test this.

mrjbq7:
Merged, thanks!

jonenst:
Just to chime in quickly, I recently tried to implement a streaming http endpoint fetching data from one database, but where you could choose to use different databases (targeted mysql/mariadb and postgres), and it felt like the traditional databases had poor support for this use case. For example, mysql has the following warning in their docs
> Therefore, if using streaming results, process them as quickly as possible if you want to maintain concurrent access to the tables referenced by the statement producing the result set.

Or
> Note that the server usually expects clients to read off the result set relatively quickly. The net_write_timeout server variable controls this behavior (defaults to 60s)

This SO question https://stackoverflow.com/questions/42292341/correct-use-of-cursors-for-very-large-result-sets-in-postgres talks about postgres and the author concludes "I don't really know why PostgreSQL behaves that way."

So don't assume too much from the database..

ps: if you have successfully used streaming results sets from databases I'd like to hear your story :)

mrjbq7:
@AlexIljin 

do all the tests work for you?  I got these errors on the build bots after merging your commits:

on sqlite

```
==== resource:basis/db/tuples/tuples-tests.factor

resource:basis/db/tuples/tuples-tests.factor: 33

Unit Test: { [ ] [ sqlite-test-db [ test-mapping ] with-db ] }

sql-table-missing
location f
table    "select_me"
```

and on postgresql

```
resource:basis/db/tuples/tuples-tests.factor: 43

Unit Test: { [ ] [ postgresql-test-db [ test-mapping ] with-db ] }

sql-table-missing
location f
table    "select_me"
```

AlexIljin:
@mrjbq7: yes, the tests do work for me, and I think I know the reason. The test database is never deleted between the test runs, and for my tests I needed to create contents, verify, modify and verify again. This means that when I run `"db.tuples" test` a second time, the test DB already has the modified contents from the last step of the previous run.
To fight this I added a `drop table` to the beginning, which now errors for you, but not for me, because I have that table from the previous run, and you don't.

How do we solve this? Shouldn't the test db be created anew for every test run?

mrjbq7:
The quick and dirty way is to try dropping and ignore errors:

```
./basis/db/tuples/tuples-tests.factor:    [ serialize-me drop-table ] ignore-errors
./basis/db/tuples/tuples-tests.factor:    [ exam drop-table ] ignore-errors
./basis/db/tuples/tuples-tests.factor:    [ bignum-test drop-table ] ignore-errors
```

I made that change in a2ded18a0b95e80f4a28f2c93f5ba8a65bd930a6.  Thanks for the quick response!

bjourne:
You are also able to use `select-me recreate-table` instead of dropping and creating.

AlexIljin:
@bjourne: I've created an SQLite DB file of about 15.8Gb in size, then started a full `update-tuples` operation on it that should add a string value to a NULL field in every row. It's not finished yet, and I'm running out of disk space, but in the two hours that it's been working now I see a 7+Gb db-journal file, and the Factor.exe memory use sits dead still on 353Mb. I think it's pretty certain that there is no 15Gb memory buffer allocated anywhere (my laptop only has 16Gb, and Process Explorer says about 6.4Gb is free).

mrjbq7:
Got this test error now:

```
==== resource:basis/db/tuples/tuples-tests.factor

resource:basis/db/tuples/tuples-tests.factor: 688

Unit Test: {
    [ V{ "test" "test2" } ]
    [
        select-me new NULL >>data
        [ "test" >>data ] update-tuples select-me new
        [ data>> ] collector [ each-tuple ] dip
    ]
}

=== Expected:
V{ "test" "test2" }
=== Got:
V{ "test2" "test" }
```

Are the results not guaranteed ordering for that test?

AlexIljin:
@jonenst: the info on streaming data from DBs is very interesting. I would expect better things from them, given the man-years of development spent.

Luckily my use-case doesn't require such concurrency. The SQLite is simply used as data storage container for a single instance of an application working on it. BUT. I wanted the application to be responsive during the long operations (8-16 hours), I wanted them to report progress and I wanted the user to have the ability to pause/abort an operation. These were my reasons for looking at the streaming interfaces, and for the processing of data in background task threads.

It seems that the database documentation pages that you quoted can't promise consistent results of a DB select operation while some other actor updates the same tables that are being read/traversed. Reading about the SQLite [WAL](http://www.sqlite.org/wal.html), I've got the impression that readers, however many, can't be affected by the writer (although there can't be more than one writer). But hey, maybe SQLite is better than those other DBs ; ))

AlexIljin:
@mrjbq7: I don't explicitly control the order of the returned rows. Is it even possible to request an ordering with our `db.tuples` interface?
Anyway, I haven't seen this error with the SQLite backend, and I don't have PostgreSQL. Are you seeing this with PostgreSQL tests?

bjourne:
Ordering is never guaranteed for queries that does not have the ORDER BY clause. So the fix ought to be to convert the result into a set or sort it.

AlexIljin:
My test run is finished. The DB grew to 18+Gb, all rows were updated, and Factor.exe still uses about 353Mb of RAM.

@bjourne: I think sorting would be sufficient in this case. I didn't think about it when I wrote the test, simply assumed that if it worked for me, then it would also work for everyone else. : ))

Issue #1930, opened 2018-02-03
Move english out of basis, since it is never used in basis?


catb0t:
I am surprised that the error messages in `core` and `basis` do not use the english vocabulary `english`, but I wonder if maybe it should be in `extra` because of this?

Aside from my `resource:extra/help/lint/coverage` vocab, `resource:extra/math/text/french` is its only other use in the distribution.

catb0t:
I suspect that `basis` vocabularies which do pluralization, have their own pluralization implementations, which is something we should look to improve

catb0t:
closed because #1947 is open.

PR #1947, opened 2018-02-14
English improvements

catb0t:
I've updated extra/math.text.english, basis/english -- and also extra/help.lint.coverage, only because its docs rely on linting `basis/english`.

i've added new functionality to english, migrated some functionality from help.markup (but intentionally didn't edit help.markup), and wrote a lot of docs.

i'm `load-all`-ing right now but I am certain that two definitions for `a/an` will not cause ambiguity because as i mentioned in #1930, english is almost never used by other vocabs, and never in `core` or `basis`.

### DIFF
--- a/basis/english/english-docs.factor
+++ b/basis/english/english-docs.factor
@@ -1,0 +1,209 @@
+! Copyright (C) 2018 Cat Stevens
+USING: arrays assocs help.markup help.syntax kernel math multiline
+sequences strings ;
+IN: english
+
+ABOUT: "english"
+
+ARTICLE: "english" "English natural language transformations"
+
+<PRIVATE
+: $0-plurality ( children -- )
+    drop {
+        "Due to the unique way in which the English language is structured, the number "
+        { $snippet "0" }
+        " is considered plural; "
+        { $snippet "1" }
+        " is the only singular quantity."
+    } print-element ;
+
+: $keep-case ( children -- )
+    drop
+    "This word attempts to preserve the letter case style of the input." print-element ;
+PRIVATE>
+
+"The " { $vocab-link "english" } " vocabulary implements a few simple ways of interacting with text in the English language, for improving generated text."
+$nl
+"Plural and singular forms:"
+{ $subsections plural? pluralize ?pluralize count-of-things singular? singularize }
+
+"Toy grammatical words:"
+{ $subsections a/an ?plural-article a10n comma-list-by simple-comma-list }
+
+"An example application:"
+{ $subsections or-markup-example $or-markup-example } ;
+
+{ pluralize ?pluralize plural? count-of-things singularize singular? } related-words
+{ a/an ?plural-article a10n comma-list-by simple-comma-list $or-markup-example or-markup-example } related-words
+
+HELP: singularize
+{ $values { "word" string } { "singular" string } }
+{ $description "Determine the singular form of the input English word. If the input is already singular, it is returned unchanged." }
+{ $notes $keep-case  }
+{ $examples
+    { $example
+        "USING: english io ;"
+        "\"CATS\" singularize print"
+        "CAT"
+    }
+    { $example
+        "USING: english io ;"
+        "\"Octopi\" singularize print"
+        "Octopus"
+    }
+} ;
+
+HELP: singular?
+{ $values { "word" string } { "?" boolean } }
+{ $description "Attempt to determine whether the word is in singular form." }
+{ $examples
+    { $example
+        "USING: english prettyprint ;"
+        "\"octopi\" plural? ."
+        "t"
+    }
+} ;
+
+HELP: pluralize
+{ $values { "word" string } { "plural" string } }
+{ $description "Determine the plural form of the input English word. If the input is already plural, it " { $emphasis "might" } " be returned unchanged." }
+{ $notes { $list { "If the input is already in plural form, an invaid construct such as " { $emphasis "friendses" } " may be generated. This is difficult to avoid due to the unpredictable structure of the English language." } $keep-case } }
+{ $examples
+    { $example
+        "USING: english io ;"
+        "\"CAT\" pluralize print"
+        "CATS"
+    }
+    { $example
+        "USING: english io ;"
+        "\"Octopus\" pluralize print"
+        "Octopi"
+    }
+} ;
+
+HELP: plural?
+{ $values { "word" string } { "?" boolean } }
+{ $description "Attempt to determine whether the word is in plural form." }
+{ $examples
+    { $example
+        "USING: english prettyprint ;"
+        "\"octopus\" singular? ."
+        "t"
+    }
+} ;
+
+HELP: count-of-things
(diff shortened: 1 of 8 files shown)


basis/english/english.factor
```diff
@@ -141,11 +185,52 @@ PRIVATE>
         [ "s" append ]
     } cond match-case ;
 
+! using glue here means we don't mutate the input number, so it isn't coerced to int
```
mrjbq7: Good idea, but after your change, don't need the comment.

mrjbq7:
Cool.

Its probably a good idea to have ``help.markup`` use ``english`` so we don't duplicate ``a/an``.

catb0t:
Of course, but I figured I'd leave it to a different branch and different pull request.

basis/english/english.factor
```diff
@@ -105,6 +107,48 @@ CONSTANT: plural-to-singular $[ singular-to-plural assoc-invert ]
 
 PRIVATE>
 
+: singular? ( word -- ? )
```
mrjbq7: It's maybe slower, but maybe this is a better less duplicative definition:

```factor
: singular? ( word -- ? )
    [ singularize ] [ = ] bi ;
```

extra/math/text/english/english.factor
```diff
@@ -112,3 +112,17 @@ M: complex number>text
         [ 0 < " minus " " plus " ? ]
         [ abs number>text " i" append ] bi
     ] bi* 3append ;
+
+: ordinal-suffix ( n -- suffix )
```
mrjbq7: I had this floating around in my utils somewhere:

```factor
: humanize ( n -- str )
    dup 100 mod 11 13 between? [ "th" ] [
        dup 10 mod {
            { 1 [ "st" ] }
            { 2 [ "nd" ] }
            { 3 [ "rd" ] }
            [ drop "th" ]
        } case
    ] if [ number>string ] [ append ] bi* ;
```

It's weird to convert the number to a string then back to a number?

basis/english/english.factor
catb0t: true! I'll change it, but note that I can't use the equivalent `[ pluralize ] [ = ] bi` for `plural?` because `"birds" pluralize` and `"friends" pluralize` give you `birdses` and `friendses` -- not a bug,  just a limitation of trying to generalize English grammar.

extra/math/text/english/english.factor
catb0t: Hmm... your implementation outputs `th` for `-101` rather than `st`, to me it's "negative one hundred first"

catb0t:
Ok, I've implemented your suggestions

basis/english/english.factor
```diff
@@ -141,11 +143,80 @@ PRIVATE>
         [ "s" append ]
     } cond match-case ;
 
+: singular? ( word -- ? )
+    [ singularize ] [ = ] bi ;
+
+: plural? ( word -- ? )
```
mrjbq7: Or maybe (even slower but less duplicative):

```factor
: plural? ( word -- ? )
    [ singularize pluralize ] [ = ] bi ;
```

extra/math/text/english/english.factor
mrjbq7: Yeah, I didn't build it for negative numbers.  Was just pointing out using math only instead of strings.

basis/english/english.factor
```diff
@@ -141,11 +143,80 @@ PRIVATE>
...
+    >lower {
+        { [ dup empty? ] [ drop f ] }
+        { [ dup plural-to-singular key? ] [ drop t ] }
+        { [ dup dup plural-to-singular at = ] [ drop t ] } ! moose -> moose
```
mrjbq7: Does this case run?  Wouldn't the ``[ dup plural-to-singular key? ]`` before it return ``t``?

mrjbq7: Maybe you mean to use ``singular-to-plural`` there.

basis/english/english.factor
catb0t: Um, I get a `Retain stack overflow` when I do that, which is pretty impressive since I don't even remember how to make one of those intentionally.

mrjbq7: Hmm, that should work (and it does when I paste it in).

catb0t: I restarted Factor and it runs fine, bizarre.

basis/english/english.factor
catb0t: I think it was a guard for an earlier iteration because it kept failing to say that `moose` etc are both singular and plural, but you're right it's redundant now

mrjbq7:
P.S., I love all the contributions!  

Feel free to incorporate (_or not!_) my feedback if you feel strongly one way or another.  I was just trying to simplify the code when it's not performance critical.  Looks nicer I think, but please remember my comments aren't from a position of absolute authority!

basis/english/english-docs.factor
```diff
@@ -0,0 +1,209 @@
...
+        "a cat, a peach, or an object"
+    }
+} ;
+
+HELP: or-markup-example
+{ $values { "markup" "a sequence of markup elements" } { "classes" "a sequence of words" } }
+{ $description "Used to implement " { $link $or-markup-example } " and demonstrate " { $link comma-list-by } "." }
+{ $examples { "See the examples in " { $link $or-markup-example } "." } } ;
+
+HELP: $or-markup-example
```
mrjbq7: Help markup maybe should be in the ``help`` vocab hierarchy rather than ``english``?

catb0t: Yes, but these words `$or-markup-example` and `or-markup-example` are only to demonstrate the use of `comma-list-by`. 

When I write the pull request which makes `help.markup` use `english`'s `a/an` then I will look into using an implementation like this for `$or` / `($instance)` so that we don't have so much natural language code in `help.markup`.

catb0t:
This vocabulary is in `basis` yet it relies on `assocs.extras`, is this OK? I'm tempted to suggest move `assoc-invert` to `assocs`, or put `assocs.extras` in `assocs` or in `basis`.

mrjbq7:
I think it's okay for now to use ``assocs.extras``.  

But... probably we should move ``assoc-invert`` and ``assoc-invert-as`` to the ``assocs`` vocab.  This would be the second or third case of ``basis`` vocabs depending on ``extra`` vocabs.

catb0t:
I tried the `resolve conflicts` button and after I resolved the `USING:` conflict, it gave me a green button that says "Commit merge" and I'm not sure what will happen when I click it, so I'm gonna leave it up to you whenever you merge this. :+1:

catb0t:
I think it's pretty funny (not necessarily bad) that you can have `<PRIVATE ... PRIVATE>` inside `ARTICLE: ... ;` :D that's what I just changed in the last commit

mrjbq7:
Yep, and that green merge button doesn't really work since github is just a mirror of the factorcode.org Git repository, so we push to that and then mirror it here.

PR #1948, opened 2018-02-16
Math.matrices improvements (rewrite)

catb0t:
71.1% of the documentation is written, probably 50-60% of the tests are written, and probably 70% of the actual main source code is written, but I'll open this pull to request feedback anyway -- please don't merge yet :P  (I'll remember not to mess around merging upstream this time)

Many things are not in their final state, so take messy code with a grain of salt (but don't refrain from pointing it out).

> I have rather accidentally (almost!) completely rewritten math.matrices. There are many new features, which will make mathematical matrix operations in Factor a breeze. 

> * Matrix kinds are now predicate classes, so you can ask `matrix?` `square-matrix?` and so on. 
> * (...)

> I originally only wanted to add documentation for every word -- seriously, that was all I initially wanted to do. I wrote quite a lot of documentation before realising that a lot of the words could use re-naming and re-organising, so I made some minor re-organisations to the implementation. 

> Then I started splitting words up and adding functionality, and I decided to implement new functionality, and give new meaning to existing functionality. 

> Now we have predicate classes, more errors, a union class, and 120 total words. Suddenly, we have over 1,300 lines of tests and over 1,000 lines of docs for like 700 lines of actual code.

And yes the two UPPERCASE_FILES in the project root will be deleted in the final release commit. :+1:


catb0t:
I haven't done anything about it yet, but I was considering a virtual sequence implementation for constant things like `<identity-matrix>` and `<box-matrix>`, and for postponing calculation of transformation functions like `kronecker`. A box matrix is always `(2 * n)+1 ^ 2` copies of the number `1`, and it's pretty easy to predict the values in an `n`-dimensional `<identity-matrix>`, so maybe we can have an innovation there? 

`arrays.shaped` seems very cool but it will need better documentation before I can understand it enough to write compatibility ;)

catb0t:
What do you mean by that? I don't plan to merge master into my branch this time, if that's what you mean.

bjourne:
In my experience (but I'm not an expert!), longer branches often becomes cumbersome to review and merge. So I'm not discouraging your work at all, just saying that, for example, if you want to work on feature A, B and C it becomes a lot easier for the reviewers if the work is split into three separate prs.

catb0t:
Well, I understand your point. I think that my goals with this branch can be summed up like this, spread across a lot of commits:

- write complete and clean documentation 
- organise code / rename words like `m.` and `v.`, and constructors like `hankel-matrix` -> `<hankel-matrix>`, fixing all uses
- add missing features and unit tests for them (inverse, determinant, etc) 

And I think that these are all related to my singular goal which is to make `math.matrices` a nice thing to use and understand.

But! If you or someone else would rather I break up those three points into separate PRs rather than this large one, then I can do that too (although, that will take 3x as long because I will have to work on them individually rather than concurrently).

bjourne:
Don't do that. The pr is fine as it is. :)

bjourne:
Looks like you have added a bunch of word aliases. I don't think that is a great because different names for the same words makes the code harder to read. Aliases aren't that well supported in Factor either. For example, if you look up the help for a word alias, you get to the help for the alias itself and not to the help of the word the alias points to.

basis/math/matrices/matrices.factor
```diff
@@ -197,77 +297,316 @@ ERROR: negative-power-matrix m n ;
...
+! top right to bottom left; reverse the result if you expected it to start in the lower left
+: anti-diagonal ( matrix -- seq )
+    >square-matrix [ swap nth-end ] map-index ; inline
+
+ALIAS: transpose flip
+
+! VERY slow implementation
+: anti-transpose ( matrix -- newmatrix )
+    [ reverse ] map transpose [ reverse ] map ;
+
```
bjourne: If it is so slow is there a use case for it?

bjourne:
I also wonder if at least not some of the code you have added to the module would be better of in a `math.matrices.extras` vocab? For example, the words for calculating determinants and randomizing matrices. We can always move them into the main `math.matrices` vocab if they seem useful enough.

basis/math/matrices/matrices.factor
```diff
@@ -197,77 +297,316 @@ ERROR: negative-power-matrix m n ;
...
+    dup [ rank ] [ dim ] bi [ = ] with map {
+        { { t f } [ (left-inverse) ] }
+        { { f t } [ (right-inverse) ] }
+        [ no-case ]
+    } case ;
+PRIVATE>
+
+! A^-1
+GENERIC: multiplicative-inverse ( matrix -- inverse )
+M: zero-square-matrix multiplicative-inverse
```
bjourne: Perhaps you can add methods to the recip generic instead of defining a new one?

bjourne:
Hm.. you have a point regarding `flip`, I'll give you that. `transpose` is a very common operation in linear algebra. I still think the other aliases `<cartesian-indices>`, `row-map`, `additive-inverse` and `cartesian-row-map`  are superfluous.

I also noticed the `math.matrices.laplace` vocab in extra/. It calculates determinants using Laplace expansion. It has happened me a few times too that I wrote something and then found that it already exists in the Factor library. 

Yes, I'd be in favor of moving the esoteric words you mention from `math.matrices` to a `math.matrices.extras` vocab. The smaller the vocab, the faster it is to compile.

basis/math/matrices/matrices.factor
catb0t: of course, it makes sense to do that -- now in `extra` :)

catb0t:
I resolved the merge conflicts. The change `[-]` in master to `ui.gadgets.sliders` is merged with the new `vdot`, but I removed the new `population-cov-ddof` from `math.matrices` because it belongs with the other `cov`ariance functions in `math.matrices.extras`. This word and unit test will be present in the next commit to my branch.

mrjbq7:
How's this PR going?

timor:
I like renaming `v.` and `m.`, but I would suggest to do that in a separate PR, and also use the `deprecated` declaration on the old words,

mrjbq7:
Hi @catb0t , how's this PR coming?  Want some help?  Ready for review?

catb0t:
Hey guys, this time of year (last year as well) is really busy for me in terms of school ending. This month, school is ending for good for me (finally), so it's very busy. 

The short story is that I will have way more time and brainpower come June. If you are curious, [here is the long story](https://gist.github.com/catb0t/6e9c68fa66988484fed47520cfad8809), for why my pulls / activity here have been so sluggish.

I'll entirely review this changeset, mark `v.` and `m.` as deprecated, make `determinant` etc `extras` things to come later, and this can be merged quite soon.

mrjbq7:
Ping @catb0t !

catb0t:
91% of `math.matrices` and 80% of `math.matrices.extras` are documented. `extras` contains unfinished words (which are already implemented in `extra/math/matrices/elimination` but they use a global `matrix` symbol), perhaps I'll keep them locally until they're not temporary definitions. 

I realised I haven't even touched `extra/math/matrices/simd`, but that's a very complicated vocab, so I'm not even going to fix things like the naming disparity between `math.matrices.extras:<rotation-matrix4>` and `math.matrices.simd:rotation-matrix4` for now, that's for a later PR.

I've fixed all the merge conflicts and made it pass `load-all test-all`. `vdot` etc is still in this PR, not another one.

basis/math/vectors/vectors-docs.factor
```diff
@@ -314,8 +314,8 @@ HELP: vclamp
   }
 } ;
 
-HELP: v.
-{ $values { "u" { $sequence real } } { "v" { $sequence real } } { "x" "a real number" } }
+HELP: vdot
+{ $values { "u" "a sequence of real numbers" } { "v" "a sequence of real numbers" } { "x" "a real number" } }
```
mrjbq7: Looks like a merge that lost the ``{ $sequence real }`` change to the docs.

BEFORE_FINAL_COMMIT
```diff
@@ -0,0 +1,20 @@
+- remove useless USING:s
```
mrjbq7: Probably don't want this file in the final squashed merge.

_update.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
mrjbq7: And this one.

basis/bootstrap/image/primitives/primitives-docs.factor
```diff
@@ -19,7 +19,7 @@ HELP: primitive-quot
 { $description "Creates the defining quotation for the primitive. If 'vm-func' is a string, then it is prefixed with 'primitive_' and a quotation calling that C++ function is generated." } ;
 
 ARTICLE: "bootstrap.image.primitives" "Bootstrap primitives"
-"This vocab contains utilities for declaring primitives to be added to the bootstrap image. It is used by the file " { $snippet "resource:core/bootstrap/primitives.factor" }
+"This vocab contains utilities for declaring primitives to be added to the bootstrap image. It is used by " { $vocab-link "bootstrap.primitives" }
```
mrjbq7: This will fail unless we re-factor to make ``core/bootstrap/primitives.factor`` an actual vocabulary instead of a script.

basis/math/matrices/matrices.factor
```diff
@@ -1,197 +1,118 @@
...
+    0 <matrix> ; inline
 
-: norm-gram-schmidt ( seq -- orthonormal )
-    gram-schmidt [ normalize ] map ;
+: <zero-square-matrix> ( n -- matrix )
+    dup <zero-matrix> ; inline
 
-ERROR: negative-power-matrix m n ;
+! main-diagonal matrix
+! running time is improved by 10% over the old implementation
```
mrjbq7: This is a good comment to have in the discussion but will quickly get dated in the source code.

basis/math/matrices/matrices.factor
```diff
@@ -200,77 +121,155 @@ ERROR: negative-power-matrix m n ;
...
+
+! well-defined for any well-formed-matrix
+: matrix-except ( matrix exclude-pair -- submatrix )
+    first2 [ rows-except ] dip cols-except ;
+
+:: matrix-except-all ( matrix-seq -- expansion )
+    matrix-seq dim [ <iota> ] map first2 cartesian-product
+    [ [ matrix-seq swap matrix-except ] map ] map ;
+
+: dim ( matrix -- dimensions )
```
mrjbq7: Probably should be ``matrix-dim`` for now, we have a lot of generic ``dim`` things?

mrjbq7:
Cool, I don't want your work to get lost.  It's gotten to be a larger set of things now.  Made a few comments and then we can merge a squashed commit.  We could also split out stuff like the ``v.`` to ``vdot`` rename to make the diff smaller.

BEFORE_FINAL_COMMIT
catb0t: it definitely won't be, it's temporary

basis/bootstrap/image/primitives/primitives-docs.factor
catb0t: This was merged wrong, I meant to keep the `$snippet` line

basis/math/matrices/matrices.factor
catb0t: sounds good

catb0t:
1. Tidy up last bits of remaining docs. 
2. Revert all `vdot` and `mdot` etc to `v.`, `m.`, `m.v` for a later PR.
2. Go through my "before final commit" todo list one last time to ensure high code quality.

catb0t:
Note to future me / others: `vdot` is used literally and separately from `math.vectors:v.` in files like `resource:basis/xmode/modes/latex.xml` and `resource:basis/xmode/modes/povray.xml`, as well as `resource:extra/webapps/fjsc/www/termlib/readme.txt`, and therefore non-.factor files must be excluded from the `[mv]dot[mv]?` or `[mv]\.[mv]?` regexes

mrjbq7:
Hah! 

Are you planning on splitting the rename into its own PR? Or would you like me to make that change and then we can take a look at the matrix change in parts or in total?

Thanks!

catb0t:
I'll split the rename into its own later PR, and this PR will be solely about the `math.matrices` and `math.matrices.extras` vocabs. 

The vdot rename PR will need to change things like `(simd-v.)` and `weighted-v.` (from `resource:extra/math/similarity/similarity.factor`) too, for consistency's sake.

catb0t:
Hey @erg - in https://github.com/factor/factor/commit/b23e52f532f803112fe00ace3ae6844ecc551e26 you created the `cartesian-matrix-map` (which is a row-map as expected) and `cartesian-matrix-column-map`. 

Previously I attempted to understand `cartesian-matrix-column-map` (which i have renamed to `cartesian-column-map` for brevity), but I couldn't understand it. 

Its definition (which is semantically identical to your original code) is:
```factor
: cartesian-column-map ( matrix quot: ( ... pair matrix -- ... matrix' ) -- matrix-seq )
    [ cols first2 ] prepose cartesian-matrix-map ; inline
```

Where `cartesian-matrix-map` is (also unchanged in operation since you wrote it) 
```factor
: cartesian-matrix-map ( matrix quot: ( ... pair matrix -- ... matrix' ) -- matrix-seq )
    [ [ matrix-dim <cartesian-indices> ] keep ] dip
    '[ _ @ ] matrix-map ; inline
```

(`<cartesian-indices>` is a generalisation/modernisation of the previous `: cartesian-indices ( n -- matrix ) <iota> dup cartesian-product ;`.)

`cartesian-column-map` does not work as a columnal analogue of `cartesian-matrix-map`; it works like so: 

```factor
USING: math.matrices io ;
SYMBOLS: a b c d e f g h ;

{ { a b c d } { e f g h } } 
[ [ "top: " write . "2nd: " write . ] keep ] cartesian-column-map
```

It prints 
```factor 
top: { a e }
2nd: { a e }
top: { b f }
2nd: { a e }
top: { c g }
2nd: { a e }
top: { d h }
2nd: { a e }
top: { a e }
2nd: { b f }
top: { b f }
2nd: { b f }
top: { c g }
2nd: { b f }
top: { d h }
2nd: { b f }
```

And the result is 
```factor
{
    { { a e } { b f } { c g } { d h } }
    { { a e } { b f } { c g } { d h } }
}
```
The shape of the output of `cartesian-column-map` is not correct: 
1. the input matrix has 4 columns and so there should be 4 output matrices, not 2. 
2. the output matrices should not be transposed, but these are. 

The same snippet with `cartesian-matrix-map` instead gives: 
```factor
top: { { a b c d } { e f g h } }
2nd: { 0 0 }
top: { { a b c d } { e f g h } }
2nd: { 0 1 }
top: { { a b c d } { e f g h } }
2nd: { 0 2 }
top: { { a b c d } { e f g h } }
2nd: { 0 3 }
top: { { a b c d } { e f g h } }
2nd: { 1 0 }
top: { { a b c d } { e f g h } }
2nd: { 1 1 }
top: { { a b c d } { e f g h } }
2nd: { 1 2 }
top: { { a b c d } { e f g h } }
2nd: { 1 3 }
```
and the result
```factor
{
    {
        { { a b c d } { e f g h } }
        { { a b c d } { e f g h } }
        { { a b c d } { e f g h } }
        { { a b c d } { e f g h } }
    }
    {
        { { a b c d } { e f g h } }
        { { a b c d } { e f g h } }
        { { a b c d } { e f g h } }
        { { a b c d } { e f g h } }
    }
}
```

The shape of the output is correct: 
1. The array contains 2 arrays, one for each row of the original matrix. 
2. Each of the 2 arrays contain 4 arrays, one for each column of the original matrix. 
3. The output matrices are not transposed without reason, compared to the input.

For `cartesian-matrix-map` the input to the quotation is correctly `pair matrix -- ..`, but the quotation input for `cartesian-column-map` seems to be like `last-column column -- ..`. 

There have never been tests or docs for these words, nor are they used anywhere. The semantics of `cartesian-column-map` now are still the same as originally in 2012.

These words are probably useful (when fixed), so I'll put them in `extra/math/matrices` for now, but I need your help and thought process for them to work properly. 

Thanks!

mrjbq7:
I'm excited to merge some of this stuff!

We also have recent contributions in the form of a ``tensors`` vocabulary which is the start of shaped floating point arrays in a manner similar to ``numpy.ndarray``.

basis/math/matrices/matrices.factor
catb0t: I finally committed the proper implementation that's just as fast as `flip`!

catb0t:
I've finished the final todo list, and finally written the fast anti-transpose implementation, so this is ready for merge. 

I was hoping to write more tests but the current ones are honestly perfectly serviceable  and writing more is better left to another PR because of how much longer that  could take.

mrjbq7:
Would it be possible for you to squash this into a single commit for merge?

catb0t:
This PR should rebase and/or merge without any hidden conflicts, sometimes GitHub web interface lies that there are "no conflicts". 

I'll squash it now.

catb0t:
I've reviewed all my changes, rebased from master, fixed erroneously regex-replaced files, and load-all test-all'd to ensure this PR is ready to go. 

Please take one last skim through the changes to make sure I haven't missed a mistake, and then it's ready for merge!

mrjbq7:
Woot!!

Merged!!

Thanks for pushing this through @catb0t !

mrjbq7:
Did @erg ever answer your question about `cartesian-matrix-map`?

catb0t:
Not yet, I might make that question an email to the whole mailing list to see if anyone else has ideas about it.

PR #1950, opened 2018-02-17
extra/pairs: 2 new words and tests

catb0t:
tiny changes to make it easier to use pairs. `>sequence` outputs the slots in the same order they were put in, rather than in `key value` order.

I considered implementing all of the sequence protocol so that you can say `1 2 <pair> [ 2 + ] map` and have it output a real sequence `{ 3 4 }`, but it seems overkill when all I want is a 2-element box that I can easily convert to an array.

I also considered moving the vocab to `basis` because I think it's useful enough to not be in `extra` but again, it's more than I want to bother with right now.

### DIFF
--- a/extra/pairs/pairs-tests.factor
+++ b/extra/pairs/pairs-tests.factor
@@ -21,3 +21,6 @@ SYMBOL: blah
 { f f } [ "fdaf" "b" get at* ] unit-test
 { 2 } [ "b" get assoc-size ] unit-test
 { { { 2 1 } { blah "bleah" } } } [ "b" get >alist ] unit-test
+{ V{ 1 2 } } [ 1 2 <pair> V{ } >sequence ] unit-test
+{ { 1 2 } } [ 1 2 <pair> >array ] unit-test
+
--- a/extra/pairs/pairs.factor
+++ b/extra/pairs/pairs.factor
@@ -38,4 +38,11 @@ M: pair delete-at
 M: pair >alist
     [ hash>> >alist ] [ [ key>> ] [ value>> ] bi 2array ] bi suffix ; inline
 
+: >sequence ( pair exemplar -- seq )
+    [ [ value>> ] [ key>> ] bi 2array ] dip like ;
+
+: >array ( pair -- array )
+    { } >sequence ;
+
 INSTANCE: pair assoc
+


mrjbq7:
I'm not sure the ``pairs`` vocab is the right place to start for a 2-element box, but...

1) Implementing the sequence protocol isn't such a big deal if that is actually desired:

```factor
M: pair length drop 2 ; 

M: pair nth-unsafe swap 0 = [ value>> ] [ key>> ] if ;

M: pair set-nth-unsafe swap 0 = [ value<< ] [ key<< ] if ;

INSTANCE: pair sequence
```

Alternatively, you can make it ``INSTANCE: pair immutable-sequence`` and not implement ``M\ pair set-nth-unsafe`` if you want it read-only.

2) Rather than using ``>array`` and ``>sequence`` which sound like "anything to array" and "anything to sequence", it might be better named ``pair>array`` and ``pair>sequence`` (and also wouldn't conflict with the ``>array`` word in ``arrays`` vocabulary).

3) We have a ``pair`` predicate in ``arrays`` which is a little useful in various places, but it's just an array of length 2.  The confusing thing about ``extra/pairs`` is its a funny named two-element box.  The first element is ``value`` and the second is ``key``?  That would be confusing if you used it for a dimension (``x/y`` or ``rows/cols`` or etc.)... if you need a box, why not just ``1 2 2array``.  

And it stores a ``hash`` which is maybe useful if using it in an ``assoc``, but no one really seems to be using the ``pairs`` vocab.

What were you hoping to use the ``pair`` type for?

mrjbq7:
Also, we have the ``tuple-slots`` word that gets all slots from a tuple as an array if thats useful for you.

mrjbq7:
Another approach I was playing with a few months ago:

```factor
MIXIN: named-tuple

M: named-tuple length tuple-size ;

M: named-tuple nth-unsafe array-nth ;

M: named-tuple set-nth-unsafe set-array-nth ;

INSTANCE: named-tuple sequence
```

Then:

```factor
TUPLE: foo x y z ;

INSTANCE: foo named-tuple
```

And using it:

```factor
IN: scratchpad T{ foo f 1 2 3 } >array .
{ 1 2 3 }
```

mrjbq7:
I cleaned up my ``named-tuple`` vocab and pushed it just now.

0ea0285e3398325fa3faa9c2085d57864d1f4f46

It also provides support for treating a tuple as an ``assoc`` in addition to ``sequence``.

Probably needs a little bit more testing, but is kind of fun.

catb0t:
That's really interesting info, thank you! I can see this feature was rather short-sighted.

> What were you hoping to use the `pair` type for?

Essentially for the players in a 2-player game, because there will always be exactly two players in [the Game of 21](https://github.com/catb0t/21-game/blob/v2.0/21-game/21-game.factor#L50-L65) -- but there is no reason to make a tuple for a little temporary object that is immediately transformed into an `interleaved`. 

If you write `computer <game-loop>` then it gives you an `interleaved` of `you` and `computer`, giving `you` the first turn. 

But if you extend the base Game of 21 to be 2 human players, by writing `SINGLETON: another-you INSTANCE: another-you player` then you can implement `M: another-you <game-loop> you <pair> <game-loop> ;` that interleaves `you` and `another-you` (and then just implement `M: another-you take-turn` and `M: another-you win` etc).

PR #1956, opened 2018-02-24
Fix #100

### LINKED ISSUE #100: Vocabularies defined higher up in the vocab-roots hierarchy shadow vocabularies defined later
"math" scaffold-work

This vocabulary is shadowed by core/math and will not reload.

Ideally, there would be an error message or a restart to rename the work/math vocabulary to something else.

AlexIljin:
Hey, guys! I had a free minute and decided to fix this pressing issue.
If you feel the copyright update is not earned, feel free to skip that commit.

### DIFF
--- a/basis/tools/scaffold/scaffold.factor
+++ b/basis/tools/scaffold/scaffold.factor
@@ -272,7 +272,19 @@ PRIVATE>
 : scaffold-platforms ( vocab platforms -- )
     [ "platforms.txt" ] dip scaffold-metadata ;
 
+: vocab-root-index ( vocab-root -- i/f )
+    vocab-roots get index ;
+
+: check-shadowed ( vocab-root string -- )
+    [ vocab-root-index ] [ find-vocab-root dup vocab-root-index ] bi*
+    swapd 2dup and [
+        < [ drop ] [
+            "Vocab with this name already exists in " prepend throw
+        ] if
+    ] [ 3drop ] if ;
+
 : scaffold-vocab ( vocab-root string -- )
+    2dup check-shadowed
     {
         [ scaffold-directory ]
         [ scaffold-main ]
--- a/basis/tools/scaffold/scaffold.factor
+++ b/basis/tools/scaffold/scaffold.factor
@@ -1,4 +1,5 @@
 ! Copyright (C) 2008 Doug Coleman.
+! Copyright (C) 2018 Alexander Ilin.
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors alien arrays assocs byte-arrays calendar
 classes combinators combinators.short-circuit fry hashtables
--- a/basis/tools/scaffold/authors.txt
+++ b/basis/tools/scaffold/authors.txt
@@ -1,1 +1,2 @@
 Doug Coleman
+Alexander Ilin
--- a/basis/tools/scaffold/scaffold-tests.factor
+++ b/basis/tools/scaffold/scaffold-tests.factor
@@ -1,4 +1,5 @@
 ! Copyright (C) 2009 Doug Coleman.
+! Copyright (C) 2018 Alexander Ilin.
 ! See http://factorcode.org/license.txt for BSD license.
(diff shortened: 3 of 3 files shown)


basis/tools/scaffold/scaffold.factor
```diff
@@ -272,7 +273,19 @@ PRIVATE>
 : scaffold-platforms ( vocab platforms -- )
     [ "platforms.txt" ] dip scaffold-metadata ;
 
+: vocab-root-index ( vocab-root -- i/f )
+    vocab-roots get index ;
```
mrjbq7: I probably wouldn't make this a new word since it's only used in one place...

mrjbq7: Also, isn't this a simpler shadow check to just see if the vocab name has a ``find-vocab-root`` already?

```factor
ERROR: shadowed-vocab name root ;

: check-shadowed ( string -- )
    dup find-vocab-root [ shadowed-vocab ] [ drop ] if* ;
```

mrjbq7: Or maybe, since some vocabularies like ``accessors`` and ``cocoa.classes`` are dynamic, this is better just to check if it exists:

```factor
ERROR: vocab-exists name ;

: check-vocab-exists ( name -- name )
    dup vocab-exists? [ vocab-exists ] when ;
```

AlexIljin: @mrjbq7: "I probably wouldn't make this a new word since it's only used in one place..."
It's called twice, so it's more readable this way, no?

Regarding the `find-vocab-root`, the idea is that we want to prevent scaffolding of vocabs that would already be shadowed on creation (see #100), so that reloading them would fail. If, on the other hand, we want to create a vocab that shadows an existing one - that's not a error, see unit-tests.

mrjbq7: I saw that in the unit tests, but I think maybe that we should be more conservative, to not allow creating a shadowed vocab or a shadowing vocab, no?

AlexIljin: @mrjbq7: maybe you're right, and the conservative approach is better.

I'm not sure the `vocab-exists` would work well for the error tuple. `vocab-exists?` clashes with the class checker word.

mrjbq7: Does it clash?  I'm using the ``vocabs.loader:vocab-exists?`` word.

Maybe it shouldn't be called ``check-vocab-exists`` since we have ``ensure-vocab-exists``, so maybe a different name like ``check-new-vocab`` or something.

AlexIljin: I'm just saying that we have explicitly defined `vocabs.loader:vocab-exists?` and implicitly (by creating a tuple) defined `tools.scaffold:vocab-exists?`. That creates ambiguity.
Or do you think it's OK to have that as long as the two vocabs are unlikely to be imported at the same time? I'm interested to know your approach to resolving such situations in general.

mrjbq7: Ooooooohhh, yes. 

Maybe better to say something easier for the user to understand like:

```factor
ERROR: vocab-must-not-exist name ;
```

AlexIljin: Thanks for the reply!

Are we waiting for more feedback here, or is it decided? Would you like to commit your proped code yourself, or would you like me to redo the PR?
I'm fine either way. I'm just glad we are reducing the issue count. : ))

mrjbq7: Sure, fixed in 68eb6833872571a8e931cf2246838e5e0e619052.

mrjbq7:
Thank you!

bjourne:
100 prs already? Wow!! :cake: :birthday: :cookie: :beers:

AlexIljin:
@bjourne: LOL, no. This PR fixes issue with id = 100. The PR statistics for Factor currently is: 18 open, 458 closed.

PR #1964, opened 2018-03-04
checksums.multi: new vocab

AlexIljin:
Hello, guys! A new cool vocab for your consideration!
Allows you to calculate multiple checksums in one pass over a data stream. I created it when I converted my DBs from SHA-1 to Blake2b: naturally you want the old hash to match before you replace it with the new one, so this saved me a full extra read of the harddrive.

Today I've added documentation, unit-tests, ran `help-lint` on it, and it's done.

### DIFF
--- a/extra/checksums/multi/multi.factor
+++ b/extra/checksums/multi/multi.factor
@@ -1,0 +1,26 @@
+! Copyright (C) 2018 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: accessors checksums checksums.common destructors fry kernel
+sequences ;
+IN: checksums.multi
+
+TUPLE: multi-checksum checksums ;
+INSTANCE: multi-checksum block-checksum
+C: <multi-checksum> multi-checksum
+
+TUPLE: multi-state < disposable states results ;
+
+M: multi-checksum initialize-checksum-state
+    checksums>> [ initialize-checksum-state ] V{ } map-as
+    multi-state new-disposable swap >>states ;
+
+M: multi-state dispose*
+    states>> dispose-each ;
+
+M: multi-state add-checksum-bytes
+    '[ [ _ add-checksum-bytes ] map! ] change-states ;
+
+M: multi-state get-checksum
+    dup results>> [
+        dup states>> [ get-checksum ] { } map-as [ >>results ] keep
+    ] unless* nip ;
--- a/extra/checksums/multi/authors.txt
+++ b/extra/checksums/multi/authors.txt
@@ -1,0 +1,1 @@
+Alexander Ilin
--- a/extra/checksums/multi/multi-docs.factor
+++ b/extra/checksums/multi/multi-docs.factor
@@ -1,0 +1,62 @@
+! Copyright (C) 2018 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: checksums checksums.common checksums.md5 checksums.sha
+destructors help.markup help.syntax kernel sequences ;
+IN: checksums.multi
+
+ABOUT: "checksums.multi"
+
+ARTICLE: "checksums.multi" "Checksumming with Multiple Algorithms"
+"The " { $vocab-link "checksums.multi" } " vocabulary makes it possible to calculate multiple checksums with just one pass over a data stream."
+$nl
+"The " { $link multi-checksum } " tuple holds a sequence of " { $link block-checksum } " instances, such as " { $link md5 } " or " { $link sha1 } ". When the " { $link initialize-checksum-state } " method is called on it, a new instance of " { $link block-checksum-state } " is created for all the " { $slot "checksums" } ", and returned as a new " { $link multi-state } " instance. You can then use " { $link add-checksum-bytes } " to stream data to it. When done, call " { $link get-checksum } " to finalize the process, read the resulting checksums and dispose of the tuple in one step. If you want to cancel the work without calling " { $link get-checksum } ", you must " { $link dispose } " of the tuple so that all implementation-specific resources are released."
+$nl
+"The " { $link checksum-bytes } " and the " { $link checksum-stream } " methods encapsulate the above protocol, including instantiation and disposal of the " { $link multi-state } " tuple."
+{ $examples
+    { $unchecked-example "USING: byte-arrays checksums checksums.md5 checksums.sha ;"
+    "\"test\" >byte-array { md5 sha1 } <multi-checksum> checksum-bytes ."
+    "{
+    B{
+        9 143 107 205 70 33 211 115 202 222 78 131 38 39 180 246
+    }
+    B{
+        169 74 143 229 204 177 155 166 28 76 8 115 211 145 233
+        135 152 47 187 211
+    }
+}" }
+    $nl
+    { $unchecked-example "USING: checksums checksums.common checksums.md5 checksums.sha"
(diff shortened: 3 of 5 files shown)


extra/checksums/multi/multi.factor
```diff
@@ -0,0 +1,31 @@
...
+C: <multi-checksum> multi-checksum
+
+TUPLE: multi-state < disposable states results ;
+
+M: multi-checksum initialize-checksum-state
+    list>> [ initialize-checksum-state ] V{ } map-as
+    multi-state new-disposable swap >>states ;
+
+M: multi-state dispose*
+    states>> [ [ dispose ] each ] when* ;
```
mrjbq7: ``states>> dispose-each``

extra/checksums/multi/multi.factor
```diff
@@ -0,0 +1,31 @@
...
+
+M: multi-state add-checksum-bytes
+    '[ [ _ add-checksum-bytes ] map! ] change-states ;
```
mrjbq7: Not necessary to ``change-states``... since you're using ``map!``.

```[ states>> ] keep '[ _ add-checksum-bytes ] map! drop``?

File: extra/checksums/multi/multi.factor [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+
+M: multi-state get-checksum
```
mrjbq7: Why are you caching the results?

```states>> [ get-checksum ] map``

File: extra/checksums/multi/multi.factor [hunk 4]
```diff
@@ -0,0 +1,31 @@
...
+IN: checksums.multi
+
+TUPLE: multi-checksum list ;
+INSTANCE: multi-checksum block-checksum
+C: <multi-checksum> multi-checksum
+
+TUPLE: multi-state < disposable states results ;
+
+M: multi-checksum initialize-checksum-state
+    list>> [ initialize-checksum-state ] V{ } map-as
```
mrjbq7: Couple things here:

1) Don't like the name ``list``, maybe ``checksums``.

2) Block checksums create ``block-checksum-state`` that have ``block-size``.  That's used to split the blocks into groups of the correct size when ``checksum-bytes`` is called.

So, I would probably make ``multi-state`` be a ``block-checksum-state``, and assert that all checksum-states that are created have the same ``block-size`` (otherwise they aren't going to work properly).  For example, ``md5`` might have 64-byte blocks, but ``metrohash-64`` has 32-byte blocks.

mrjbq7: You also, might want to implement ``checksum-bytes``:

```factor
M: multi-checksum checksum-bytes
    list>> [ checksum-bytes ] with map ;
```

mrjbq7: Actually, nevermind about the ``block-checksum-state``.  Your delegation of ``add-checksum-bytes`` is just fine.

mrjbq7:
Do you have an implementation of blake2b to contribute?

extra/checksums/multi/multi.factor
AlexIljin: We need to keep the object instance on the stack, so it's actually
```
    [ dup states>> ] dip '[ _ add-checksum-bytes ] map! drop ;
```
vs.
```
    '[ [ _ add-checksum-bytes ] map! ] change-states ;
```
And the latter wins the Code Golf, am I right?

mrjbq7: Hah, sure I'm just saying since you're mapping in place you don't need the "set states".  But sure, it's not really a performance issue.

extra/checksums/multi/multi.factor
AlexIljin: Actually, I'm afraid that some of the disposed states may not handle the repeated `get-checksum` well. Not sure if that's a real issue, or an imagined one.

extra/checksums/multi/multi.factor
mrjbq7: And also, if you're a ``block-checksum``, then you inherit ``checksum-bytes`` which just defers to your ``checksum-state`` implementation.

extra/checksums/multi/multi.factor
mrjbq7: Why would you make repeated calls to ``get-checksum``?  Seems like that's not necessarily supported by any checksum so why add caching here?

extra/checksums/multi/multi.factor
AlexIljin: Nice one, will fix!

extra/checksums/multi/multi.factor
AlexIljin: Well, it's kind of "there" in the API: you can call `get-checksum` whenever you want. I'm not calling it twice in any of my code, but it's not explicitly forbidden in the documentation, so I support the use case. Why not call `get-checksum` several times after the data is piped? You have the finished state, you can request the result as many times as you like. The data are there, in the object.

I never saw it unsupported or forbidden or "not guaranteed", so I support it in my hashes. E.g. the `checksums.process` vocab keeps the data after the underlying process is terminated. I think our API/protocol is not very clear in this regard, so I'm trying to be on the safe side.

mrjbq7: That makes sense, but I'd defer the caching to the checksums you're using rather than putting it in the multi instance personally.

But if you do that, I'm not sure I would dispose the ``states``, since that's being done in ``M\ multi-state dispose``.

```factor
M: multi-state get-checksum
    dup results>> [
        dup states>> [ get-checksum ] map [ >>results ] keep
    ] unless* nip ;
```

AlexIljin:
@mrjbq7: "Do you have an implementation of blake2b to contribute?"
Unfortunately, the answer is no. I'm currently using a combination of `checksums.process` with "b2sum.exe" from the Git for Windows package, and a (normally unpublished) functions from a [custom build](https://github.com/AlexIljin/phc-winner-argon2) of the [Argon2](https://github.com/P-H-C/phc-winner-argon2) DLL.

I had my eyes on the highly optimized [Go implementation](https://github.com/minio/blake2b-simd), but last time I checked the language still didn't support creating the Windows DLLs [very well](https://github.com/z505/goDLL).

I use both versions depending on the file size to be hashed: b2sum.exe has much better performance, but for smaller files the overhead of starting the external process outweighs the speed advantage. The experimentally determined boundary is about 22Mb: smaller files will be hashed by the DLL faster. And since there are typically many more small files on a harddrive (think photos and mp3s), than there are large ones, the combined approach gives significant benefits on a 2Tb disk.

I'm not interested in (contributing) a native Factor implementation, because I need performance to be IO-bound, not CPU-bound.

extra/checksums/multi/multi.factor
AlexIljin: I think the last change you suggested is acceptable. I wanted to dispose of things as early as possible (in case user forgets to call `dispose` on the `multi-state`, and only ever calls the `get-checksum`), but I guess it mixes up the responsibilities between the two methods. I wanted the stronger, more encapsulated guarantees of the protocol, and you suggest we let the weaker implementations seep through to the surface. I can roll with that, even though it's not how I normally prefer to build systems.

extra/checksums/multi/multi.factor
AlexIljin: Renamed `list` to `checksums`.

AlexIljin:
All comments fixed.

mrjbq7:
Merged, thanks!!

PR #1965, opened 2018-03-04
Misc

AlexIljin:
Some other random stuff I have accumulated.


basis/db/sqlite/sqlite.factor
```diff
@@ -161,6 +161,7 @@ M: sqlite-db-connection persistent-table ( -- assoc )
         { +user-assigned-id+ { f f f } }
         { +random-id+ { "integer" "integer" f } }
         { +foreign-id+ { "integer" "integer" "references" } }
+        { +primary-key+ { f f "primary key" } }
```
mrjbq7: do you have a test case that shows this working where it failed before?

AlexIljin: Of course:
```
USING: db db.tuples db.types db.sqlite ;
TUPLE: t data ;
t "t" { { "data" "data" +primary-key+ } } define-persistent
"test.db" <sqlite-db> [ t ensure-table ] with-db
```

```
no-sql-type
type +primary-key+
To continue, pick one of the options below:
Abort
```

mrjbq7:
Merged, thanks!

PR #1983, opened 2018-03-22
basis/math: add logn to math.functions (fix #1981)

### LINKED ISSUE #1981: Factor omits a logn function
perhaps it has been intentionally omitted or I just can't find it, but there is no `logn` function I can see in the standard library. 

`logn(x) = log(x) / log(n)` or `: logn ( n x -- log ) [ log ] bi@ swap / ;` , so it is very easy to implement logn.

like in #1980, unless someone tells me otherwise i will create a pull request with this feature.

catb0t:


mrjbq7:
Merged, thanks!

PR #1993, opened 2018-05-06
make the caret color themeable

twopir:
Move the editor caret color into a theme hook instead of delegating it to each editor instance.


mrjbq7:
Cool.  Do you want to use different caret colors in each theme?

twopir:
Decided to pick a different color for the caret in dark-theme fyi.

mrjbq7:
Merged, thanks!

PR #2002, opened 2018-06-01
math.functions docs: add logn to "powers and logarithms" article

catb0t:
logn function was omitted from the subsections in "powers and logarithms"


mrjbq7:
Merged, thanks!

PR #2004, opened 2018-06-17
Nipd branch

erg:
All the stack shufflers you need but were afraid to ask for in core/. Taking opinions...

To bootstrap: http://downloads.factorcode.org/images/nipd/
``./factor -i=boot.unix-x86.64.image``


extra/xkcd/xkcd.factor
```diff
@@ -13,8 +13,8 @@ IN: xkcd
 
 : comic-image ( url -- image )
     http-get nip
-    R/ http:\\/\\/imgs\.xkcd\.com\\/comics\\/[^\.]+\.(png|jpg)/
-    first-match >string load-http-image ;
+    R/ \/\/imgs\.xkcd\.com\\/comics\\/[^\.]+\.(png|jpg)/
+    first-match >string "http:" prepend load-http-image ;
```
mrjbq7: ?

erg: It broke. This should go into master/ anyway, oops

AlexIljin:
I like this branch : ))

jonenst:
You could also replace the "curry keep" usages with the new keepd ?
```
$ grep "curry keep" -ri core/ basis/ extra/
core/io/io.factor:    [ 2dup (new-sequence-for-stream) swap ] dip curry keep
core/io/io.factor:    [ swap (new-sequence-for-stream) ] curry keep
core/io/io.factor:    [ [ stream-read-unsafe ] curry keep resize ] bi ; inline
core/io/streams/sequence/sequence.factor:    ] dip [ <sequence-copy> (copy) drop ] 3curry keep ; inline
core/sequences/sequences.factor:    swapd [ curry keep ] curry ; inline
basis/classes/struct/struct.factor:    [ [ (writer-quot) call( value struct -- ) ] with 2each ] curry keep ;
basis/classes/struct/struct.factor:            over [ swapd [ call( value struct -- ) ] curry keep ] [ 2drop ] if
basis/formatting/formatting.factor:    ] curry keep neg? [ CHAR: - prefix ] when ;
basis/formatting/formatting.factor:    ] curry keep neg? [ CHAR: - prefix ] when ;
extra/noise/noise.factor:    [ [ v> -1.0 float-4-with vand ] curry keep v+ ] bi ; inline
```

erg:
@jonenst Good call, missed that one

erg:
Merged in 9f213f96f6589534d2ed5b3875ca05386b9afd5e

PR #2006, opened 2018-06-26
Nipd2

erg:
test some stuff


erg:
Seems to work -- it downloaded the nipd2 boot image!

PR #2000, opened 2018-05-20
elevate.py port to basis/elevate

catb0t:
this is a near-exact port of https://github.com/barneygale/elevate | #1994 

1. should this go in `extra` or `basis`? 
2. i'm looking for help to finish and test the Windows and Mac OSX implementations. I have access to a Windows machine, but not a Mac or Applescript.

3. specifically: i don't know how to use `with-out-parameters` properly (to avoid `local-allocation-error`) with `GetProcessToken` etc to implement `windows already-root?`

4. the documentation is almost done, but not quite there yet, as there are a bunch of unfixable-bugs (in e.g `gksudo` messaging) and limitations that need to be mentioned somewhere.

5. (when this is merged, `apple-script-quote` should probably be moved into `cocoa.apple-script`)

PR #2010, opened 2018-07-05
Elevate erg

erg:
Test me, baby!

Rebased and fixed circularity (?) on mac...

Same as one patch more + https://github.com/factor/factor/pull/2000


basis/elevate/macosx/macosx.factor
```diff
@@ -0,0 +1,28 @@
+USING: accessors arrays cocoa.apple-script elevate
+elevate.unix.private formatting io.launcher kernel locals
+sequences system ;
+IN: elevate.macosx
+
+<PRIVATE
+: apple-script-elevated ( command -- )
+    first quote-apple-script
+    "do shell script %s with administrator privileges without altering line endings"
```
erg: The ``%s`` here needs to be rethought. We have a command as ``{ "ls" "-al" }`` and that doesn't work with the format string here.

basis/elevate/macosx/macosx.factor
```diff
@@ -0,0 +1,28 @@
...
+! TODO
+M:: macosx elevated ( command replace? win-console? posix-graphical? -- process )
+    already-root? [
+        <process> command >>command 1array
+    ] [
+        ! graphical through applescript
+        posix-graphical? [
+            command apple-script-elevated
+        ] when
+        posix-elevated  "lol3" throw
```
erg: i left in

basis/elevate/macosx/macosx.factor
```diff
@@ -0,0 +1,28 @@
...
+    ] if "lol" throw ;
```
erg: some debugging...oops

basis/elevate/macosx/macosx.factor
catb0t: https://github.com/barneygale/elevate/blob/master/elevate/posix.py#L37

catb0t: the format argument in the Python is `quote_applescript(quote_shell(args))`, which is supposed to be replaced by `quote-apple-script` but I couldn't find our `quote-shell` word even though i know we have one

catb0t:
GKSudo is deprecated, and needs to be demoted from the primary method on Linux. PKExec is very complicated to get working properly and more code is needed (https://wiki.archlinux.org/index.php/Running_GUI_apps_as_root)

catb0t:
for GUI applications to work with PKExec there needs to be an .xml file in a system folder, so we have to do a "first-time setup" thing where we ask to create the file  https://askubuntu.com/questions/288087/can-i-use-pkexec-in-a-python-script-or-a-desktop-file

catb0t:
I don't understand why the CI is failing for this also? None of the tests failed and none of the programs exited with 1.

 I'm gonna pull this branch but I think maybe if you push again the CI might run properly?

mrjbq7:
Merged, thanks.

PR #2018, opened 2018-07-12
Windows registry

AlexIljin:
Hey, guys! I've added a new word that allows you to set values in the Windows registry. What do you think?

### DIFF
--- a/basis/windows/advapi32/advapi32.factor
+++ b/basis/windows/advapi32/advapi32.factor
@@ -1317,7 +1317,14 @@ FUNCTION: LONG RegDeleteKeyExW (
 ALIAS: RegDeleteKeyEx RegDeleteKeyExW
 
 ! : RegDeleteValueA ;
-! : RegDeleteValueW ;
+
+FUNCTION: LONG RegDeleteValueW (
+        HKEY    hKey,
+        LPCWSTR lpValueName
+    )
+
+ALIAS: RegDeleteValue RegDeleteValueW
+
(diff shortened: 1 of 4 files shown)


mrjbq7:
Any chance you could make a test for that?  like flip-flop a registry value, or increment, or something?

AlexIljin:
An example usage would be something like this:
```
USING: windows.registry windows.advapi32 fry ;
: add-to-PATH ( path -- )
    [ HKEY_CURRENT_USER "Environment" "Path" ] dip
    '[ _ ";" glue ] change-registry-value ; 
```
There is no function to create a value, all I could do is to append something to PATH, and then remove it again.

mrjbq7:
Well, some kind of test, for some kind of registry value, maybe something thats not important unlike the path?

erg:
The CI is spinning, glorious!

AlexIljin:
Why did the build tests failed? Did I break something?

AlexIljin:
I'm working on a test scenario.

AlexIljin:
I've added a test, and a new word that was necessary (`delete-value`) for the test to run cleanly.

mrjbq7:
@erg you want to look this over and merge and/or comment on this PR?

erg:
It works, squashed/merged. Thanks!

erg:
Oh, and I think the ``leaks`` tests need to be fixed a bit. It seems like we still have random errors.

PR #2020, opened 2018-07-15
vm: WIP remove warnings, use C++ style casts...there are hundreds of …

erg:
…warnings left.

Work in progress, this seems like a monumental undertaking for questionable gain. There are a couple of things I might commit separately, like a stray semi-colon.

Any thoughts on doing this cleanup?

### DIFF
--- a/vm/master.hpp
+++ b/vm/master.hpp
@@ -1,13 +1,13 @@
-#ifndef __FACTOR_MASTER_H__
-#define __FACTOR_MASTER_H__
+#ifndef FACTOR_MASTER_H
+#define FACTOR_MASTER_H
 
-#ifndef _THREAD_SAFE
-#define _THREAD_SAFE
-#endif
+//#ifndef _THREAD_SAFE
+//#define _THREAD_SAFE
+//#endif
 
-#ifndef _REENTRANT
-#define _REENTRANT
-#endif
+//#ifndef _REENTRANT
+//#define _REENTRANT
+//#endif
 
 #include <errno.h>
 
@@ -139,4 +139,4 @@ namespace factor { struct factor_vm; }
 #include "mvm.hpp"
 #include "factor.hpp"
 
-#endif // __FACTOR_MASTER_H__
+#endif // FACTOR_MASTER_H
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ addons:
       - cmake
       - libaio-dev
       - libsnappy-dev
+      - clang-6.0
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ./build.sh deps-macosx ; else ./build.sh deps-apt-get ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew ls --versions snappy > /dev/null || brew install snappy; fi
@@ -61,6 +62,7 @@ before_install:
   - git remote set-branches --add origin clean-macosx-x86-64
   - git remote set-branches --add origin clean-macosx-x86-32
   - git fetch # so we can see which vocabs changed versus origin/master...
+  - clang --version
 script:
   - echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, TRAVIS_PULL_REQUEST_BRANCH=$TRAVIS_PULL_REQUEST_BRANCH"
   - export CI_BRANCH="${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ addons:
       - cmake
       - libaio-dev
       - libsnappy-dev
+      - libc++-dev
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ./build.sh deps-macosx ; else ./build.sh deps-apt-get ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew ls --versions snappy > /dev/null || brew install snappy; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ branches:
   - clean-macosx-x86-32
 addons:
   apt:
+    update: true
     packages:
       - links
       - libblas-dev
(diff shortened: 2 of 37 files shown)


catb0t:
I really like this, a road to the VM being cleanly documented, well-defined and stuff. 

But, you meant to use `-Wall` as in `-W all`, there is no `-Weverything` option in any compiler.

erg:
@catb0t ``-Wall`` already exists in the makefile.

``-Weverything`` is the whole point of this branch, it enables a ton more warnings. Try it!

catb0t:
oh, i think GCC has no `Weverything` but clang does, i've just never seen it documented anywhere.

vm/master.hpp
```diff
@@ -1,13 +1,13 @@
...
-#define _THREAD_SAFE
-#endif
+//#ifndef _THREAD_SAFE
+//#define _THREAD_SAFE
+//#endif
 
-#ifndef _REENTRANT
-#define _REENTRANT
-#endif
+//#ifndef _REENTRANT
```
mrjbq7: Why is this changed?

nomennescio:
It might not be a fashionable opinion, but in many typical cases using 'nullptr' is overkill over the literal '0', which has been and still is the ISO C++ standard's way of unambiguously expressing a null pointer (and the only way for ISO C). Only in cases where a literal (!) '0' is used as argument to an overloaded function with both integer and pointer argument variants (or templated variants thereof), is 'nullptr' needed to disambiguate. In my opinion denser code does give improved readability and is less error-prone.

mrjbq7:
Some discussion here, it suggests the opposite:

> Code will become more safe, intuitive, and expressive. ``if (ptr == nullptr);`` rather than ``if (ptr == 0);``

https://dzone.com/articles/what-exactly-nullptr-is-in-c

bpollack:
As someone who wrote C++ full-time for almost a decade, I'm with dzone, for whatever it's worth

nomennescio:
> Some discussion here, it suggests the opposite:
> 
> > Code will become more safe, intuitive, and expressive. `if (ptr == nullptr);` rather than `if (ptr == 0);`
> 
> https://dzone.com/articles/what-exactly-nullptr-is-in-c

I don't want to nitpick and throw around with credentials, but I've been developing C and C++ software for 20+ years, including work on C++ compilers, and a popular article doesn't really cut it in my opinion compared to the C++ standard (I won't go into the details). Using `nullptr` is not wrong, it's just in many cases overkill and not idiomatic. The mentioned example is a prime example of idiom how you should preferably NOT deal with null pointers. To be clear, we're *only* talking about null pointer *literals* here

Instead of
```
if (ptr == nullptr)
```
it's more idiomatic to write (and for sure shorter)
```
if (!ptr)
```

See also section 7.3.15 of the C++ standard (e.g. https://open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4868.pdf):
"A prvalue of arithmetic, unscoped enumeration, pointer, or pointer-to-member type can be converted to a
prvalue of type bool. A zero value, *null pointer value*, or null member pointer value is converted to false;
any other value is converted to true.", see also 7.3.1 on boolean conversions.

Note that by this rule, the above code is *exactly* equivalent to:
```
if (ptr == false)
```
and
```
if (!ptr)
```
Now comparing with literal `false` does seem silly, at least it does to me (Reminds me of idiom used by beginners: `if (v) { return true; } else { return false; }` instead of `return v;`)

The C++ standard has (e.g. https://open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4868.pdf) in section 7.3.12:
"*A null pointer constant is an integer literal (5.13.2) with value zero* or *a prvalue of type std::nullptr_t*. A
null pointer constant can be converted to a pointer type; the result is the null pointer value of that type (6.8.3)
and is distinguishable from every other value of object pointer or function pointer type. Such a conversion
is called a null pointer conversion. Two null pointer values of the same type shall compare equal."

The only place where the distinction is relevant is expressions where a null pointer constant (!) is passed to a function or template, that has explicit overloading on the corresponding argument for both integer and pointer type. *ONLY* in those case (i.e. when passing *literals*, *NOT* passing pointers variables) the compiler cannot decide if the integer literal with value (!) zero is a null pointer constant or an integer constant. 

I can assure you those cases are extremely rare, and in my opinion don't warrant the use of verbose and "breaking" idiom, "just to be sure", when there's no issue at all.

On the same footing, I don't like the use of the `NULL` macro, as it basically predates modern C *and* C++ standards, and serves no other purpose than to accommodate ancient compilers that do not properly support null pointer constants. Old habits die slowly, old compilers even slower.

To summarize: using `0` for a null pointer constant is 100% idiomatic C++, and in most of the cases (setting pointer variables) very clear. The rare cases where an explicit `nullptr` is required doesn't warrant its use everywhere.

I've become quite used to the somewhat quirky stances the broader C++ community sometimes has developed, so I'm not surprised to find myself at odds with some of its "wisdoms". As such the Factor community is a breath of fresh air were many sane and inspiring choices have been made, and which I truly enjoy.

To be sure, I'm just expressing my argued opinion, if you have a different conviction in this, that's fine.

mrjbq7:
That's a fair point -- in that I really dislike ``if (foo == false)`` or ``if (foo == true)`` when foo has an implied truthiness.  So if the cleanup is ``if (!foo)`` instead of ``if (foo == NULL)`` or ``if (foo == nullptr)`` then I can get behind that.

nomennescio:
> That's a fair point -- in that I really dislike `if (foo == false)` or `if (foo == true)` when foo has an implied truthiness.
> 
> So if the cleanup is `if (!foo)` instead of `if (foo == NULL)` or `if (foo == nullptr)` then I can get behind that.

When rereading what I wrote, I noticed I made a mistake; in `if (ptr==nullptr)`, `nullptr` is not actually implicitly converted to `false`, although in other contexts it could be.  Because of the truthiness of pointers, it's only "as if" you would write `if (ptr==false)`. But still, in `if (!ptr)`, the value of the pointer *is* converted to a truth value, and if it's a null pointer value, that would be `false`. So it remains that pointers have an implied truthiness and `if (!ptr)` is more idiomatic.

mrjbq7:
Closing for now. I recommend a rebasing against the new git branch and redoing this at some point.

PR #2028, opened 2018-08-02
gitignore: ignore macOS and Unix binaries

bpollack:
I also sorted the `.gitignore` so future diffs will be easier. The sorting and the addition are separate commits, in case we decide that's gross.


.gitignore
```diff
@@ -1,33 +1,34 @@
...
+*.so
+*~
+.#*
 .*.swm
 .*.swn
 .*.swo
 .*.swp
+.DS_Store
+.gdb_history
+/factor
```
bpollack: Other option here would just be too make it `factor`, which removes the need for the following line, but I wasn't sure that was actually desirable.

.gitignore
```diff
@@ -1,33 +1,34 @@
...
+Factor.app/Contents/MacOS/factor
+Factor.app/Contents/_CodeSignature
+a.out
+checksums.txt
+factor.com
+factor.image.fresh
 logs
```
bpollack: I *suspect* we may want this and the following line to begin a `/` so that they only apply in the repo root, but I didn't fold that into this PR.

.gitignore
mrjbq7: I think that makes sense.

mrjbq7: Oh wait.

.gitignore
mrjbq7: I meant to say *this* makes sense -- apply only at repo root.

bpollack: I'll update the PR with what I think we're agreeing on and then do a sanity check before merging. :)

.gitignore
```diff
@@ -1,33 +1,34 @@
...
-work
-*.bak
-.#*
-checksums.txt
-*.so
-a.out
+.DS_Store
+.gdb_history
+/factor
+/logs
```
bpollack: Okay, added `/logs` and `/work` to rooted ignores.

.gitignore
bpollack: I'm not sure I understand how that relates to this diff; `factor` in `.gitignore` won't match e.g. `zealot.factor`, just any file named exactly `factor` at any point in the directory hierarchy. If you want to ban anything with the word `factor` in it, it'd need to be `*factor*`. (Which I'm assuming we don't want, but just saying.)

If you're saying that zealot negates the need for this rule entirely, I don't think that's correct, either; running `./build.sh update` was why I proposed adding these rules.

I might be missing something here, though; zealot's new, and I may be misunderstanding your point.

bpollack:
On a totally different note, I have no idea why Travis is dying here. Like, I get that it's trying to clone a nonexistent branch name, but I don't get from the `.travis.yml` why it's trying to do that in the first place.

erg:
https://api.travis-ci.org/v3/job/411477118/log.txt

It's failing to lint check ``with-scoped-allocation`` because it leaks some resources. I don't think it really leaks things, they are cleaned up with promises that aren't fulfilled when the ``leaks`` word finishes.

Search the logs for ``==== FAIL`` to find lint and other errors.


We had ``factor`` in the .gitignore and it masked ``extra/zealot/factor/factor.factor``

erg:
I'm making that an ``$unchecked-example`` until we can figure out why it's breaking.

Also, I think you know, but PRs have to be merged to factorcode which pushes to github every 5 minutes.

bpollack:
@erg Yep, I know. Does that mean you're cool with me landing this?

erg:
Sure!

PR #2033, opened 2018-08-05
docs for splitting

AlexIljin:
Some documentation cleanup and a minor update (I was looking for the word like `split-indices`, but it was not in the help topic, so I could not find it).


mrjbq7:
Merged, thanks!

PR #2034, opened 2018-08-05
Rosetta fix

AlexIljin:
Some minor fixes in the `rosetta-code.ternary-logic` vocab. Just something I stumbled over while browsing randomly.


mrjbq7:
Merged, thanks!

PR #2035, opened 2018-08-05
splitting.extras[-docs|-tests]: add split-lengths

AlexIljin:
`split-lengths` is a new word that I could not find anywhere. It can be useful to split streams into chunks, for example when interpreting things like [AUTHENTICATE_MESSAGE](https://msdn.microsoft.com/en-us/library/cc236643.aspx).


bpollack:
I'm missing something. Isn't this the same as `grouping.group`?

mrjbq7:
Maybe it’s `cum-sum0 split-indices`?

AlexIljin:
`cum-sum split-indices` fits the bill, and actually performs twice as fast as my implementation.

PR #2038, opened 2018-08-11
furnance.auth: fix documentation

bpollack:
The docs were attempting to reference the CouchDB auth provider, but that fails, because the help vocab didn't include `furnance.auth.couchdb`.

That said, the CouchDB auth provider has such major caveats with it (see the docs) that I think I personally might prefer axing the reference entirely. Thoughts?


mrjbq7:
+1 for axing. 

Or making it a string vocab link so it doesn’t pull extra/furnace/auth/couchdb in when furnace.auth is imported.

bpollack:
If you're on board axing, I'll just do that. Tomorrow AM I'll just delete the `furnace.auth.couchdb` reference and land that plus the grammar cleanup.

PR #2041, opened 2018-08-12
add a FreeDesktop Desktop Entry file that users may customise and use

catb0t:
This is an executable text file which tells FreeDesktop-compliant desktop environments information about the Factor program, so that it can be included in menus and other application launchers (mostly in BSD/Linux desktops). 

`Version` refers to the .desktop spec version, not the program version.

to use this file as a launcher, it should be copied to a location like `/home/$USER/.local/share/applications/` and the paths should be changed to real ones.

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

### DIFF
--- a/misc/Factor.desktop
+++ b/misc/Factor.desktop
@@ -1,0 +1,14 @@
+#!/usr/bin/env xdg-open
+[Desktop Entry]
+Type=Application
+Version=1.0
+Name=Factor IDE
+Comment=Factor Programming Environment (Development Mode)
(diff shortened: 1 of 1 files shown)


bpollack:
I think this is a good start, but given that we've already got templating *and* full XDG support, why not just add a word that makes the `.desktop` file and installs it? I think the code to do so would literally be less than the `.desktop` file itself.

catb0t:
very good point, the existing `xdg` vocabulary needs a little bit of extending in order to do this but that's a much better idea.

PR #2043, opened 2018-08-13
fix escape-strings being broken; all tests pass

catb0t:
this was breaking load-all for me so i fixed it

`counts` is just `count` with a curry, and `values-of` isn't needed when the input to `sum` is already an array of numbers :) this passes the vocab's tests

### DIFF
--- a/basis/escape-strings/escape-strings.factor
+++ b/basis/escape-strings/escape-strings.factor
@@ -1,6 +1,6 @@
 ! Copyright (C) 2017 John Benediktsson, Doug Coleman.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: assocs assocs.extras combinators kernel math math.order
+USING: assocs assocs.extras combinators fry kernel math math.order
 math.statistics sequences sequences.extras sets ;
 IN: escape-strings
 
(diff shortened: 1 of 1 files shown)


catb0t:
never mind, now i see the problem is my out-of-date boot image. still i think `basis` vocabs shouldn't rely on words from `extra`

catb0t:
the fact that the tests pass the same with  `[ of ] with map sift` (values-of) and with just `sum 0 =` means `escape-strings` could use better tests however

jonenst:
The tools.coverage tool can be used to find out which paths of the code are not called:
```USING: escape-strings tools.coverage ; "escape-strings" test-coverage```
```
    {
        tag-payload
        { [ [ escape-string ] dip prepend ] [ escape-string ] }
    }
    {
        escape-simplest
        {
            [
                dup { 39 34 13 10 32 } counts {
                    {
                        [
                            dup { 39 13 10 32 } values-of
                            sum 0 =
                        ]
                        [ drop "'" prepend ]
                    }
                    {
                        [ dup 34 of not ]
                        [ drop "\"" "\"" surround ]
                    }
                    [ drop escape-string ]
                } cond
            ]
            [ dup { 39 13 10 32 } values-of sum 0 = ]
            [ drop "'" prepend ]
            [ dup 34 of not ]
            [ drop "\"" "\"" surround ]
            [ drop escape-string ]
        }
    }
```

Here are some tests that bring the coverage to 100%:
```
{ "'foo" } [ "foo" escape-simplest ] unit-test
{ "'\"foo\"" } [ "\"foo\"" escape-simplest ] unit-test

{ "\"'foo\"" } [ "'foo" escape-simplest ] unit-test
{ "\"foo bar\"" } [ "foo bar" escape-simplest ] unit-test

{ "[[\"foo\" \"bar\"]]" } [ "\"foo\" \"bar\"" escape-simplest ] unit-test
{ "[[\"'\"]]" } [ "\"'\"" escape-simplest ] unit-test

{ "tag:[[foo]]" } [ "foo" "tag:" tag-payload ] unit-test
```

PR #2048, opened 2018-08-17
build.sh: correctness / safety and fixups

catb0t:
i found some improvements to make to the build.sh buildscript. I don't know  `CMD` sowell anymore, but maybe I'll look at `build.cmd` sometime too.

* fixed: backtick substitition -> `$(...)` backticks are broken and old, and the form `$(...)` is preferred, safer, fewer bugs etc

* fixed: interpreting arrays as strings (`$*` where there should be `"$@"`)

* fixed: `local x local y` declares the local variables `x`, `local`, and `y` (don't do that)

* correctness: some arguments and substitutions need `"quoted"` to prevent word splitting 

* `CLANG_VERSION >= x` is now globally accessible as `$REQUIRED_CLANG_VERSION` rather than being a magic number hidden away in a function (perhaps we should get it from the Makefile instead??)

* `local -r` like `declare -r` declares `readonly` (const) variables, used where opportune 

* correctness / safety: Factor's `command-line.startup` doesn't have any `-script` argument anymore so i removed it from the invocations of `./$FACTOR_BINARY`, so that it doesn't do something un-expected 

* largest single change: performance improvement (especially if your L1 cache is about 80kb) in c_find_word size

  * make `$CC` compile STDIN
  * return the word size as an exit code instead of printing it
 
`WORD` will always be of the range 16 - 128, so it's a valid POSIX exit code. 

this means we don't have to include a header, and we don't have to call `printf`.

this also reduces the function's number of disk / cache writes, to just 1 per invocation (not including `$DELETE`).

then `check_ret` is used on `$CC` instead of on the executable itself. (`check_ret` on a program that always returns 0 is pretty pointless anyway)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -681,7 +681,7 @@ install_deps_apt() {
 }
 
 install_deps_pacman() {
-    sudo pacman --noconfirm -S gcc clang make rlwrap git wget pango glibc gtk2 gtk3 gtkglext gtk-engines gdk-pixbuf2 libx11 screen tmux
+    sudo pacman --noconfirm -Syu gcc clang make rlwrap git wget pango glibc gtk2 gtk3 gtkglext gtk-engines gdk-pixbuf2 libx11 screen tmux
     check_ret sudo
 }
 
--- a/build.sh
+++ b/build.sh
@@ -27,7 +27,7 @@ test_program_installed() {
 
 # return 1 on found
 test_programs_installed() {
-    installed=0;
+    local installed=0;
     $ECHO -n "Checking for all($*)..."
     for i in "$@" ;
     do
@@ -54,7 +54,7 @@ exit_script() {
 }
 
 ensure_program_installed() {
-    installed=0;
+    local installed=0;
     $ECHO -n "Checking for any($*)..."
     for i in "$@" ;
     do
@@ -243,7 +243,7 @@ check_factor_exists() {
 find_os() {
     if [[ -n $OS ]] ; then return; fi
     $ECHO "Finding OS..."
-    uname_s=$(uname -s)
+    local uname_s=$(uname -s)
     check_ret uname
     case $uname_s in
         CYGWIN_NT-5.2-WOW64) OS=windows;;
@@ -550,7 +550,7 @@ check_url() {
             return 1
         fi
     elif [[ $DOWNLOADER_NAME == 'curl' ]]; then
-        code=$(curl -sL -w "%{http_code}\\n" "$1" -o /dev/null)
+        local code=$(curl -sL -w "%{http_code}\\n" "$1" -o /dev/null)
         if [[ $code -eq 200 ]]; then return 0; else return 1; fi
     else
         $ECHO "error: wget or curl required in check_url"
@@ -562,10 +562,10 @@ check_url() {
 # Otherwise, just use `master`
 set_boot_image_vars() {
     set_current_branch
-    url="http://downloads.factorcode.org/images/${CURRENT_BRANCH}/checksums.txt"
+    local url="http://downloads.factorcode.org/images/${CURRENT_BRANCH}/checksums.txt"
     check_url $url
     if [[ $? -eq 0 ]]; then
-        CHECKSUM_URL="http://downloads.factorcode.org/images/${CURRENT_BRANCH}/checksums.txt"
+        CHECKSUM_URL="$url"
         BOOT_IMAGE_URL="http://downloads.factorcode.org/images/${CURRENT_BRANCH}/${BOOT_IMAGE}"
     else
         CHECKSUM_URL="http://downloads.factorcode.org/images/master/checksums.txt"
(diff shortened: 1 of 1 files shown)


bpollack:
Nice. I want to check this on a couple Linux distros just to make sure nothing exciting happens, but I think this should be a good merge.

build.sh
```diff
@@ -133,16 +135,16 @@ semver_into() {
 }
 
 clang_version_ok() {
-    CLANG_VERSION=`clang --version | head -n1`
+    CLANG_VERSION=$(clang --version | head -n1)
     CLANG_VERSION_RE='^[a-zA-Z0-9 ]* version (.*)$' # 3.3-5
     if [[ $CLANG_VERSION =~ $CLANG_VERSION_RE ]] ; then
         export "CLANG_VERSION=${BASH_REMATCH[1]}"
-        local CLANG_MAJOR local CLANG_MINOR local CLANG_PATCH local CLANG_SPECIAL
+        local CLANG_MAJOR CLANG_MINOR CLANG_PATCH CLANG_SPECIAL
```
bpollack: Bah, great catch. Thanks.

build.sh
```diff
@@ -275,28 +277,27 @@ find_architecture() {
...
 
 c_find_word_size() {
     $ECHO "Finding WORD..."
-    C_WORD=factor-word-size
-    write_test_program
-    $CC -o $C_WORD $C_WORD.c
-    WORD=$(./$C_WORD)
-    check_ret $C_WORD
-    $DELETE -f $C_WORD*
+    C_WORD="factor-word-size"
```
bpollack: Most of the quotes you've added makes sense to me, but this (and one other I'll flag) seem unnecessary. Are you doing this purely for consistency? If so, why not also e.g. inside the conditionals on lines 411-412, for example?

(I'm not saying this wrong; I'm just genuinely unclear.)

build.sh
```diff
@@ -479,7 +480,7 @@ git_fetch_factorcode() {
 }
 
 cd_factor() {
-    cd factor
+    cd "factor"
```
bpollack: Here's the other quote I don't get. It's not wrong as such, but I don't get why we need them here.

build.sh
```diff
@@ -443,12 +444,12 @@ git_clone() {
...
 
 update_script_name() {
-    $ECHO `dirname $0`/_update.sh
+    $ECHO "$(dirname $0)/_update.sh"
 }
 
 update_script() {
-    update_script=`update_script_name`
-    bash_path=`which bash`
+    local -r update_script=$(update_script_name)
```
bpollack: Why `local -r` here, but not for pretty much all other variables? I don't see either `update_script` or `bash_path`  used in multiple places, so I don't think this was causing namespace pollution before. Maybe I'm missing something though?

catb0t: `local` should be used as much as possible for variables that don't need to be global -- someone comes along and adds a new function accidentally using the same global names, and it doesn't work.

but aside from here and `clang_version_ok`, there are no other local variables (or variables that could be made local) in this entire file, everything is globals. it works, and i don't want to change that much.

as for `-r`, the same is true, there are no other possible places to add it.  in `clang_version_ok`  the variables have to be mutable (or the function re-designed). 

these are readonly in this function because there's no reason they should need to change after being set.

build.sh
catb0t: any references to disk filenames I personally think should be in quotes, mostly if these become variables and / or possibly contain spaces in the future. I know `cd factor` is perfectly legal (and i use it every day) but in a build script it feels like a bug waiting to happen.

catb0t: as for the conditionals at line 411, I know `$OS` etc will never contain spaces and putting `linux` in quotes despite `==` working on barewords does seem pointless.

build.sh
catb0t: by the way, `local -r x` is like `const x;` in C: if you only declare a `const` variable without also defining it, you can never define it later because it is `const`. 

so I would like these to be defined exactly once with `local -r` for const correctness' sake, but that would require rewriting the whole function.

catb0t:
now I see what you mean, I'll go add `local` to the function-local variables that need it

build.sh
catb0t: the `-r` is because these shouldn't be changed later in the function, and I added `local` because maybe someone will come along and use those names local in another function, and be surprised their code breaks

catb0t: (it also satisfies the linter Shellcheck who doesn't like `local x=y` for "masking return values", although I did ignore Shellcheck in most places)

build.sh
erg: It's like the airquotes we always do with our fingers when we talk about "Factor".

bpollack:
Okay, I think this is good to merge at this point; any concerns @erg, @mrjbq7 ?

mrjbq7:
Merged, thanks!

PR #2054, opened 2018-08-22
(fix Travis) furnace.auth docs: #2050 load needed vocabs; don't load extra

### LINKED ISSUE #2050: Travis has been failing `./factor -run=zealot.cli-changed-vocabs | paste -s -d " " -` recently
All the recent pull requests have failed the Linux CI with the message:

```
The command "./factor -run=help.lint `./factor -run=zealot.cli-changed-vocabs | paste -s -d " " -`" exited with 1.
```

but without any error from Factor. 

Search this full build log or any other recent Linux one for `exited with 1`: https://api.travis-ci.org/v3/job/417495274/log.txt

catb0t:
closes #2050

### DIFF
--- a/basis/furnace/auth/auth-docs.factor
+++ b/basis/furnace/auth/auth-docs.factor
@@ -1,7 +1,9 @@
 USING: assocs classes help.markup help.syntax kernel
 quotations strings words words.symbol furnace.auth.providers.db
 checksums.sha furnace.auth.basic furnace.auth.features.deactivate-user
-furnace.auth.providers furnace.auth.providers.null math byte-arrays http ;
+furnace.auth.features.edit-profile furnace.auth.features.registration
+furnace.auth.features.recover-password furnace.auth.providers furnace.auth.login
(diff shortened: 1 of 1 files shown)


erg:
Doesn't ``auto-use`` with an empty ``USING:`` list trim these vocabs off again since they aren't directly needed? Let's find a way to change the docs so that clicking on an unloaded vocabulary can load it or load just the docs instead.

mrjbq7:
Why are they needed?

$vocab-link can still link to an unloaded vocab.

erg:
Because of #2050 when zealot runs tests without calling ``load-all`` we get these messages.

erg:
So the fix is to use ``$vocab-link`` instead of ``$link`` I guess?

mrjbq7:
Well if linking to something concrete then the using should add it with auto-use.

catb0t:
It's because of `$subsections "vocab-name"` though, `$link` is not the problem source. Should there be a `$vocab-subsection`?

PR #2072, opened 2018-10-24
fix absolute path on windows

cjallen88:
append-path ignores the first argument if the second is an absolute path, breaks the vocab on windows


mrjbq7:
Merged, thanks!!

PR #2078, opened 2018-11-27
editors.visual-studio-code: search PATH on Windows

melted:
- A \\\\?\ prefixed path returned by 'which' will fail with a Windows
error when used to start the editor, so drop that prefix if present.

### DIFF
--- a/basis/editors/visual-studio-code/visual-studio-code.factor
+++ b/basis/editors/visual-studio-code/visual-studio-code.factor
@@ -40,6 +40,7 @@ M: linux find-visual-studio-code-invocation
 M: windows find-visual-studio-code-invocation
     {
         [ { "Microsoft VS Code" } "code.exe" find-in-applications ]
+        [ "code.cmd" ]
     } 0|| ;
 
 M: visual-studio-code editor-command ( file line -- command )


melted:
Hmm, as one typically would want to run the result of 'which' it would be better to fix up the path in that word.

mrjbq7:
What does a “?” Prefixed path mean on Windows?

bpollack:
It switches the paths into non-length-limited full paths, getting you outside the 260-character limit. These are what paths actually look like internally to the Windows kernel.

mrjbq7:
Hmm, maybe we should handle those more natively then?

erg:
Looks like ``which`` should stop do the stripping of ``\\?\`` instead?

melted:
It's added to paths when normalizing paths on Windows on this line https://github.com/factor/factor/blob/ae74a794e1ef8e76a1fbf6fe07b649662dc17874/basis/io/files/windows/windows.factor#L359

What \\?\ does is described in https://docs.microsoft.com/sv-se/windows/desktop/FileIO/naming-a-file#win32-file-namespaces
Key line: "Many but not all file I/O APIs support "\\?\"; you should look at the reference topic for each API to be sure."

I think it would make sense to not use that prefix for paths. There are other ways around MAX_PATH for those who care about that (Python lately has begin to offer to raise that system-wide in the installer).

melted:
I think I'll do a separate PR with that, and then rebase this on that.

The scope of that change is a little bit bigger than to make VS Code work from the PATH. :-)

melted:
I found a much simpler solution to finding vscode in the PATH. CreateProcess will look in the PATH, but it needs the correct extension.

erg:
Hi @melted 

I had an issue where my VSCode broke because it's now in ``C:\Users\erg\AppData\Local\Programs\Microsoft VS Code`` so I added that to the program files paths. But your suggestion also fixes it, so I added it too.

Thanks!

PR #2087, opened 2018-12-24
db.tuples[{-docs,-tests}]: add reject-tuples

AlexIljin:
Another useful function for the "streaming" ORM API.

### DIFF
--- a/basis/db/tuples/tuples-docs.factor
+++ b/basis/db/tuples/tuples-docs.factor
@@ -139,7 +139,18 @@ HELP: delete-tuples
 { $description "Uses the " { $snippet "tuple" } " as an exemplar object and deletes any objects that have the same slots set. If a slot is not " { $link f } ", then it is used to generate an SQL statement that deletes tuples." }
 { $warning "This word will delete your data." } ;
 
-{ insert-tuple update-tuple update-tuples delete-tuples } related-words
+HELP: reject-tuples
+{ $values
+     { "query/tuple" tuple }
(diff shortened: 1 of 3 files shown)


erg:
Merged!

erg:
This breaks the tests in travis:

https://api.travis-ci.org/v3/job/473729881/log.txt


```
Must Fail: { [ paste new 1 >>n delete-tuples ] }

--> test failed!
```

AlexIljin:
Did this coincide with the SQLite update?

mrjbq7:
Doesn’t your test delete from the dB? Maybe that impacts a subsequent test.

AlexIljin:
Yes, but my test deletes a record that was added immediately before, and in a separate "select_me" table at the end of the unit-test file. That test I created specifically to test my words `each-tuple`, `update-tuples`, and now `reject-tuples`.

The failing test is way up in the first half of the unit-test file, and works with the "PASTE" table.

I know there are side-effects possible between the tests due to insufficient cleanup, but so far I can't see any connection.

AlexIljin:
When I run `"db.tuples" test` it commit 6cae2046b5df5ad975206a0eb09e3d64d56f9897, I get no test failures. But I don't have PostgreSQL installed, only SQLite.

mrjbq7:
Good argument! I’m traveling and haven’t looked yet but makes sense.

erg:
I no longer think it is your patch. It is weird how flakey out tests are on travisci...

Issue #2063, opened 2018-09-02
Long strings in Inspector show only integer values, not characters


nomennescio:
When using the Inspector on long strings (Factor 0.98), the prefix of the string is shown as a string of characters, but the Contents pane shows key-value pairs where the key is the index in the string, and the value just an integer value of the (utf8?) character code, but NOT the value displayed as character.

This makes inspecting a string in the Inspector unnecessary difficult. Please show (also) characters.

mrjbq7:
Do you like the indexed character by character view? Or prefer some other way to inspect the string?

nomennescio:
My internal Unicode look-up table is quite limited, so I would prefer a presentation which (also) shows the corresponding glyph/grapheme for each indexed character.

AlexIljin:
Unicode is tricky to display by-code-point, as some have to be collated to produce the correct glyph, but for 90 of the cases I, too, find it inconvenient to Inspect a `string` as a run of digits.

@nomennescio, have a look at the `prettyprint.config` vocabulary. Setting `string-limit?` to `f` may eliminate the need for Inspecting the `string`s in the first place. See also `length-limit` and `line-limit`.

nomennescio:
@AlexIljin thanks for the tip on prettyprint.config. I would still need to inspect strings occasionally.

AlexIljin:
Having just experimented with the settings mentioned above (by me), I can sadly say that the only useful result comes from those if you want to prettyprint a string with the `.` word. The Inspector still shows the truncated string in its window heading, the Data Stack in the Listener also truncates the sting, as well as the Walker in its Data stack view. Those are all the places that often make me sad when I can't see the end of a long pathname my code is working on.

mrjbq7:
It’s easy enough to change the behavior and have less sadness if we know what changes to make!

AlexIljin:
Counterintuitively, all the three places of sadness I mentioned derive their limitation from `prettyprint.config:margin` - which is supposed to be the wrapping margin, and not an "ellipsis generator". Reading the docs, I would expect the ellipsis to be controlled by the `length-limit` in the various Data Stack representations.

BTW, I don't recommend setting `margin` to `f`, and then calling the Inspector for a `string` : )) That case is, apparently, not handled well.

[...]

PR #2090, opened 2018-12-27
Inspect strings

AlexIljin:
Add the character values output to the Inspector GUI tool as suggested in https://github.com/factor/factor/issues/2063
I hope you don't mind the little documentation update tacked on to this PR.


mrjbq7:
Picked!

PR #2095, opened 2018-12-31
Merge ryu to extra

alex-ilin:
Hello!

  Will you accept a merge like this? It adds a new root commit to the repository.


erg:
Could you please rebase on master and squash as one commit? That would make it easy for me to just cherry-pick it in rather than have merge commits etc. :)

AlexIljin:
Done.

erg:
Yes, adding float16/32 would be good. Or at least adding tests so we're sure it works. (I haven't tried or investigated the issue)

Issue #2103, opened 2019-01-07
heaps bug


jonenst:
Heaps have a bug when you remove entries from the middle of the heap:

```
[let
<min-heap> :> heap
0  dup heap heap-push
5  dup heap heap-push
2  dup heap heap-push
6  dup heap heap-push* :> DEL
3  dup heap heap-push
4  dup heap heap-push
1  dup heap heap-push
DEL heap heap-delete
heap [ drop . ] slurp-heap
]
0
1
3
2 ! BUG
4
5
```
@mrjbq7 looking at the source code, it seems you rewrote them in 3aa14d7570c 

The following patch fixes the bug but I don't think it should go in as is (see questions/suggestions after the patch):

```
diff --git a/basis/heaps/heaps.factor b/basis/heaps/heaps.factor
index b92bfe3..fae8d8f 100644
--- a/basis/heaps/heaps.factor
+++ b/basis/heaps/heaps.factor
@@ -133,7 +133,7 @@ M: heap heap-push*
     ] while drop
 
     tmp over heap data-set-nth
-    heap n rot sift-down ; inline
+    heap 1 rot sift-down ; inline
 
 PRIVATE>
 
```

Reading the source, I had the following questions/suggestions:
   - sift-up and sift-down are exchanged compared to the binary heaps wikipedia article https://en.wikipedia.org/wiki/Binary_heap , rename them ?
  - your sift-up bubbles the element down to a leaf of the tree and then bubbles it back up to the highest possible node, why not stop at the right time during the bubble down (like you did for your sift-down)
   - The bug was that when removing elements, after the bubble down, the bubble up part was limited to where the element was deleted. But in some cases you need to bubble it up even higher. Here, why not bubble down or up according to the value of the parent directly?

let me know what you think

Cheers

mrjbq7:
Hmmmm! Bug!

The intent was to follow similar logic to the comment in siftdown and siftup here, which has some alleged advantages. 

https://github.com/reingart/pypy/blob/master/lib-python/2.7/heapq.py

I’m traveling today, but will look later.

mrjbq7:
Could be a problem only affecting min-heaps?

I noticed there are two versions of sift-up and sift-down... one for min-heaps and one for max-heaps.  Maybe the bug is we only have one version.

```factor
[let
<max-heap> :> heap
0  dup heap heap-push
5  dup heap heap-push
2  dup heap heap-push
6  dup heap heap-push* :> DEL
3  dup heap heap-push
[...]

PR #2105, opened 2019-01-10
Heaps

jonenst:
I implemented what I suggested in #2103 .

Then I saw that the you used the same names as pypy for sift-up and sift-down so I didn't rename them.

And then I saw that the benchmarks went from 0.6s to 0.8s consistently. It turns out bubbling down to the bottom and then back up is a known (but counter-intuitive..) optimization ! It was even explained in the link you provided (I missed it the first time because it's at line 260...) : 
> We *could* break out of the loop as soon as we find a pos where newitem <=
> both its children, but turns out that's not a good idea, ...

https://github.com/reingart/pypy/blob/9d6b4fb5eaabe6d2bbe63e063709d72d86f8b153/lib-python/2.7/heapq.py#L261

So now it's just tests, comments and the bubble up or down in heap-delete.

Also I found this https://stackoverflow.com/a/10163422 where it is suggested to do what I suggested in #2103 (using sift-up with 0 as a limit..) . I was wondering if this was another case of counter intuitive performance optimization, but this time I'd be really surprised if it was. A proper benchmark would be needed to be sure though.

Please review and merge !

### DIFF
--- a/basis/heaps/heaps-tests.factor
+++ b/basis/heaps/heaps-tests.factor
@@ -27,8 +27,24 @@ IN: heaps.tests
 { 0 } [ <max-heap> heap-size ] unit-test
 { 1 } [ <max-heap> t 1 pick heap-push heap-size ] unit-test
 
+DEFER: (assert-heap-invariant)
+
+: ((assert-heap-invariant)) ( parent child heap heap-size -- )
+    pick over < [
+        [ [ heapdata-compare f assert= ] 2keep ] dip
+        (assert-heap-invariant)
+    ] [ 4drop ] if ;
+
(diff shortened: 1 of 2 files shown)


mrjbq7:
Awesome, thanks!  

Merged.

I also made a change in d29052c7b912ce4439c412801d12029ddf6f4c88 that credits ``cpython`` instead of ``pypy`` (since I discovered the comments originated there).  And I modified ``heap-delete`` to use locals so I could understand the logic a little more clearly.  See what you think, maybe.

jonenst:
Great. Also you can simplify the USING: list since combinators are not needed anymore (it was for cond).

PR #2113, opened 2019-02-09
Re-add FreeBSD Support

Silverbeard00:
Using previous commits and with some slight updates I've gotten FreeBSD running again in factor.

The only files this commit touches is _GNUMakeFile_, _system.factor_ and _vm/platform.hpp_,  everything else should be an addition.  I have tested this on a FreeBSD machine with a fresh clone (including a boot image made from Linux) and it seems to build properly.  The command used is _./factor -i=*boot-image-name* -exclude="ui ui.tools"_.  

For some reason this still pulls in the _ui_ and _ui.tools_ libraries which requires continuing twice from the debugger because it tries to load libraries that require Linux.  Otherwise, it creates a working _factor.image_ just fine.

_unix.monitors.freebsd_ still needs to be expanded on,  and the _ui_ libraries will need to be implemented.  I plan on working on this next.  I also took a go at making a version of the _alien.libraries.finder_ for FreeBSD too.

If there are any issues please let me know and I'll work on them.  As far as I can tell this seems stable enough to at least allow people to build for FreeBSD without really effecting the rest of the factor code.


mrjbq7:
Very cool! Now it looks like time to make a FreeBSD VM for me to test with.

erg:
I tried it out, it seems to work!

I made a couple commits you can pull into this PR. ``./build.sh update`` works now, and ``sudo ./build.sh deps-pkg`` installs the command-line factor deps and ripgrep and rlwrap because I like them.

https://github.com/factor/factor/commits/freebsd-work

It's a little ugly with the:
```
IN: scratchpad fatal error for monitor root ``resource:core``: T{ no-method { object freebsd } { generic (monitor) } }
fatal error for monitor root ``resource:basis``: T{ no-method { object freebsd } { generic (monitor) } }
fatal error for monitor root ``resource:extra``: T{ no-method { object freebsd } { generic (monitor) } }
fatal error for monitor root ``resource:work``: T{ no-method { object freebsd } { generic (monitor) } }
```

Maybe fix this and I'll merge the PR?

erg:
Somewhere along the way the handbook started pulling in the UI (maybe it always did?)

To not pull in the UI:
``./factor -i=boot.unix-x86.64.image  -exclude="ui ui.tools handbook"``

Silverbeard00:
@erg Awesome!  I'm excited that build.sh will work now.  I'll get those changes in.

Also yeah those monitor errors are pretty ugly.  I'll work on it and get them sorted out; hopefully this weekend.

Silverbeard00:
Latest commit removes the error messages on startup.

erg:
There's a problem with a case statement when you do ``"io" test``. Hmmm

erg:
Check this out. It bootstraps if you apply these changes.


```
diff --git a/basis/images/loader/gtk/platforms.txt b/basis/images/loader/gtk/platforms.txt
index 9603130824..53926d300d 100644
--- a/basis/images/loader/gtk/platforms.txt
+++ b/basis/images/loader/gtk/platforms.txt
@@ -1,2 +1,3 @@
 linux
 macosx
+bsd
diff --git a/basis/ui/text/pango/platforms.txt b/basis/ui/text/pango/platforms.txt
index a08e1f35eb..a26481a4e1 100644
--- a/basis/ui/text/pango/platforms.txt
+++ b/basis/ui/text/pango/platforms.txt
@@ -1 +1,2 @@
 linux
+bsd
```

Silverbeard00:
There's some other tests failing on "_io_" test that I'll look into along with the case statement one.  I'm getting the same thing.

erg:
Hey, I got the UI running again!
I'll push to my branch soon. Sorry if we duplicated efforts...

erg:
Pushed.

Silverbeard00:
That's great! 

 I noticed in the "_io test_" that my _stat.freebsd.factor_ was incorrect, since it wasn't able to actually load the function from libc.  It was _stat64_ originally when it was supposed to just be _stat_.  So I went through all the stat files to double check their structure declarations (which needed a few small changes based on freebsd source) and now when I run "_io test_" I'm getting a crash for "_Invalid header in base_size_" which seems to come from _vm/slot_visitor.hpp_.  

I also seem to get a crash on building your commit with this for the same error but on the file _ui/event-loop/event-loop.factor_.  

"_critical_error: Invalid header in base_size: 80ed76700_"

I'll push the updated files now.  I'm still trying to wrap my head around it.

erg:
One thing that I'm getting is ``"io" test`` from the UI crashes with ``lstat64`` missing, and it looks like FreeBSD doesn't support ``lstat64`` at all. I am not seeing the invalid header crash yet, maybe if I fix the lstat64 issue I'll see it.

If you check out 7378ba77172b8073ac0707a2d3ada2ffebc74c4c in another directory  you can see what the last working version of everything was before getting removed.

erg:
I see the error now. If you do ``vm-path file-info`` you'll see it breaks with memory corruption because we should be using ``stat64`` or the struct is wrong or something. Once that's fixed more things should work.

Silverbeard00:
[sys/sys/stat.h](https://github.com/freebsd/freebsd/blob/master/sys/sys/stat.h)

I've been going off of this.  I'm going to keep messing around with it.  I think the structure is wrong too.

Silverbeard00:
Ok. I just tested this newest commit (twice) and "_io test_" will now complete.  There are still 11 test errors but at least the factor process isn't dying anymore.

And yeah an automatic FFI for some of these declarations would be amazing.  A CFFI parser could work too.

erg:
Let's merge tonight and I'll put up a test machine next weekend.

erg:
I want to prototype a syntax like this to avoid so many files:

<BSD
    <FREEBSD
        <32BIT
            STRUCT: stat ... ;
        32BIT>
        <64BIT
            STRUCT: stat ... ;
        64BIT>
    FREEBSD>
    <NETBSD
        <32BIT
            STRUCT: stat ... ;
        32BIT>
        <64BIT
            STRUCT: stat ... ;
        64BIT>
    NETBSD>
    <OPENBSD
        <32BIT
            STRUCT: stat ... ;
        32BIT>
        <64BIT
            STRUCT: stat ... ;
        64BIT>
    OPENBSD>
    <DRAGONFLYBSD
        <32BIT
            STRUCT: stat ... ;
        32BIT>
        <64BIT
            STRUCT: stat ... ;
        64BIT>
    DRAGONFLYBSD>
BSD>

Silverbeard00:
That’s a good idea.  I’m going to experiment with a way to just point to header files and get typedefs and structs.  

So maybe in your <64BIT there’d be something like:

<INCLUDE
“Something.h” “Stat.h”
INCLUDE>

And then maybe 

<FUNC
FUNCTION: 
FUNC>

Just throwing that out there for what I’m looking into.  I want to make something to have it be more automatic.  Else it’s going to be a pain in the butt doing the other 3 bsds by hand.  Sorry if formatting is bad,  on my phone.

Crest:
If you want to support FreeBSD 11.x and FreeBSD 12.x there is one more case, because FreeBSD 12 increased the bit width of lots fields in struct stat e.g. ino_t went from 32 Bit to 64 Bit. The libc hides this through symbol versioning, but if I remember correctly Factor doesn't go through the libc.

Silverbeard00:
That's a good point.  I'm gonna mess around with it this weekend.  There's a way or two I can think of to get around that but I'm not sure if it's feasible yet.

catb0t:
The Pony programming language is testing FreeBSD on CirrusCI https://github.com/ponylang/ponyc/pull/3039/commits/4e8936f666347f5f61da8ea097d1060b3f546ce3

Issue #1978, opened 2018-03-17
UI gets corrupted after using Factor normally, cross-platform rendering bug of some kind


erg:
After normal use, the UI gets corrupted in ways that look like this screenshot. If there is a bug open already, please dup this. Also, I am not 100% sure it happens on Linux, but please confirm it does.

![corrupt](https://user-images.githubusercontent.com/27430/37560890-7c36e13a-2a0e-11e8-8044-ce1833a72087.png)

AlexIljin:
Apart from me complaining on the mailing list, I don't think we have a bug registered so far.
Here's a relevant video that I made two years ago. That stuff still happens, but randomly: sometimes after a long session, sometimes almost immediately. In the video you can see that the vertical scroll bar is not present, so the work session was very short (and I only ever open one Listener window).

AlexIljin:
https://www.youtube.com/watch?v=uqdERZRTJ4Y

AlexIljin:
This one time I tried digging into the contents of the Toolbar when this issue happened. I hoped there was some corruption of the cached rendered strings, but those were fine as far as I could tell.

catb0t:
Have never ever seen this on Linux that I can recall.

AlexIljin:
It doesn't only affect the Toolbar. It seems like some string pieces (a character or a repeated substring) gets substituted for a different one - throughout the Listener scrollable contents. I couldn't find the issue in the **cache,** but it sure seems like it's got to be there or about.
The Toolbar may be most affected because it's the oldest contents of the window, or something.

mrjbq7:
More likely OpenGL texture corruption somehow like rebinding a texture ID over one that already existed for a different string. Had a buggy GL function call in minesweeper produce a similar result. But have not seen this corruption in normal use.

AlexIljin:
After updating to the latest Factor build, this issue seems to be happening more often than before, and quicker from the start of the working session. Also, Factor application seems to randomly hang after mundane operations like starting a search in the help browser, running a unit-test, reloading vocabs with `F2` or starting a mouse selection in the listener.

AlexIljin:
[...]

PR #2118, opened 2019-02-26
ui.text: simplify code by removing a type test and using if-empty

AlexIljin:
Remove the PRIVATE section of the vocab. Replace an inline type check with dynamic dispatch.

### DIFF
--- a/basis/ui/text/text.factor
+++ b/basis/ui/text/text.factor
@@ -51,30 +51,21 @@ HOOK: line-metrics font-renderer ( font string -- metrics )
 
 HOOK: string>image font-renderer ( font string -- image loc )
 
-<PRIVATE
+GENERIC: draw-text ( font text -- )
 
-: string-empty? ( obj -- ? )
-    dup selection? [ string>> ] when empty? ;
-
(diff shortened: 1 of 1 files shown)


mrjbq7:
I think you lost the short circuit for empty? Strings.

mrjbq7:
Ah, never mind. I see what you did with if-empty.

AlexIljin:
This patch is not working. For a reason I can't yet grasp, the Listener stops drawing the selection in the input field. Easy to test: type some text and press `C-a`. If the entered text appears selected, everything is OK. With this patch the text doesn't look selected.

AlexIljin:
OK, I understood the reason why the selection wasn't displayed correctly with this patch. The texture cache must get the `selection` object to draw it different from the plain `string`, and with this patch it always got the `string`, so it drew `selection` the same way it remembered to draw the initial string (because the plain text always gets to be displayed before its selected version, and sits in the cache afterwards).

If I disable caching (to prevent #1978 from happening), this patch works as expected.

AlexIljin:
Oops, sorry for reopening.

PR #2124, opened 2019-03-24
prettyprint-docs: fix the obsolete parser code in an example

AlexIljin:
As suggested by @mrjbq7.
It works now.


mrjbq7:
Merged, thanks!

PR #2125, opened 2019-03-29
windows.ole32: add a couple of constants

AlexIljin:


erg:
Pulled!

I put some changes into windows-drag branch

https://github.com/factor/factor/tree/windows-drag

PR #2127, opened 2019-04-08
Extra:  Added Raylib Support

Silverbeard00:
Howdy folks.

I come bearing a humble offering of complete Raylib bindings.  [Raylib Homepage](https://www.raylib.com/)

Raylib is a very clean library along the likes of SDL and Allegro.  I haven't tested out every single function yet but what I have messed with has been surprisingly pleasant.  One very cool thing is that it seems to work perfectly with the REPL.  I've had no issue iteratively making stuff, it seems incredibly stable too.   Deployment works great too (and at only 4mb for the little demo!). 

Maybe the lispers will let us into their game jams now :stuck_out_tongue: 

As usual please let me know if there's anything you'd like cleaned up.  Thanks!

hello-factor


erg:
This is very cool!!

It only works with the mingw download file I guess because the msvc file doesn't have symbols?

Please add a tags.txt to ``raylib.demo`` with ``demos`` as the tag so that ``"demos" run`` picks it up.

When I run the demo, it makes my Factor window tiny (it needs to respect retina settings somehow) and I get a popup with this error:

```
An error occurred while drawing the world T{ world f ~array~ ~array~ f f ~vector~ ~array~....
This world has been deactivated to prevent cascading errors.
gl-error-tuple
function f
code     1281
string   "Invalid value"
```

erg:
tiny-factor

erg:
I added the raylib.dll files to https://downloads.factorcode.org/dlls

Silverbeard00:
tags.txt added.  

I'll boot into Windows later tonight and test it out on there.  (only tested Linux so far)

I'm not sure why it's being funky with the retina settings or why that error message is popping up but I'll look into it the best I can.  Hopefully some of it reproduces on my end in Windows.

Silverbeard00:
I tried the "run" thing on Linux and now I'm getting the same error.  At least it's not platform specific.  Looking into it.

Silverbeard00:
I realized I was doing quite a few things wrong in my demo code.  I've updated this and now `"demos" run` works smoothly with the raylib.demo and no error pops up.  (At least on my machine)

Please let me know if the retina issue still exists.

erg:
On Windows it's still broken on retina, and now the X in the top right of the window doesn't close it. The popup still happens, but I think been it's an ongoing issue for awhile.

Silverbeard00:
Took a little longer than I thought to setup a windows dev environment.  Having problems getting factor to recognize the .DLL in the first place.  I ran it through a DLL checker just to make sure the symbols were exported.

Anyway I'll be looking at the issues soon hopefully.

Silverbeard00:
Tried it out just now with the latest windows x64 binary release of factor,  and added in my raylib folder.  Used the DLL off of factorcode.org.

Everything appears to work.  Window closes when X'd,  arrow keys work for movement,  no errors popping up.  

Capture

This would seem to point to a retina specific issue to me at least.  I  could try poking around and seeing if it's maybe something weird GLFW is doing or some strange interaction between the factor window and raylib,  but it would be kinda hard for me to test if anything works as I don't have a retina display :(.

Also, I'm finding some cool modules which should be similarly easy to write bindings for.  That includes a gui, and physics library that plug into raylib.   It'd be pretty cool to have a nice little "game programming" suite in Factor.  I'll have to find some time to work on them.

I'm open to any thoughts you have on the issue.  Thanks for all the help btw.

Silverbeard00:
I added raygui and rayicon modules.  

I'm not sure if it's the best idea in the world but I also updated raylib.ffi so that when you **USE:** it,  it checks if the dll/so it has access to includes symbols for the raygui and rayicon modules.  If it does then it loads those modules as well.  It seems to work alright from what I've tested.  

raylib1

I spent a brief amount of time testing the gui module out and it seems to work so far.  There's one small bug I'm working on for that module specifically.  (Text for 'widgets' doesn't reload when you destroy the window and make a new one for some reason).

raylib2

I'll have to work on better demos.  It definitely isn't a very factor-y way to program and it'd probably be nice to have some utility functions to make the experience more enjoyable.

Silverbeard00:
I talked with the creator of Raylib and he says there's no automatic pixel scaling for hidpi included.  

I looked into glfw and it seems raylib is not using some of the stuff glfw added for this situation.  I've let him know I'm going to test it out some of these settings tonight and he says he'll add it if it works.  Apparently he didn't know about it at all.  

If I'm lucky getting raylib to respect dpi settings should prevent it from screwing up the factor window.  (Or at least narrow down the issue).

Silverbeard00:
Linux and Apple hidpi should be working in raylib 2.5 on Factor. (Releasing in a few weeks)  Windows is proving to be a tougher beast to conquer but I'm going to pull it off somehow... Also the bindings are now updated to the new version about to be released.  

The raylib leader has graciously [added it](https://github.com/Silverbeard00/raylib-factor) to the [bindings list](https://github.com/raysan5/raylib/blob/master/BINDINGS.md),  hopefully I can get the icon on the raylib website someday.  Maybe we can get Factor some more exposure/use?  I'd be pretty happy with that.  It'd be cool to make some tutorials for learning Factor by making little "games" with raylib.

Now to just get the retina issue fixed on Windows.

Silverbeard00:
The issue to me is still bewildering.  The factor listeners dpi reverts from the first call of init-window.  This makes me think somehow that the opengl context of the listener is being messed with.  I've been looking over the "terrain" example,  the ui.backend code, the world stuff, but none of it seems very helpful in understanding how to make sure the opengl contexts aren't mangling each other or if they even are.  This is with a newer dll that supports proper dpi in raylib too.

It seems like most of the examples are still using the ui as a backend,  yet raylib handles its own context and windowing.  I'm honestly not sure how to proceed now that raylib itself is taken out of the equation.

mrjbq7:
What is the issue exactly that you're trying to fix?  And how would I reproduce?

It's exciting to get these bindings, and would be fun to have supported!

Silverbeard00:
The error should  reproduce on any Windows computer with a HIDPI display.  I have a 4k display that it reproduces on.

This is the listener freshly started...
normal

This is the listener after calling init-window to create a raylib window.  (If you run this on your computer the raylib window may be very small because I'm using an updated dll I made that supports hidpi,  but the raylib dll that erg hosted will work for the error.  It'll be in the new raylib release when it hits.  It does not fix the listener error)

You can see that the listener has unscaled itself down to being very small.
wrong

This is the same thing but running the demo code.  This version exits properly when the window is X'd but leaves the listener small.
wrong3

The error is with init-window somehow causing the Factor listener on hidpi to unscale itself.

mrjbq7:
In the first image, the text looks blurry.  Is it scaled up, instead of native HiDPI resolution?

Silverbeard00:
I have Windows set at 150% scaling which is apparently the recommended one for 4k.  So it seems Windows is scaling up the listener.  It is a tiny bit blurry because of that.

Retina displays apparently make the listener tiny too,  which I can only assume is a scaling issue as well.

AlexIljin:
As I understood the issue, in the first image the Factor window is scaled up to 150% according to the Windows settings, but as soon as `init-window` is called, it reverts to scaling 100%, no longer respecting the Windows settings.

Personally, I use my 4K screen with scaling set to 100%, so this issue wouldn't affect me, but you have to have a large enough screen to roll that way.

mrjbq7:
That seems like some kind of window changed callback that incorrectly gets window bounds (using either wrong API or not adjusting for scaling). Maybe tracing calls to window >>dim would help find it.  

Can try and look through code later.

If the window is resized does it still stay at 100%?

Silverbeard00:
I tried resizing the listener while the game was running and got this...
holycrap

traceback

Pretty freaky.

AlexIljin:
Holy trap!

Silverbeard00:
I wish I could discern more,  I've spent several hours over the past week banging my head against the code trying to figure out why.  For some reason the resizing issue goes away if I run it with `[ main ] with-ui`.  In raylib.demo of course.  The scaling issue remains however.  

I mostly develop in emacs on Linux so this hasn't really been an issue that effects me.  Hopefully it's somehow an easy fix I'm missing,  then I can move on to trying to write some tutorials and maybe an entity system.  I know latex fairly well so a little mini-book isn't out of the question either.

This appears to be a really annoying issue,  but let me know if I can help.   With any lucky, maybe mrjbq7 will see what the problem is immediately.

mrjbq7:
Should we get it ready to merge and then make an issue to track that specific problem?

Silverbeard00:
If you guys do want to do that I'll update it with the latest bindings I'm working with (kept it separate from this pull for some reason), and squash the commits down into 1. 

Also, if this is merged the .dll will have to be updated.  The one currently on the factor website is for Raylib 2.0 instead of 2.5 (which just got released).

I'm not particularly happy with how crummy the bindings are when using the listener.  It's on my checklist to get to fixing support on Windows, but then this project became a much bigger endeavor with the book and other components I want to add.  

Please let me know if there is anything else you'd like to see cleaned/added to the pull and I'll get it done.

mrjbq7:
I love the idea of pushing forward, so lets merge the Raylib 2.5 bindings when you are ready!

And then let's revisit (maybe by opening another issue to track it) the out of memory / HiDPI problems.

We've never really made HiDPI "work" on Windows so I'm not surprised there is a case like that but shouldn't be too hard to solve maybe?

mrjbq7:
Let me know when ready to merge!

Silverbeard00:
Ok, I think that should do it.   Should be good to go! @mrjbq7

mrjbq7:
Merged, thanks!

mrjbq7:
Now, to figure out that Windows HiDPI bug... is that the only problem that you're having?

mrjbq7:
I can't seem to run ``raylib.gui-demo``, I get this error:

```
Cannot resolve C library function
Library: f
Symbol: GuiLock
DlError: none
```

mrjbq7:
On macOS with ``brew install raylib``... version 2.5.

Silverbeard00:
I haven't experienced any issues but I'll keep note.

What you've just experienced is that _raygui_ doesn't come with raylib by default.  You have to compile it in if you want it.  If it does exist in the .dll/.so/etc then I have it setup to check for that and load the extra bindings.

Crap, I missed that _raylib.modules.gui_ is in the USING: of the gui demo.  It shouldn't be,  raylib.ffi will load it if it exists in the compiled library.

Silverbeard00:
Maybe the gui demo shouldn't exist if it requires a special library.  That's my bad.  People can have the option to use the extra bindings if they want to compile it into their raylib.

mrjbq7:
I wonder how I get ``raygui`` with ``brew install raylib`` on macOS...

Silverbeard00:
I know I could generate the .so and .dll to cover linux and windows as far as adding in raygui by default is concerned.  I'm talking with the raylib people to see what the best way to go about it for mac is.  Adding raygui to things surprisingly comes up a lot over there.  I'm surprised they don't just add it in by default.

mrjbq7:
Currently we are building raylib using homebrew on this archive:

https://github.com/raysan5/raylib/archive/2.5.0.tar.gz

using this command:

```
  def install
    system "cmake", ".", "-DSTATIC_RAYLIB=ON",
                         "-DSHARED_RAYLIB=ON",
                         "-DMACOS_FATLIB=OFF",
                         "-DBUILD_EXAMPLES=OFF",
                         "-DBUILD_GAMES=OFF",
                         *std_cmake_args
    system "make", "install"
  end
```

If there is a modification we should make like ``-DRAYGUIYESPLEASE``, let me know, I can make a patch for homebrew...

Silverbeard00:
It's really simple to add it,  I don't know if it's possible to do with brew because I haven't really used it.  

The steps to adding raygui starting with that archive would be.

Add the two files from [Raygui src/](https://github.com/raysan5/raygui/tree/master/src) into the raylib **src/** folder. 

Then in _src/core.c_ you have to add 
`#define RAYGUI_IMPLEMENTATION`
`#include "raygui.h"`

after 

`#define RAYMATH_IMPLEMENTATION `
`#include "raymath.h" `  

And that's it.  I'm not sure if that'd be easy to automate using brew.

Silverbeard00:
https://github.com/Silverbeard00/raylib/archive/2-5-0-gui.zip

I forked it and made the changes for raygui, then trimmed down all the extraneous things out of the folder like the games and examples.

Might be able to point brew at this now and you'll get raygui for free.

PR #2131, opened 2019-04-18
calendar[-docs]: add week-number

AlexIljin:
The new `week-number` word in the `calendar` vocab returns the week number of a `timestamp` according to ISO 8601 (https://en.wikipedia.org/wiki/ISO_week_date#Calculating_the_week_number_of_a_given_date).

Should we maybe use this new word for the "W" format in `strftime`? This could make it compatible with [PHP date("W")](https://www.php.net/manual/en/function.date.php) and possibly other languages. Currently the `formatting` vocab uses its own `<PRIVATE` version of `week-of-year`, which prompted me to write the new `week-number` in the first place, when I needed to do `today week-number .`.

In any case, it seems like the `week-of-year` is something that should be in the `calendar` vocab (along with `day-of-week` and `day-of-year`), not in a private section of the `formatting` vocab.

### DIFF
--- a/basis/calendar/calendar-docs.factor
+++ b/basis/calendar/calendar-docs.factor
@@ -401,6 +401,22 @@ HELP: day-of-year
     }
 } ;
 
+HELP: week-number
+{ $values { "timestamp" timestamp } { "[1,53]" integer } }
+{ $description "Calculates the ISO 8601 week number from 1 to 53 (leap years). See " { $snippet "https://en.wikipedia.org/wiki/ISO_week_date" } }
+{ $examples
+    "Last day of 2018 is already in the first week of 2019."
+    { $example "USING: calendar prettyprint ;"
+               "2018 12 31 <date> week-number ."
+               "1"
+    }
+    "2020 is a leap year with 53 weeks, and January 1st, 2021 is still in week 53 of 2020."
+    { $example "USING: calendar prettyprint ;"
+               "2021 1 1 <date> week-number ."
+               "53"
+    }
+} ;
+
 HELP: sunday
 { $values { "timestamp" timestamp } { "new-timestamp" timestamp } }
 { $description "Returns the Sunday from the current week, which starts on a Sunday." } ;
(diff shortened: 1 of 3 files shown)


mrjbq7:
Do you prefer “week-number” or “week-of-year”? I feel like the latter is what you typically find in strftime documentation. 

As for the others, sure make a PR to move them to calendar vocab if useful for people!

AlexIljin:
The name `week-number` was the name I initially searched for, when I needed the function. It's the intuitive name for me , and easier to type. On the other hand, `week-of-year` is consistent with the names of other similar functions in the `calendar` vocab. Initially I thought that I'll make an alias for myself, but then I found the differences.

Semantically, it feels like the two names return slightly different things. The `week-number` is something related to business/financial calendar, which uses ISO 8601. As you can see from the examples in the documentation patch, a date may be in one Gregorian year, but its `week number` belongs to another (previous or next) year. The `week-of-year` implies that we give it a year number, and it returns the number of the week within that year, possibly 0-based, if the year didn't start on a Monday.

Implementation-wise, the existing `week-of-year` word tries to work for any first day convention, since it accepts the extra parameter: it should work for weeks starting on Monday, Sunday, Saturday, or any other day. This means that it takes an extra parameter, which you can't guess it would from the name alone.

All in all, I think we need both. The `week-of-year` should return the 0-based week number within the year of its `timestamp` parameter, but it should default to weeks starting on Sunday, because that seems to be the API convention of the `calendar` vocab. There should be a variant of it, maybe `week-of-year*`, which would take the starting day number as an extra parameter. The `week-number` should simply return the ISO 8601 value, as it is now, and it should be used in the `strftime` "W" format. The `week-of-year` should be used for the `strftime` "U" format.

This is how I see it today. If you agree with me, I can prepare the additional patches.

mrjbq7:
Good ideas! 

Merged, thanks!

gifti258:
We could also use %V for the ISO week number as per strftime(3) and leave %U and %W as they are.

PR #2132, opened 2019-04-21
extra: cleanup USING: lists

AlexIljin:
A follow-up to commit ba41385f.


mrjbq7:
Ah, right.  Merged, thanks!

PR #2135, opened 2019-04-22
continuations[-docs]: add the finally word

AlexIljin:
This is a useful word from the try..finally..end block of the [Delphi](http://edn.embarcadero.com/article/25665) language (or [JavaScript](https://www.w3schools.com/jsref/jsref_try_catch.asp), for that matter). Basically, it allows you to set a chunk of code to be the finalization part, which always gets executed even if the try block throws an exception.


mrjbq7:
Merged, thanks!

Issue #2119, opened 2019-03-10
Problem with key-up events on Windows


mrjbq7:
Reported on the mailing list.

> On my windows machine (Windows 7), when I hit 1 key which was in a separate numeric pad, 
>Gesture log window showed  :
>
>  T{ key-down { sym “1” } }
>  User input: 1 
>
>per one hit. The other hand, when I hit normal 1 key, it showed :
>
>  T{ key-down { sym “1” } }
>  User input: 1 
>  T{ key-up { sym “1” } }
>
>per one hit.

Information about their configuration:

> Windows 7 Home Premium
> Service Pack 1 32bit 
> 
> Japanese 
> 
> Japanese PS/2 keyboard (109) 
> 
> 32bit Factor  factor-windows-x86-32-0.98

And report of bug on a different system:

> I confirm the same behavior on Windows 10, latest development build of 64-bit Factor.
> It doesn't depend on the numpad or on physical keys.
> 
> For example, when I use the normal English/US layout, the keys for semicolon and comma (";" and ",") generate only the "key-down" events. Pressing "q" and "w" generates both "key-down" and "key-up".
> If I switch to the Programmer Dvorak layout, the physical keys that used to be "q" and "w" in the US layout now will produce the ";" and "," correspondingly, and they will again produce only the "key-down" events in the gesture-logger. Pressing the keys that formerly produced only "key-downs" as ";" and "," will now produce both "key-down" and "key-up", being mapped to "s" and "w".
> 
> It seems that the physical keyboard or the layout is not to blame. For some reason Factor eats some of the key-up events for punctuation characters.
> 
> There are many such characters that I can see treated differently by Factor: #&$^-=\`:;~/'"?@_[]{}(*+
These Factor's treatment of these keys is not dependent on the physical key, but on the logically mapped character it produces under the currently selected layout.
> 
> Numpad keys may be different in that even when Programmer Dvorak produces "a"-"f" or "x", they still don't give a ""key-up" event.

AlexIljin:
Related issue: #1565.

kusumotonorio:
I wrote the code for both this issue and #1565.
https://github.com/factor/factor/issues/1565#issuecomment-479512182 

It ignores caps-lock at key-down and key-up events. And number keys on the numeric pad,  ",", ".", etc provide key-up events.

https://github.com/kusumotonorio/factor/blob/windows-kn-plan2/basis/ui/backend/windows/windows.factor

[...]

Issue #1565, opened 2016-03-16
Inconsistent handling of keyboard shortcuts with caps lock on


nicolas-p:
I have defined a keyboard shortcut like this:

``` factor
environment-gadget "general" f {
    { T{ key-up f f "w" } add-word }
} define-command-map
```

On a Mac, it doesn't matter if caps lock is tuned on or off, pressing W always triggers the action. On Windows, it only works if caps lock is turned off.

I don't know what is the expected behaviour (in this particular case, I prefer the Mac behaviour), but this inconsistency across platforms is strange.

mrjbq7:
So make ``Ctrl-w`` and ``Ctrl-W``  or ``w`` and ``W`` do the same thing?

nicolas-p:
Yes, obviously I can just write:

```factor
environment-gadget "general" f {
    { T{ key-up f f "w" } add-word }
    { T{ key-up f f "W" } add-word }
} define-command-map
```

But maybe it would be worth looking at how Factor handles keyboard gestures to see it there isn't a bug somewhere.

mrjbq7:
@nicolas-p wait, I didn't read the bug report properly -- did you say **in Factor** on the mac that ``W`` and ``w`` both trigger the action?

nicolas-p:
Yes, this is what happens. I guess you can easily check it yourself.

mrjbq7:
Okay, so I don't use caps lock much (I rebind it to Control), but I did some testing and found that apparently when the caps lock key is on the ``charactersIgnoringModifiers`` does not uppercase the key (unlike when the shift key is pressed).

This "fixes" it, but I don't want to apply this change without understanding when and when not to uppercase the characters:

```diff
diff --git a/basis/ui/backend/cocoa/views/views.factor b/basis/ui/backend/cocoa/views/views.factor
index 25b4359..6bea6ca 100644
--- a/basis/ui/backend/cocoa/views/views.factor
+++ b/basis/ui/backend/cocoa/views/views.factor
@@ -68,8 +68,10 @@ CONSTANT: key-codes
     }
 
 : key-code ( event -- string ? )
-    dup -> keyCode key-codes at
-    [ t ] [ -> charactersIgnoringModifiers CF>string f ] ?if ;
+    dup -> keyCode key-codes at [ t ] [
+        [ -> charactersIgnoringModifiers CF>string ]
+        [ -> modifierFlags NSAlphaShiftKeyMask bitand 0 > [ >upper ] when ] bi f
[...]

PR #2136, opened 2019-04-29
Windows Factor's key-down/up event improvements

kusumotonorio:
This change solves some issues with Windows Factor's key-down and key-up events. ( #2119 and #1565 )

1. Ignore caps lock state in key-down/up events
Currently, key-down/up events are affected by the state of caps lock. This behavior is different from Mac Factor.
When you press and release the A key while in the state of caps lock, Mac Factor and Windows Factor each generate an event similar to the following:
Mac Factor:
```
T{ key-down { sym "a" } }
User input: A
T{ key-up { sym "a" } }
```
Windows Factor:
```
T{ key-down { sym "A" } }
User input: A
T{ key-up { sym "A" } }
```
This behavior of the Windows version is not desirable if you want to use the key-down event to execute certain commands by keystrokes. This change ignores the state of caps lock on key-down / up events.

2. Improved key-up events
Windows Factor does not raise key-up events for printable characters other than alphabets and numbers. Also, key-up events are not generated for printable characters on numeric keypads.
The new code solves this problem.

Note
In the past, in Windows Factor, the code that generates the printable character key-down event and the code for character input were closely related. The change has separated them, so we need to test well for character inputs. Also related to this, there are changes in the key-down event that occurs.

For example, when typing ¤ using AltGr on a French keyboard,
```
T{ key-down { mods { C+ A+ } } { sym "¤" } }
```
event was generated, but
```
T{ key-down { mods { C+ A+ } } { sym "$" } }
```
will be. I think this event that generated by the new code is reasonable.

The words called from Windows when the user performs key-down or key-up had become almost the same contents, so they were changed to share the word that takes the word that actually generates the event as an argument.
If you think that this implementation cost is high in terms of execution time, I am ready to rewrite two words with almost the same content.


erg:
I tested this and it seems really good.

Could you please make this change for handling the errors for ``GetKeyboardState`` and ``ToUnicode``? Unless it's better to ignore the error for some reason.

```
-            keyboard-state GetKeyboardState 0 = not [
-                VK_CONTROL VK_CAPITAL [ 0 swap keyboard-state set-nth ] bi@
-                wParam scan-code keyboard-state chars 2 0 ToUnicode dup 0 = not [
-                    1 <=  [
-                        1 chars nth 8 shift 0 chars nth bitor
-                    ] [
-                        3 chars nth 8 shift 2 chars nth bitor ! dead-key
-                    ] if
-                    1string f hWnd send-key-down/up execute( sym action? hWnd -- )
-                ] [ drop ] if
-            ] when
+            keyboard-state GetKeyboardState win32-error<>0
+            VK_CONTROL VK_CAPITAL [ 0 swap keyboard-state set-nth ] bi@
+            wParam scan-code keyboard-state chars 2 0 ToUnicode dup win32-error=0/f
+            1 <=  [
+                1 chars nth 8 shift 0 chars nth bitor
+            ] [
+                3 chars nth 8 shift 2 chars nth bitor ! dead-key
+            ] if
+            1string f hWnd send-key-down/up execute( sym action? hWnd -- )
```

Finally, we like to keep a flat git history, so could you please rebase your branch onto master and squash it into one patch? I could do it but then the patches come from me..

kusumotonorio:
Thank you for your advice!
I changed their error handling. and rebased. But, unfortunately, I'm not sure if I could get rid of the  patch.

By the way, GetKeyboardState is: 
> If the function succeeds, the return value is nonzero. 
If the function fails, the return value is zero. 

The other hand, win32-error<>0 is:
> : win32-error<>0 ( n -- ) zero? [ win32-error ] unless ;

Changed code is running well, but I'm confused. Am I misunderstanding something?

erg:
Merged, thank you!

kusumotonorio:
I'm glad it was solved.

PR #2138, opened 2019-05-04
Gtk key down up

kusumotonorio:
I wrote the cord which fixed key-down/up event issues on gtk backend Factor. It improves:
- events about numeric keys on the keypad with shift, 
- the influence by the state of caps-lock, 
- unnecessary event generation with an individual press of shift, control or alt keys 
- illegal "\0" event generation
on key-down/up events.

### DIFF
--- a/basis/ui/backend/gtk/gtk.factor
+++ b/basis/ui/backend/gtk/gtk.factor
@@ -540,7 +540,7 @@ M: gtk-ui-backend (with-ui)
 M: gtk-ui-backend stop-event-loop
     gtk_main_quit ;
 
-os linux? [
+os { linux freebsd } member? [
     gtk-ui-backend ui-backend set-global
 ] when
 
--- a/basis/ui/backend/gtk/gtk.factor
+++ b/basis/ui/backend/gtk/gtk.factor
@@ -171,19 +171,23 @@ CONSTANT: events-mask
     event win [ key-event>gesture ] [ window ] bi*
     over :> gesture
     gesture sym>> length 1 = [
-        gesture sym>> first LETTER?
-        S+ event event-modifiers member? not and [
-            gesture [ >lower ] change-sym drop
-        ] when
+        { { [ gesture sym>> first LETTER?
+              S+ event event-modifiers member? not and ] 
+            [ gesture [ >lower ] change-sym drop ] }
+          { [ gesture sym>> first letter?
+              S+ event event-modifiers member? and ]
+            [ gesture [ >upper ] change-sym drop ] }
+          [ ]
+        } cond
     ] when
     event keyval>> shift-ignore-keys key? [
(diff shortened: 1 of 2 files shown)


basis/ui/backend/gtk/gtk.factor
```diff
@@ -157,14 +157,39 @@ CONSTANT: events-mask
...
+    ] when
+    event keyval>> shift-ignore-keys key? [
+        gesture [ S+ swap remove ] change-mods drop
+    ] when
+    over sym>> "\0" =
+    event keyval>> exclude-keys-keydown/up key? or [
+        2drop t
+    ] [
+        propagate-key-gesture t
+    ] if ;
```
timor: If it always returns `t`, why not put one `t` behind the `if` instead of at the end of both quotations?

basis/ui/backend/gtk/gtk.factor
```diff
@@ -157,14 +157,39 @@ CONSTANT: events-mask
...
-
+:: key-event>gesture ( event -- key-gesture )
+    event event-modifiers
+    event keyval>> code>sym drop
+    dup integer? not [
+        drop event keyval>> key-sym
+    ] [
+        gdk_keyval_to_unicode 1string f
+    ] if
+    event type>> GDK_KEY_PRESS = [ <key-down> ] [ <key-up> ] if ;
```
timor: You could save a couple of character here if wou wanted to:
```suggestion
    event type>> GDK_KEY_PRESS = <key-down> <key-up> ?;
```

basis/ui/backend/gtk/gtk.factor
kusumotonorio: That's right. I will change it as advice.

basis/ui/backend/gtk/gtk.factor
kusumotonorio: That's good.

mrjbq7:
I’m not sure this suggestion works.

basis/ui/backend/gtk/gtk.factor
timor: > I’m not sure this suggestion works.

You are right, it does not. Sorry about that...

kusumotonorio: Oh! Surely, it seems better to leave it as "if". Then, I will leave this as it is.

kusumotonorio:
Thank you, Jhon. It certainly seems so.

mrjbq7:
We should review this and merge if it's working!  @erg

erg:
@kusumotonorio I tried this and it returns `t` which causes keyboard strokes not to show as characters. Changing it to `f` causes it to work but I don't see the features mentioned above, like shift+numpad doesn't type anything.

Could you please rebase this onto master and see if you can fix it? I will merge sooner next time. Thanks!

PR #2139, opened 2019-05-06
[Suggestion] Don't jump to USING: section after updating

timor:
The auto-update-usings functionality is really useful.  However, I think it is better to preserve point in the buffer when this is called.  The proposed `USING:` change is already displayed in an extra buffer for confirmation.

From my experience, staying at point was the correct behavior in virtually all cases, which involved at least two mark jumps.

To reproduce:

1. Edit factor file spanning more than one screen, start factor listener with `M-x run-factor`
2. Add code at the bottom which uses a word from a vocab which is not yet in the `USING:` list
3. Save the file, and `M-x fuel-run-file` or `M-x fuel-update-usings`, confirm suggested changes
4.a Current behavior: point jumps to the update `USINGS:` line
4.b Suggested behavior: point remains at last edit position


mrjbq7:
I don't use FUEL, so I'm not sure.  Who are our emacs users?  @bjourne @bpollack @erg ?

AlexIljin:
I use the last stable Spacemacs (over a year old at this point), but I very rarely develop Factor code in it.

AlexIljin:
If there were clear instructions (including the key combination to update the USING: list) how to reproduce the "wrong" behavior, I could test the fix. I do have FUEL installed and working.

timor:
I updated the issue description with the steps to reproduce.

In the factor layer of the current Spacemacs development branch, which I am using, the default bindings are `, c c` for `fuel-run-file` and `, r u` for `fuel-update-usings`, respectively.

AlexIljin:
I was able to reproduce the problematic behavior, the key combo for me was `C-c C-e u`.
I will try to apply the patch and see how it goes.

mrjbq7:
For what it’s worth, I’m not sure we should change a part of the file the user isn’t looking at, and probably why it jumps there for visual verification of the change being made?

timor:
Right before the command finishes, the user is presented with a an extra window like this:

```
Current vocabulary list:

  kernel sequences


Correct vocabulary list:

  math sequences


Type 'y' to update your USING: to the new one.
```

The differences are highlighted with color.  It would make sense to jump there if the user really needs to verify that the change did not introduce any unwanted behavior.  However, I have not once encountered this command messing up something it shouldn't.  Even if it did, it would be just one undo away.

mrjbq7:
Oh, ok! That’s a good idea then!

misc/fuel/fuel-debug.el
```diff
@@ -308,20 +308,21 @@ the debugger."
...
-        (kill-region begin end))
-    (re-search-forward "^IN: " nil t)
-    (beginning-of-line)
-    (open-line 2)
-    (insert "USING: "))
-  (let ((start (point))
-        (tokens (append uses '(";"))))
-    (insert (mapconcat 'substring-no-properties tokens " "))
-    (fill-region start (point) nil)))
+  (save-excursion
```
AlexIljin: Basically, what you did is just adding this `save-excursion`, right?

AlexIljin: Yes, when I hide the whitespace changes I see it was like that.

AlexIljin:
I have tested it, and it works. No side-effects are to be expected from a change like this.
I would accept this PR.

misc/fuel/fuel-debug.el
timor: That's correct.

mrjbq7:
Merged, thanks @timor !

AlexIljin:
Congrats on your first contribution, @timor!
👍

PR #2146, opened 2019-06-11
FUEL: make `fuel-refactor-inline-word` work with hyphenated words

timor:
The emacs command `kill-word`, used by `fuel-factor-inline-word` does not
delete the complete word if that contains hyphens.  Using `kill-sexp` instead
exhibits correct behavior.

Reproduction:

1. Create some definitions:

```
: foo-bar ( x -- y )
    1 + dup . ;

: baz ( x -- y )
    foo-bar ;  ! place point on 'foo-bar' for step 3
```
2. Compile file
3. Place point  on `foo-bar` inside definition of `baz`
4. run `M-x fuel-refactor-inline-word`, notice that the `-bar` part is not removed before the definition of `foo-bar` is inserted


mrjbq7:
Merged, thanks!

PR #2148, opened 2019-07-05
UNICODE TEXT SEGMENTATION 12.0.0

kusumotonorio:
I modified unicode.factor and breaks.factor in order to make Factor work in accordance with the latest UAX No.29. I expect that it can handle characters that include almost any emoji properly.
It is for branch unicode-12.1.0.


erg:
Great work! I merged this and master back into the factorcode ``unicode-12.1.0`` branch.

Issue #2154, opened 2019-07-12
Type declarations/inference for mutable local variables


timor:
Is it possible to either declare the type of a mutable local variable in a lexical closure like this:

```
[let 0 :> x! [ x + dup x! ] ]
```

As far as I can tell, inference also does not work because the frontend compiler translates the variable access into `set-slot` and `slot` primitive calls, and loses any value information during propagation.

mrjbq7:
You can always use a "low level" type declaration.  Here's the word with the generic "plus" operation:

```factor
IN: scratchpad [let 0 :> x! [ x + dup x! ] ] optimized.
[ { 0 } >R "COMPLEX SHUFFLE" 2 slot + dup R> 2 set-slot ]
```

But if you ``declare`` that the stack before the "plus" is ``{ fixnum fixnum }`` then it will use the more specialized ``fixnum+`` word:

```
[let 0 :> x! [ x { fixnum fixnum } declare + dup x! ] ] optimized.
[ { 0 } >R "COMPLEX SHUFFLE" 2 slot fixnum+ dup R> 2 set-slot ]
```

Right now, I don't think we can annotate local type declarations like you are doing.  Typed mutable locals would be a nice addition, and maybe not too hard?

timor:
Yes, that certainly is a workaround, but you would have to put it before all assignments and after all uses of the variable.

I guess what I am really looking for is correct type inference where possible. This should always be possible if no read-before-write access occurs.  For write-before-read, a declaration is still needed though, if I am not mistaken.

How would local type declaration syntax look like?  Would it be implemented in a macro-like manner, just expanding into `{ ... } declare` forms for all variable uses? 

I think it would be nicer if it was possible to declare it once at definition, and have the compiler be able to propagate SSA value info to the read sites automatically.  This could be done in the propagation pass if the `"outptus"` word-property quotation of `set-slot` stores information about what is stored in a box, and the corresponding `"outputs"` property of `slot` could then maybe look at all possible input types and compute some kind of union type on that.

As far as I can tell, the whole thing really is just an optimization for one specific case though: Lexical closures with mutable closed-over variables.  If I am not mistaken, this is the only thing that can not be implemented without local variables.

timor:
The `local-value` and `set-local-value` words in `locals.backend` are  currently defined `inline`.  If they weren't, their word-properties for "outputs" and "constraints" could be used for storing info about the input and output values.

[...]

PR #2155, opened 2019-07-15
[RFC] Locals value info propagation

timor:
Adds an additional frontend pass to calculate possible value types at reader sites
of (mutable) local variables.

This is a proof-of-concept for tackling #2154. Probably fails in every case except the one I tested it with.  No idea if this is the correct place to implement this.  Should probably be integrated into the propagation pass.

Basic idea:
1. Record input value information for `set-local-value` calls, Assuming `local-value` calls return the initialization value
2. Traverse the tree again, annotating `local-value` outputs with the union type information of all `set-local-value` calls and the initial value

Has not been tested on branches, but keeping an assoc-stack, this should also work.  Currently every call to `set-local-value` is considered for all reader sites, which is probably wrong.

No idea what this does to overall compilation times.

### DIFF
--- a/basis/compiler/tree/locals/locals-tests.factor
+++ b/basis/compiler/tree/locals/locals-tests.factor
@@ -1,0 +1,21 @@
+USING: compiler.tree.builder compiler.tree.locals compiler.tree.optimizer kernel
+kernel.private locals math namespaces prettyprint ;
+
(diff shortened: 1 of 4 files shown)


timor:
This was very rudimentary.  I did some work on a better approach and will create a PR for that when ready.

PR #2156, opened 2019-07-20
 Inline input using input methods in Mac Factor

kusumotonorio:
This PR provides inline input using input methods in Mac Factor.

To support this internationalization of text input, I have made changes to the editor gadget and the propagation mechanism for key gestures.

Input using Input methods are underlined until their input is confirmed. The editor gadgets changes for this feature are intended for future use with Windows Factor and GTK Factor.
During inline input using an input method, the key gesture is exclusively passed to it.

This PR supports the input of emojis using input methods. Therefore, this PR is for branch unicode-12.1.0.

The Mac Factor's text input protocol changes from NSTextInput to NSTextInputClient.

### DIFF
--- a/basis/documents/documents-docs.factor
+++ b/basis/documents/documents-docs.factor
@@ -59,6 +59,12 @@ HELP: set-doc-range
 { $errors "Throws an error if " { $snippet "from" } " or " { $snippet "to" } " is out of bounds." }
 { $side-effects "document" } ;
 
+HELP: set-doc-range*
+{ $values { "string" string } { "from" "a pair of integers" } { "to" "a pair of integers" } { "document" document } }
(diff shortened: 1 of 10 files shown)


erg:
I noticed a couple of issues.

1) I hold down "e" and try to make an ``ē`` it actually inserts ``eē``, two characters instead of one
2) The input window appears in the bottom left corner instead of over the character as in Chrome. 

This is great though! Merged.

kusumotonorio:
Thank you for trying this on, @erg.
Those problems must be fixed.
In particular, the second one bothers me because the Factor does not provide inline input. I may have made a mistake in creating this PR.
I am sorry to trouble you, but could you show me the image of the input window? Let me think about the cause.

kusumotonorio:
Oh, I understand what you say now. The second issue was solved.

kusumotonorio:
I could fix the first issue too. 
I will be testing for problems with those changes for a while.

mrjbq7:
Whats the plan with that branch?  Are we getting it to a place to merge to ``master`` soon?

Maybe after last few bugs fixed?

mrjbq7:
@erg

erg:
The collation algorithm is wrong. There is a ``ducet`` table which has some entries that are right and some that are wrong. The official docs are hard to follow and I haven't found a good implementation to compare against.

PR #2157, opened 2019-07-26
Better support to NSTextInputClient

kusumotonorio:
This PR is for unicode-12.1.0, mainly intended to improve the NSTextInputClient support for input other than input methods. For example, if you hold down "e", an input window appears at the "e" position, and you can select its variation.

Unfortunately, there is one big problem with this PR.
Since the code can't get the correct replacementRange argument in  setMarkedText:selectedRange:replacementRange, I use a guess instead.
A subcontract word is like this:
```
:: update-marked-text ( gadget str selectedRange replacementRange -- )
    replacementRange location>>  NSNotFound = not     ! [ 
    replacementRange length>> NSNotFound = not and [  ! erase this line
        gadget editor-caret first
        dup gadget editor-line
        [ 
            replacementRange length>> ! location>>
            >codepoint-index
            2array gadget set-caret
        ] [
            replacementRange length>> 1 + ! [ location>> ] [ length>> ] bi +
            >codepoint-index
            2array gadget set-mark
        ] 2bi
        gadget earlier-caret/mark dup
        gadget preedit-start<<
        0 1 2array v+ gadget preedit-end<<
    ] when
...
```

I don't know why this parameter of this method is the only problem. In my research, the "length" of this NSRange parameter contains a value that would otherwise be in "location", and "location" contains a random value. It is possible that the wrong value is passed because my code does not return the correct value to the system, but it is very strange that a value can be placed in other member variables of the structure.

I created a PR to get you to check the feature, but you might not want to merge it until the problem is solved.

I wrote a test method that accepts the same number of arguments of the same type, and checked the parameters for the values of the arguments received, but the method didn't have this problem. For now, the cause and solution are still unknown.

### DIFF
--- a/basis/ui/backend/cocoa/views/views.factor
+++ b/basis/ui/backend/cocoa/views/views.factor
@@ -193,7 +193,7 @@ IMPORT: NSAttributedString
 
 :: >codepoint-index ( str utf16-index -- codepoint-index )
     0 utf16-index 2 * str utf16n encode subseq utf16n decode length ;
-
+    
 :: >utf16-index ( str codepoint-index -- utf16-index )
     0 codepoint-index str subseq utf16n encode length 2 / >integer ;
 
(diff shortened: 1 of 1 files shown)


basis/ui/backend/cocoa/views/views.factor
```diff
@@ -445,38 +463,58 @@ PRIVATE>
...
-                gadget support-input-methods? [
-                    gadget preedit? [
-                        gadget [ remove-preedit-text ] [ remove-preedit-info ] bi
-                    ] when
-                    str gadget user-input* drop                    
-                    f gadget preedit-selection-mode?<<
-                ] [ 
-                    str window user-input
-                ] if
+                gadget [
```
mrjbq7: previously we didn't have this wrapped in ``gadget [ ... ] when``, does it need that now?

basis/ui/backend/cocoa/views/views.factor
```diff
@@ -445,38 +463,58 @@ PRIVATE>
...
             ] when
         ] ;
-    
+
     METHOD: char hasMarkedText [
             self window :> window
             window [
                 window world-focus :> gadget
-                gadget preedit? [ 1 ] [ 0 ] if
+                gadget [
```
mrjbq7: Same comment here...

basis/ui/backend/cocoa/views/views.factor
kusumotonorio: As you say, this new extra guard should not be needed. However, when I was working on better NSTextInputClient support for this code, the word that was supposed to be dispatched to a gadget was dispatched to "f", Mac Factor downed.

The problem may be a transitional problem because the backend has been changed since it will not be replicated, but I was afraid it would be to someone. So, I added that guard to all methods of NSTextInputClient protocol.

erg:
Merged as is!

This fixed both of the issues for me. (input completion area is now at cursor, and only one character is entered instead of two)

PR #2158, opened 2019-07-27
Improved background rendering when selecting text including emojis in Mac Factor

kusumotonorio:
This PR is for unicode-12.1.0.
It corrects an issue where text containing characters represented as surrogate pairs in UTF-16 cannot display the selection correctly in Mac Factor.


erg:
Missed this one. Merged into unicode branch. Thanks!!

PR #2159, opened 2019-08-02
Decodes when surrogate pairs are entered etc.

kusumotonorio:
This PR is for branch unicode-12.1.0.

It improves the display of character selection, fixes problems with selecting text to the left with the mouse, and decode when surrogate pairs are entered on Windows Factor.
 ```line-offset>x``` and ```x>line-offset``` in ```uniscribe.factor``` are defined as non-private.

### DIFF
--- a/basis/ui/backend/windows/windows.factor
+++ b/basis/ui/backend/windows/windows.factor
@@ -9,7 +9,9 @@ threads ui ui.backend ui.clipboards ui.event-loop ui.gadgets
 ui.gadgets.private ui.gadgets.worlds ui.gestures ui.pixel-formats
 ui.private windows.dwmapi windows.errors windows.gdi32
 windows.kernel32 windows.messages windows.offscreen windows.opengl32
-windows.types windows.user32 assocs.extras byte-arrays ;
+windows.types windows.user32 assocs.extras byte-arrays
+io.encodings.string ;
+FROM: unicode => upper-surrogate? under-surrogate? ;
 SPECIALIZED-ARRAY: POINT
 QUALIFIED-WITH: alien.c-types c
 IN: ui.backend.windows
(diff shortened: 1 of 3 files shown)


basis/ui/backend/windows/windows.factor
```diff
@@ -347,11 +349,26 @@ CONSTANT: exclude-keys-wm-char
...
-       ] unless
+        ctrl? alt? xor [ ! enable AltGr combination inputs
+            wParam {
+                { [ dup upper-surrogate? ] [
+                      upper-surrogate-wm-char set-global ]
+                }
+                { [ dup under-surrogate? ] [
+                      drop
+                      upper-surrogate-wm-char get-global [
+                          1string wParam 1string 2array "" join
```
mrjbq7: Hmm, I noticed we don't have ``2string``, but you could do this instead:

```factor
wParam "" 2sequence
```

erg: maybe we should add {2,3,4}string

erg:
I pulled this into a new patch because there were a ton of merge conflicts because it's not rebased off master. Thank you for the fix!

What is left to do? We should start thinking about another release with the input method fixes.

mrjbq7:
Not necessarily a new release until we think about what else should be 0.99 (new parser? Etc).

But we should get some new nightly builds soon? When are you merging to master?

kusumotonorio:
The mrjbq7's notice is good. Please use it.

erg:
It's been merged to master since a few days ago.

basis/ui/backend/windows/windows.factor
erg: Sorry. Fixed!

PR #2161, opened 2019-08-07
Compute and use output type information when compiling words

### LINKED ISSUE #417: Function output type inference
I was talking to Joe and he pointed out that function output types aren't used for compiler type inference.

For example this:

``` factor
IN: scratchpad : foo ( -- x ) 2 ;
IN: scratchpad : bar ( -- y ) foo sqrt ;
IN: scratchpad \ foo optimized.
[ 2 ]
IN: scratchpad \ bar optimized.
[ foo sqrt ]
```

Should probably infer as `[ foo M\ fixnum sqrt ]`, no?

timor:
This is intended to fix #417.

When the frontend compiler arrives at the `#return` node during the propagation pass, that node's input value information corresponds to the return type of the word outputs.

The first commit basically does the following:
- When compiling, expose the word which is being compiled in a global variable (Couldn't find a better way to determine where to put the type information)
- At the end of the propagation pass, store output type information in the word's property `output-infos`
- During the propagation pass, when determining the initial types of a `#call` node's outputs, check whether the word has information in it's `output-infos` property, and use these

This breaks some unit test in the dead-code elimination pass, because when a word returns a literal, it's output type information also records that literal, and then when turning an optimized tree back into a quotation, the literal is inserted instead of the original word call.  One could argue that this is actually correct behavior of an optimizer, and that the unit tests should be fixed.  One could also argue that a word should not be inlined unless marked inline, regardless of what it returns.  The latter issue is addressed by the second commit, which removes the information whether a return value is a literal.

I don't know what the correct way is to test this, but I reran all tests in the `compiler.tree` test as well as reload a lot of vocabs, and it did not seem to break anything.

As I am mostly interested in words storing their compiled output types, I did not (and don't know how to) extensively test these changes regarding generated code speed/size or compile time differences.

### DIFF
--- a/basis/compiler/tree/propagation/simple/simple.factor
+++ b/basis/compiler/tree/propagation/simple/simple.factor
@@ -1,12 +1,11 @@
 ! Copyright (C) 2008, 2010 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors alien.c-types arrays assocs classes classes.algebra
-classes.algebra.private classes.maybe classes.tuple.private
-combinators combinators.short-circuit compiler.tree
+classes.tuple.private combinators combinators.short-circuit compiler.tree
 compiler.tree.propagation.constraints compiler.tree.propagation.info
 compiler.tree.propagation.inlining compiler.tree.propagation.nodes
-compiler.tree.propagation.slots continuations fry kernel make
-sequences sets stack-checker.dependencies words ;
+compiler.tree.propagation.output-infos compiler.tree.propagation.slots
+continuations fry kernel sequences stack-checker.dependencies words ;
 IN: compiler.tree.propagation.simple
 
 M: #introduce propagate-before
@@ -49,12 +48,6 @@ M: #declare propagate-before
             swap predicate-constraints assume
         ] [ 2drop ] if
     ] if* ;
-
-ERROR: invalid-outputs #call infos ;
-
-: check-outputs ( #call infos -- infos )
-    over out-d>> over [ length ] bi@ =
-    [ nip ] [ invalid-outputs ] if ;
 
 : call-outputs-quot ( #call word -- infos )
     dupd
@@ -111,12 +104,22 @@ ERROR: invalid-outputs #call infos ;
     "default-output-classes" word-prop
     [ class-infos ] [ out-d>> length object-info <repetition> ] ?if ;
 
+
+! Force literal? to f to prevent non-specified inlining.  Class/Interval info is
+! still propagated.
+: copy-output-infos ( #call word -- infos )
+    2dup check-copied-output-infos
+    [ nip "output-infos" word-prop clone [ f >>literal? ] map ]
+    [ default-output-value-infos ] if
+    ;
+
 : output-value-infos ( #call word -- infos )
     {
         { [ dup \ <tuple-boa> eq? ] [ drop propagate-<tuple-boa> ] }
         { [ dup sequence-constructor? ] [ propagate-sequence-constructor ] }
         { [ dup predicate? ] [ propagate-predicate ] }
         { [ dup "outputs" word-prop ] [ call-outputs-quot ] }
+        { [ dup "output-infos" word-prop ] [ copy-output-infos ] }
         [ default-output-value-infos ]
     } cond ;
 
--- a/basis/compiler/compiler.factor
+++ b/basis/compiler/compiler.factor
@@ -4,7 +4,10 @@ USING: accessors assocs classes classes.algebra combinators
 combinators.short-circuit compiler.cfg compiler.cfg.builder
 compiler.cfg.builder.alien compiler.cfg.finalization
 compiler.cfg.optimizer compiler.codegen compiler.crossref
-compiler.errors compiler.tree.builder compiler.tree.optimizer
+compiler.errors
+compiler.tree
+compiler.tree.propagation.output-infos
+compiler.tree.builder compiler.tree.optimizer
 compiler.units compiler.utilities continuations definitions fry
 generic generic.single io kernel macros make namespaces
 sequences sets stack-checker.dependencies stack-checker.errors
@@ -28,6 +31,7 @@ SYMBOL: compiled
     H{ } clone dependencies namespaces:set
     H{ } clone generic-dependencies namespaces:set
     HS{ } clone conditional-dependencies namespaces:set
+    dup word-being-compiled namespaces:set
     clear-compiler-error ;
 
 GENERIC: no-compile? ( word -- ? )
@@ -111,7 +115,9 @@ M: word combinator? inline? ;
     ! If the word contains breakpoints, don't optimize it, since
     ! the walker does not support this.
     dup optimize? [
-        [ [ build-tree ] [ deoptimize ] recover optimize-tree ] keep
+        [ [ build-tree ] [ deoptimize ] recover optimize-tree
+          [ update-output-infos ] keep
+        ] keep
         contains-breakpoints? [ nip deoptimize* ] [ drop ] if
     ] [ deoptimize* ] if ;
 
--- a/basis/compiler/tree/propagation/known-words/known-words.factor
+++ b/basis/compiler/tree/propagation/known-words/known-words.factor
@@ -1,15 +1,15 @@
 ! Copyright (C) 2008, 2010 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors alien alien.accessors alien.data.private arrays
-assocs byte-arrays byte-vectors classes classes.algebra classes.tuple
+USING: accessors alien alien.accessors alien.data.private arrays assocs
+byte-arrays byte-vectors classes classes.algebra classes.tuple
 classes.tuple.private combinators compiler.tree.comparisons
 compiler.tree.propagation.constraints compiler.tree.propagation.info
-compiler.tree.propagation.simple compiler.tree.propagation.slots fry
-generic.math hashtables kernel kernel.private layouts locals math
-math.floats.private math.functions math.integers.private
-math.intervals math.libm math.parser math.partial-dispatch
-math.private namespaces sbufs sequences slots.private splitting
-stack-checker.dependencies strings strings.private vectors words ;
+compiler.tree.propagation.simple compiler.tree.propagation.slots continuations
+continuations.private fry generic.math hashtables kernel kernel.private layouts
+locals math math.floats.private math.functions math.integers.private
+math.intervals math.libm math.parser math.partial-dispatch math.private
+namespaces sbufs sequences slots.private splitting stack-checker.dependencies
+strings strings.private vectors words ;
 FROM: alien.c-types => (signed-interval) (unsigned-interval) ;
 IN: compiler.tree.propagation.known-words
 
@@ -391,3 +391,9 @@ generic-comparison-ops [
 \ tag [
     drop fixnum 0 num-types get [a,b) <class/interval-info>
 ] "outputs" set-word-prop
+
+
+! If info from `current-continuation` and the dummies is used, `ifcc` breaks
+\ current-continuation [ object-info ] "outputs" set-word-prop
+\ dummy-1 [ object-info ] "outputs" set-word-prop
+\ dummy-2 [ drop object-info ] "outputs" set-word-prop
--- a/basis/compiler/tree/propagation/output-infos/output-infos.factor
+++ b/basis/compiler/tree/propagation/output-infos/output-infos.factor
@@ -1,0 +1,72 @@
+USING: accessors combinators combinators.short-circuit compiler.tree
+compiler.tree.propagation.info continuations io kernel math namespaces sequences
+words ;
+IN: compiler.tree.propagation.output-infos
+
+FROM: namespaces => set ;
+
+! * Using Stored Output Infos During Propagation
+ERROR: invalid-outputs #call infos ;
+
+: check-outputs ( #call infos -- infos )
+    over out-d>> over [ length ] bi@ =
+    [ nip ] [ invalid-outputs ] if ;
+
+: null-infos? ( infos -- ? )
+    [ null-info = ] any? ;
+
+: literal-infos? ( infos -- ? )
+    [ literal?>> ] any? ;
+
+: check-consistent-effects ( #call infos -- ? )
+    [ check-outputs ] [
+        dup invalid-outputs? [
+            2drop
+            "FIXME: Inconsistent stack effect output for compiled word: " write
+            word>> name>> print
+            f
+        ] [ rethrow ] if
+    ] recover
+    ;
+
+! This is quite verbose, mainly for catching things which indicate other problems.
+: check-copied-output-infos ( #call word -- ? )
+    "output-infos" word-prop
+    {
+        { [ 2dup check-consistent-effects not ] [ 2drop f ] }
+        { [ [ word>> name>> ] dip dup null-infos? ]
+          [ drop "WARNING: ignoring NULL infos from " prepend write nl f ] }
+        ! { [ dup literal-infos? ]
+        !   [ drop "WARNING: ignoring LITERAL infos from " prepend write nl f ] }
+        [ 2drop t ]
+    } cond
+    ;
+
+! * Storing Inferred Output Infos
+
+: should-store-output-infos? ( nodes -- infos/f )
+    [
+        { [ length 2 > ] [ but-last last ] } 1&&
+        #terminate? not
+    ]
+    [ last dup #return?
+      [ "STRANGE: last node not return, not storing outputs" print ] unless
+      node-input-infos
+    ]
+    bi and ;
+
+ERROR: duplicate-output-infos word infos ;
+
+: update-output-infos ( nodes -- )
+    word-being-compiled get [
+        ! dup "output-infos" word-prop [ duplicate-output-infos ] when*
+        swap should-store-output-infos?
+        [
+            [ drop ] [
+                [ clone f >>literal? ] map ! This line prevents literal propagation
+                "output-infos" set-word-prop
+            ] if-empty
+        ] [
+            drop
+        ] if*
+    ] [ drop ] if* ;
--- a/basis/compiler/tree/tree.factor
+++ b/basis/compiler/tree/tree.factor
@@ -4,6 +4,8 @@ USING: accessors arrays assocs kernel namespaces sequences
 stack-checker.visitor vectors ;
 IN: compiler.tree
 
+SYMBOL: word-being-compiled
+
 TUPLE: node < identity-tuple ;
 
 TUPLE: #introduce < node out-d ;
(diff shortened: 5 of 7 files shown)


mrjbq7:
Very cool! I’ll take a look and experiment some with this over the weekend! I appreciate you pushing on these ideas, unsure about details without reviewing the commits but thank you!!

erg:
It's not valid to replace a word with its literal result, if the word's body has side effects.

timor:
> It's not valid to replace a word with its literal result, if the word's body has side effects.

That is of course true.  As far as I can tell, the words in question aren't being inlined per se, only the "decompilation" of the optimized word returns the literal instead of the word call.

timor:
Since I only tested this interactively by reloading the compiler.tree vocab before, I added a commit which fixed a dependency issue.  Now loading the compiler vocabs succeeds, but the build fails (locally, as well as here for Travis) later during compilation, in a backend pass:

```
...
Checking for multimedia extensions... SSE4.2 detected
Loading resource:basis/cpu/x86/sse/sse.factor
Building with POPCNT support

Compiling....................................................................................................................................................................................Loading v
ocab:bootstrap/bootstrap-error.factor                                                                                                                                                                
(U) Quotation: [ c-to-factor => ]
    Word: c-to-factor
(U) Quotation: [
        2dup [
            ~quotation~ dip
            dup terminated?>> ~quotation~ ~quotation~ if
        ] 2dip => rot [ 2drop ] [ wrong-values ] if
    ]
    Word: call-effect
(U) Quotation: [
        [ ~quotation~ dip dip ] dip => dup terminated?>>
        [ 2drop f ]
        [ dup in>> length swap out>> length check-datastack ] if
    ]
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(U) Quotation: [
        nano-count f parser-quiet? set-global
        default-image-name "output-image" set-global
        default-components "include" set-global
        "" "exclude" set-global
        strip-encodings (command-line) parse-command-line
        os windows? [ "windows" require ] when "staging" get
        [ "stage2: deployment mode" print ]
        [ "debugger" require "listener" require ] if
        load-components => nano-count over -
        core-bootstrap-time set-global
        run-bootstrap-init nano-count swap -
        bootstrap-time set-global print-report "staging" get
        [ "vocab:bootstrap/finish-staging.factor" run-file ]
        [ "vocab:bootstrap/finish-bootstrap.factor" run-file ] if
        f error set-global f original-error set-global
        f error-continuation set-global
        "output-image" get save-image-and-exit
    ]
(U) Quotation: [ [ nip call ] 3keep => [ 1 + ] 2dip (each-integer) ]
(U) Quotation: [ swap [ 2dip ] dip => ]
    Word: 3dip
(U) Quotation: [ swap [ dip ] dip => ]
    Word: 2dip
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [
        2dup [
            ~quotation~ dip
            dup terminated?>> ~quotation~ ~quotation~ if
        ] 2dip => rot [ 2drop ] [ wrong-values ] if
    ]
    Word: call-effect
(U) Quotation: [
        [ ~quotation~ dip dip ] dip => dup terminated?>>
        [ 2drop f ]
        [ dup in>> length swap out>> length check-datastack ] if
    ]
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(U) Quotation: [
        dup source-loaded?>> [ dup load-source ] unless
        => dup docs-loaded?>> [ dup load-docs ] unless drop
    ]
(U) Quotation: [ [ compose ~quotation~ curry recover ] [ drop ] 2bi => call ]
    Word: cleanup
(U) Quotation: [ [ 2keep ] dip => call ]
    Word: 2bi
(U) Quotation: [ swap [ dip ] dip => ]
    Word: 2dip
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(U) Quotation: [
        +parsing+ >>source-loaded? dup vocab-source-path
        [ parse-file ] [ ~quotation~ ] if*
        [ +parsing+ >>source-loaded? ] dip
        [ % ] [ ( -- ) call-effect ] if-bootstrapping
        => +done+ >>source-loaded? load-conditional-requires
    ]
(U) Quotation: [
        2dup [
            ~quotation~ dip
            dup terminated?>> ~quotation~ ~quotation~ if
        ] 2dip => rot [ 2drop ] [ wrong-values ] if
    ]
    Word: call-effect
(U) Quotation: [
        [ ~quotation~ dip dip ] dip => dup terminated?>>
        [ 2drop f ]
        [ dup in>> length swap out>> length check-datastack ] if
    ]
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [
        nl "Compiling..." write flush {
            not
            ?
            2over
            array?
            hashtable?
            vector?
            tuple?
            sbuf?
            tombstone?
            curried?
            composed?
            callable?
            quotation?
            curry
            compose
            uncurry
            array-nth
            set-array-nth
            wrap
            probe
            (get-namestack)
            layout-of
        } compile-unoptimized "." write flush
        { bitand bitor bitxor bitnot } compile-unoptimized "." write
        flush { + * 2/ < <= > >= shift } compile-unoptimized
        "." write flush
        { new-sequence nth push pop last flip } compile-unoptimized
        "." write flush
        { hashcode* = equal? assoc-stack (assoc-stack) get set }
        compile-unoptimized "." write flush {
            member-eq?
            split
            harvest
            sift
            cut
            cut-slice
            subseq-start
            index
            clone
            set-at
            reverse
            push-all
            class-of
            number>string
            string>number
            like
            clone-like
        } compile-unoptimized "." write flush {
            lines
            prefix
            suffix
            unclip
            new-assoc
            assoc-union!
            word-prop
            set-word-prop
            1array
            2array
            3array
            ?nth
        } compile-unoptimized "." write flush os windows? [
            "GetLastError" "windows.kernel32" lookup-word
            "FormatMessageW" "windows.kernel32" lookup-word 2array
            compile-unoptimized
        ] when os unix? [
            "(dlerror)" "alien.libraries.unix" lookup-word
            1array compile-unoptimized
        ] when { malloc calloc free memcpy } compile-unoptimized
        "." write flush loaded-vocab-names
        [ vocab-words compile-unoptimized "." write flush ] each
        => " done" print flush "alien.syntax" require
        "io.streams.byte-array.fast" require
    ]
(U) Quotation: [ [ nip call ] 3keep => [ 1 + ] 2dip (each-integer) ]
(U) Quotation: [ swap [ 2dip ] dip => ]
    Word: 3dip
(U) Quotation: [ swap [ dip ] dip => ]
    Word: 2dip
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [ vocab-words compile-unoptimized => "." write flush ]
(U) Quotation: [ recompile => t f modify-code-heap ]
    Word: compile
(U) Quotation: [
        H{ } clone compiled
        [ ~quotation~ filter ~quotation~ each compiled get >alist ]
        with-variable => "--- compile done" compiler-message
    ]
    Method: M\ optimizing-compiler recompile
(U) Quotation: [ swap >n call => ndrop ]
    Word: with-variables
(U) Quotation: [
        [ compile? ] filter
        [ compile-word yield-hook get ( -- ) call-effect ] each
        => compiled get >alist
    ]
(U) Quotation: [ [ nip call ] 3keep => [ 1 + ] 2dip (each-integer) ]
(U) Quotation: [ swap [ 2dip ] dip => ]
    Word: 3dip
(U) Quotation: [ swap [ dip ] dip => ]
    Word: 2dip
(U) Quotation: [ swap [ call ] dip => ]
    Word: dip
(U) Quotation: [ compile-word => yield-hook get ( -- ) call-effect ]
(O) Word: compile-word
(O) Word: backend
(O) Word: build-cfg
(O) Word: (build-cfg)
(O) Word: finalize-stack-shuffling
(O) Word: visit-block
(O) Word: visit-edge
(O) Word: insert-peeks
(O) Word: local-loc>global
(O) Word: +
(U) Quotation: [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]
T{ no-math-method f f f + }
```

Unfortunately I don't know a lot about the backend compiler, or how to debug this interactively.

timor:
I tracked this down a bit, and the error happens during loading of `bootstrap.compiler` vocab, specifically here: https://github.com/factor/factor/blob/41db42a5714819ab271ba04fae547a226fd7e47d/basis/bootstrap/compiler/compiler.factor#L123

Some added debug statements revealed that the offending vocab, which is being recompiled there, is `compiler.cfg.stacks.padding`, but I don't know (yet) which during which word's recompilation.

Running `"bootstrap.compiler" reload` from the listener did not trigger the error, though.

Is there a way to get a data stack backtrace along with the call stack on error during build?

timor:
@mrjbq7 Did you have a chance to take a look at this?

basis/compiler/tree/propagation/simple/simple.factor
```diff
@@ -6,9 +6,11 @@ combinators combinators.short-circuit compiler.tree
 compiler.tree.propagation.constraints compiler.tree.propagation.info
 compiler.tree.propagation.inlining compiler.tree.propagation.nodes
 compiler.tree.propagation.slots continuations fry kernel make
-sequences sets stack-checker.dependencies words ;
+namespaces sequences sets stack-checker.dependencies words ;
 IN: compiler.tree.propagation.simple
 
+SYMBOL: word-being-compiled
```
mrjbq7: Is this information not available anywhere else?

basis/compiler/tree/propagation/simple/simple.factor
```diff
@@ -62,6 +64,10 @@ ERROR: invalid-outputs #call infos ;
     [ "outputs" word-prop ] bi*
     with-datastack check-outputs ;
 
+! Force literal? to f to prevent non-specified inlining.
+: copy-output-infos ( #call word -- infos )
```
mrjbq7: Can you give an example of the inlining this is preventing?

basis/compiler/tree/propagation/simple/simple.factor
timor: I also asked myself the same question and tried to trace in which context a word is being compiled.  It seems that definition is separated from compilation quite well.  

I set a breakpoint on `compile-word` and inspected the dynamic variables, but could not find any which would indicate which word is being compiled.

The alternative would be to change the interface of `frontend` to either take the word currently defined or return the return value info as additional stack output.

Even if I didn't miss something there, I don't know which version would be better.  Having the word being compiled in context during compilation and not touching the interface of `frontend` made the most sense to me.  Open to suggestions on this...

basis/compiler/tree/propagation/simple/simple.factor
timor: From `compiler.tree.propagation.tests`:
```factor
TUPLE: foo bar ;
...
: that-thing ( -- class ) foo ;

{ f } [ [ that-thing new ] { new } inlined? ] unit-test
```
will fail, returning `{ t }`, because `[ that-thing new ] optimized.` shows
```factor
[
    f
    { foo 1 1 tuple 236985587512 foo 266632840776318362 }
    <tuple-boa>
]

```
although `that-thing` is not declared inline.  The other unit test failures are along the lines of that.

Other strange things also happen, e.g.:
```
IN: scratchpad : test ( x -- y class ) 1 + foo ;
IN: scratchpad \ test optimized.
[ 1 + \ foo ]
IN: scratchpad [ test ] optimized.
[ test ]
IN: scratchpad [ test new ] optimized.
[
    test ( 11235067 11235068 -- 11235067 ) f
    { foo 1 1 tuple 236985587512 foo 266632840776318362 }
    <tuple-boa>
]
```
The word seems to work as intended though, so this might just be a limitation somewhere in the tree debugger.

Since the intention of this PR is "only" to provide inferred output types, not preventing this seemed to result in much more aggressive optimization than expected.

mrjbq7: Okay, have to look into this more.  I'm not sure if this is an issue or not...

mrjbq7:
Note for later: we have a lot of places where we ``inline`` for type propagation, but maybe with this feature it's not required anymore.  That would be awesome.  Need to do a code audit.

basis/compiler/tree/propagation/simple/simple.factor
mrjbq7: Wait, your example seems busted.

```factor
TUPLE: foo bar ;
...
: that-thing ( -- class ) foo ;
```

All that ``that-thing`` should do is push the ``foo`` symbol:

```factor
IN: scratchpad \ that-thing optimized.
[ \ foo ]
```

So, why is it making a ``foo`` instance now?

mrjbq7: Oh never-mind, you're doing ``that-thing new``... hmmm.

timor:
I was finally able to narrow it down some more.  It happens during bootstrap compilation
of the word `visit-insn` from `"compiler.stacks.padding"`.  After trying for almost forever to reproduce it, I noticed that `visit-insn` actually never has been compiled.

I wrote some helper words for detecting uncompiled word which probably should have been compiled.  I will open a separate issue for this.

So it seems like the changes from this PR don't cause the compilation errors, but just trigger compilation of this word.

The behavior below is already present in 0.98!

```factor
IN: scratchpad USING: compiler.cfg.stacks.padding compiler ; \ visit-insn
IN: scratchpad dup [ compile? . ] [ no-compile? . ] [ word-optimized? . ] tri
t
f
f
IN: scratchpad compile-word
! -> sequence-is-immutable
```

So it looks like this word should be compiled, but hasn't been. Trying the above results in a
a sequence-is-immutable error.  Note that this is not the same error as the one in the backtrace above.  Depending on what is compiled in which order, different errors occur, but always during compilation of a small set of words somewhere in `"compiler.cfg.stacks.*"`.

So I guess regarding this PR there are two ways to go: 

1. Work around the problematic words by preventing compilation.
2. Fix the underlying issue

Number 2 will probably require someone with more knowledge about the backend compiler than me...

timor:
I finally got bootstrapping to work!

timor:
Ok, now most of the propagation seems to work like intended.  Now the tests fail because continuations are broken.  I will try to provide a good repro, but I probably could use some help with the continuation details. 

It seems like there already is quite some magic at work to keep the optimizer from messing with them.

timor:
Well, that went better than expected!  Continuations seem to be working correctly again.
Although I don't really understand the restrictions of the stack layout regarding continuations, fixing some word outputs ensured that `ifcc` and friends don't look any different than before.

mrjbq7:
Did you have any info on whether this change reduces the image sizes or improves any benchmark performance?

timor:
> Did you have any info on whether this change reduces the image sizes or improves any benchmark performance?

No not yet. I don't know whether that makes sense before all the tests pass? The Travis builds here fail on some test which does seem to hang.

I can run stuff from the running image, though. What would be a good way to test the size and speed changes?

mrjbq7:
I meant that you got bootstrap to work, so these image sizes:

```factor
IN: scratchpad make-my-image
```

```
$ /bin/ls -l boot*.image
-rw-r--r--  1 jbenedik  staff  2431948 Aug  6 07:16 boot.unix-x86.32.image
-rw-r--r--  1 jbenedik  staff  4125496 Aug  6 07:16 boot.unix-x86.64.image
-rw-r--r--  1 jbenedik  staff  2432172 Aug  6 07:16 boot.windows-x86.32.image
-rw-r--r--  1 jbenedik  staff  4125720 Aug  6 07:16 boot.windows-x86.64.image
```

And the resulting factor image after ``./factor -i=boot.unix-x86.64.image`` or whatever:

```
$ /bin/ls -al factor.image
-rw-r--r--  1 jbenedik  staff  106964424 Aug 19 19:36 factor.image
```

And regarding the tests, many of them don't work without setting up stuff external dependencies like postgresql or memcached or whatever.

But you can run the benchmarks (they take a little while to all load and run):

```
$ ./factor -run=benchmark
```

And show the results before and after?

timor:
Making the images is unfortunately not so easy for me, since on NixOS I have to stick to a certain version and some patches to get factor to bootstrap in the first place... 

If you have a Ubuntu machine or some other officially supported platform, could you perhaps check the influence? I'll see if I get some results from the benchmarks...

timor:
While testing some sizes of compiled words of vocabs, I noticed that they don't seem to be independent from load/compile order.  Is that plausible?

timor:
> Can you file bugs or make PRs for nixos compatibility? I don't want people to have to be on old versions.

@erg
I can do that.  Do you prefer one issue containing all NixOS problems, or one per problem?

erg:
@timor Just one patch/branch (many commits is fine)

timor:
> Note for later: we have a lot of places where we `inline` for type propagation (not because we need to for example with combinators or inline quotations), but maybe with this feature it's not required anymore. That would be awesome. Need to do a code audit.

It looks like this feature only helps if the input type is known and constant, e.g. methods, typed words.  If not constrained completely, sometimes the output value info is at least narrowed down a it. If you rely on inlining for inferring output types based on inputs, this feature can probably not provide that.

timor:
> @timor Just one patch/branch (many commits is fine)
> 
> edit: i thought you had patches, i guess just use your judgement. checkboxes are nice too

see #2180

timor:
I just stumbled over this again:
> While testing some sizes of compiled words of vocabs, I noticed that they don't seem to be independent from load/compile order. Is that plausible?

I guess for inlining and such, compile order is not important.  For this to work as intended however, 
a caller word would have to be compiled after it's callees have been propagated,

Alternatively, It could be done like splicing in finalization, where a callee is recompiled and then cached.  I guess for words with mutual dependencies this still creates a problem, as output info calculation would still depend on compilation order?

timor:
I moved most of the code into a separate file and cleaned this up a bit.  I used these modifications for a while now, and they seem to work, but most of the possible optimizations do not happen yet.

As suspected, the biggest issue regarding the usage of inferred output value info is compile order.  I am working towards a different PR for that, which would use this one as a basis.

mrjbq7:
We have spent time trying to make the compiler have a consistent order and contributions towards that goal are welcomed!

Do you have a test case showing an example where it compiles different order?

timor:
> We have spent time trying to make the compiler have a consistent order and contributions towards that goal are welcomed! Do you have a test case showing an example where it compiles different order?

The compilation order is consistent.  Sorry if I was unclear about that.  The problem is that the compilation order inside one compilation unit does not take into account any dependencies.  Thus, most of the time, words which could make use of inferred output infos are compiled before their dependencies at the moment.

timor:
The approach in this PR had less impact than I hoped.  For the most part, that is because output type inference which is independent of the input types at a call site only leads to a relatively small amount of non-trivial optimizations.

I think what I really intended with this was some kind of interprocedural optimization.  I have some code which performs (possibly expensive) propagation of non-inline word calls and caches the input->output types during compilation, which I haven't got around to creating a PR yet, though.

Closing this since I don't use this approach anymore and didn't test it accordingly.

Issue #2160, opened 2019-08-04
Quirks of input methods on different platforms


erg:
I tried some input methods with other languages on macOS and Windows. I'm not sure these matter much, just something to note.

- Arabic input
- [ ] double-quote puts blinking cursor on the wrong side and typing another character teleports it leftwards
- [ ] ``⌘ + k`` and ``⌘ + K`` don't clear the output because it's not ``k`` it's ``ن``

- Chinese (pinyin)
- [ ] double-quote character ``“ ”`` alternates between code points 8220 and 8221, brackets are 12304 and 12305, etc ``【 】``

kusumotonorio:
The last of these reports seems to be improved if it is rewritten as follows. (Since 'a' is also ignored, it is not a reaction with Chrome at all.)
```
ui/backend/cocoa/views/views.factor:


    METHOD: char validateUserInterfaceItem: id event
    [
        self window :> window
        window [
            window world-focus :> gadget
            gadget [
                gadget preedit? [ 0 ] [
                    window event -> action utf8 alien>string validate-action
                    [ >c-bool ] [ drop self event SUPER-> validateUserInterfaceItem: ] if
                ] if
            ] [ 0 ] if
        ] [ 0 ] if
    ] ;
```

kusumotonorio:
Oops, when you press  ```⌘ + a```, the 'a' input seems to be valid.

mrjbq7:
@kusumotonorio there seems to be a regression with the completion popup on macos.

For example, try and do ``fixnum-f<TAB>`` (where that's the tab button).

Then pick the first thing on the list (``fixnum-fast``) and then hit the space button.

It looks like this broken:

```factor
IN: scratchpad fixnum-f fixnum-fast
```

It should look like this:

```factor
IN: scratchpad fixnum-fast
```

kusumotonorio:
I also tried it. Well, it's strange. It seems to work well with TAB and RET, but what is the difference?

As for mischief, I made the following changes. (" " → "c")
```
completion-popup H{
    { T{ key-down f f "TAB" } [ table>> row-action ] }
    { T{ key-down f f "c" } [ table>> row-action ] }   !  " " -> "c" 
} set-gestures
```
It works fine with "c".

kusumotonorio:
By the way, why is it possible to choose complementation in RET?
Is it defined somewhere else?

kusumotonorio:
I found a solution to the problem @mrjbq7  pointed out.
[...]

PR #2163, opened 2019-08-10
NSTextInputClient Improvements

kusumotonorio:
It solves the following two problems pointed out in https://github.com/factor/factor/issues/2160.
- Improves the behavior when text is selected from the completion pop-up and confirmed with the space key
- Improves the operation where the key combination is passed on to Factor while you are typing using the Input Method.

It also moves the definition location of methods in the NSTextInputClient protocol and changes the file that defines ```preedit?```.


mrjbq7:
Merged, thanks!

PR #2164, opened 2019-08-15
Support declaring intervals as well as classes for the optimizing compiler

timor:
A `{ class } declare` call will now also refine the value info if the `class`
word has a valid interval in the "declared-interval" property.

This allows certain optimizations during value propagation passes in the
frontend compiler.

For easier usage, a new vocabulary `math.intervals.predicates` is supplied.

### DIFF
--- a/basis/compiler/tree/propagation/info/info.factor
+++ b/basis/compiler/tree/propagation/info/info.factor
@@ -2,6 +2,7 @@
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors arrays assocs byte-arrays classes
 classes.algebra classes.singleton classes.tuple
+classes.intersection classes.union
 classes.tuple.private combinators combinators.short-circuit
 compiler.tree.propagation.copy compiler.utilities kernel layouts math
 math.intervals namespaces sequences sequences.private strings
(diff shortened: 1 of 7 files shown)


mrjbq7:
Can you provide a test case showing a difference in behavior before and after this change?  With an example of some kind of declare?

I'm not sure if this is what you intended, but something like this:

```factor
IN: scratchpad [ { fixnum } declare 30 < ] optimized.
[ 30 fixnum< ]

IN: scratchpad PREDICATE: small-fixnum < fixnum 25 < ;

IN: scratchpad [ { small-fixnum } declare 30 < ] optimized.
[ 30 fixnum< ]
```

timor:
Do you mean here or as a unit test?

For the given example, this should illustrate it:
```factor
IN: scratchpad PREDICATE: foo-fixnum < fixnum abs 25 < ;

IN: scratchpad [ { foo-fixnum } declare 30 < ] optimized.
[ 30 fixnum< ]

IN: scratchpad INTERVAL-PREDICATE: small-fixnum < fixnum -25 25 (a,b) ;

IN: scratchpad [ { small-fixnum } declare 30 < ] optimized.
[ drop \ t ]
```

This allows optimizations like these:
```factor
IN: scratchpad [ { foo-fixnum foo-fixnum } declare + ] optimized.
[ fixnum+ ]

IN: scratchpad [ { small-fixnum small-fixnum } declare + ] optimized.
[ fixnum+fast ]

IN: scratchpad INTERVAL-PREDICATE: positive < fixnum 0 (a,inf] ;
IN: scratchpad INTERVAL-PREDICATE: negative < fixnum 0 [-inf,a) ;
IN: scratchpad [ { positive negative } declare number= ] optimized.
[ 2drop f ]

IN: scratchpad [ { positive negative } declare [ abs ] bi@ ] optimized.
[ >R R> -1 fixnum* ]
```

This is also quite useful in combination with #2161:

```factor
IN: scratchpad TYPED: foo ( x: small-fixnum x: small-fixnum -- y ) + ;
IN: scratchpad \ foo "output-infos" word-prop .
V{
    T{ value-info-state
        { class fixnum }
        { interval
            T{ interval { from { -50 t } } { to { 50 t } } }
        }
    }
}
```

mrjbq7:
Thanks @timor !

mrjbq7:
I reverted this temporarily, I want to understand something first before merging again.

Issue #2166, opened 2019-08-20
Defining NSNotFound to Fit the System


kusumotonorio:
I defined the constant ```NSNotFound``` in views.factor. https://github.com/factor/factor/blob/master/basis/ui/backend/cocoa/views/views.factor

```
CONSTANT: NSNotFound 9223372036854775807 inline
```

However, this value is correct in a 64 bit environment, but should be different in a 32 bit environment. I don't have a Mac with a 32 bit OS, so I can't really try it out, but probably because of this, the NSTextInputClient implementation will not work correctly there.

How should this code be written to specify different values in 64 bit and 32 bit environments?

mrjbq7:
You could do something like this:

```factor
USING: layouts literals ;

CONSTANT: NSNotFound $[ 32bit? 0x7fffffff 0x7fffffffffffffff ? ]
```

mrjbq7:
I should note, this is probably not as good because it's obscure but I think this would work on 32-bit and 64-bit:

```factor
CONSTANT: NSNotFound $[ cell-bits 1 - 2^ 1 - ]
```

kusumotonorio:
Wonderful. The second code is cool. But, as you say, it's better to have the first one that can clearly tell the difference between 32 bit and 64 bit.

kusumotonorio:
Since ```NSNotFound``` is a general Cocoa constant, it is better to put the new definition in ```basis/cocoa/types/types.factor```. How do you like it?

kusumotonorio:
Please close this issue.

erg:
Thank you!

PR #2178, opened 2019-09-08
NSNotFound for both 32 bit and 64 bit environments

kusumotonorio:
PR for the issue https://github.com/factor/factor/issues/2166.


mrjbq7:
Merged, thanks!

PR #2179, opened 2019-09-09
math.intervals: Consistent handling of special intervals

timor:
Make both `empty-interval` and `full-interval` singletons, use generic functions
and methods where they are special-cased.

All words which work with interval points should also now work with the special
intervals.

### DIFF
--- a/basis/compiler/tree/propagation/info/info.factor
+++ b/basis/compiler/tree/propagation/info/info.factor
@@ -83,7 +83,7 @@ UNION: fixed-length array byte-array string ;
 : empty-set? ( info -- ? )
     {
         [ class>> null-class? ]
-        [ [ interval>> empty-interval eq? ] [ class>> real class<= ] bi and ]
+        [ [ interval>> empty-interval? ] [ class>> real class<= ] bi and ]
     } 1|| ;
 
 ! Hardcoding classes is kind of a hack.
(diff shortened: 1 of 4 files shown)


mrjbq7:
I wonder if “special interval” is needed.  Seems like you could just define both methods instead of using the union to define one.

timor:
Do you mean this?
```factor
GENERIC: special-interval? ( int -- ? )
M: empty-interval special-interval? drop t ;
M: full-interval special-interval? drop t ;
M: interval spceial-interval? drop f ;
```
Or getting rid of `special-interval?` completely?

In the former case, having the union is more concise, and dispatch code after finalization should be identical anyway.

In the latter case:
There is on call site in `compiler.tree.propagation.info` which is not generic, and the binary operations in `math.intervals` itself would have to be multi-methods, I think.

It seems easier to not touch that abstraction.

mrjbq7:
I see, the latter but didn’t realize that leaked out and was used. No worries!

mrjbq7:
Merged, thanks!

mrjbq7:
This change caused a test case to fail, which was the result of exposing a compiler issue doing type conversion to bignum...

Instead of reverting, I made a minimal change to your code in e256a4ba18fbb7916db2a22ead55cbcbf7cf920d that widened the result slightly but prevented the compiler bug and opened #2203 to look into it later.

Issue #2180, opened 2019-09-09
NixOS compatibility


timor:
There are several problems right now with Factor under NixOS.  This issue is intended to keep track of the details.

NixOS has some properties which prevent Factor working as intended out-of-the box:

- Non-FHS filesystem, no `/usr/lib`, `/bin/ls`, `/usr/share/` etc.  This means that all references to absolute paths like `/usr/share/timezone` must be patched during build phase.
- Strict sandboxed build environment
- All installed software and build time pre-requisites resides in a read-only store
- "Clone our current stable/master branch and develop your software" does not work well and is discouraged.  Development environments are assumed to be stable and reproducible.

The following issues need to be addressed to get it working without any major hack-arounds:

- [ ] Trying to use `ldconfig` when loading libraries fails. (#2171)
- [ ] `work` resource path can not be in the same location as factor installation (#364, #936, [workaround](https://github.com/timor/factor/commit/812d79ccfa1d702afc12561aa8c05a9b4a158620))
- [ ] Non-deterministic binaries (#2114).  For building Factor or deploying an Application once, this is not a problem.  However, packages included upstream are usually built twice to ensure deterministic build products.
- [ ] Staging images can not be cached when using Factor deployment for building Nix packages, because all build inputs must be deterministic.  This is an inherent limitation unless a stable set of images is distributed with/can be compiled from a given Factor version at installation time.

To run the core tests after building:
- [ ] Some unit tests assume FHS layout ([workaround](https://gist.github.com/timor/57ed3e6f4a6b5c03320e6c7013641cbc))
- [ ] Some unit tests assume internet connection
- [ ] Some unit tests use local network services.  Note that these kinds of tests are possible under NixOS, but run as OS-level tests, with dedicated VM setup.

If the development version is intended to be distributed:
- [ ] Official release is quite dated (#2177).  NixOS upstream prefers tagged releases with increasing version numbers.  E.g. to package the development version, it would be useful to have "0.99-pre1,0.99-pre2,..." releases which can be referenced when fetching the sources (even if they don't have any semantic meaning)

mrjbq7:
Great list of stuff to fix!  I need to get a NixOS installation I can test with, or have the intersection between Factor users and NixOS users contribute some patches.

By the way, I'm not sure what you mean by "official release is quite dated".  Our last release of version 0.98 was July 2018 and 320 commits ago?  Is that "quite dated"?

timor:
[...]

PR #2195, opened 2019-10-15
Create shell.nix for building/running on NixOS

kwshi:
NixOS uses different conventions for locating/loading dynamic libraries, which, by default, makes Factor unable to build or run (errors like `./factor: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory`).  

One workaround is to open a _Nix shell_ (conceptually similar to something like `pipenv shell`), which sets up some environment variables and builds a "conventional" environment that puts dynamic libraries in places where programs _like Factor_ normally expect to find them.  The `shell.nix` file added here defines the configuration for the Nix shell (dependencies, etc.) and makes it possible for a NixOS user to build and run Factor from source by doing
```bash
nix-shell
./build.sh update
./factor
```

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,20 @@
+{ pkgs ? import <nixpkgs> {} }:
+(pkgs.buildFHSUserEnv {
+  name = "factor";
+  targetPkgs = pkgs: (with pkgs; [
+    # for running factor
+    gtk2-x11
+    glib
+    gdk_pixbuf
+    gnome2.pango
+    cairo
+    gnome2.gtkglext
+
+    # for building factor
+    clang
+    git
+    curl
+    binutils
+  ]);
+  runScript = "bash";
+}).env


mrjbq7:
Merged, thanks!

timor:
@kwshi Thanks for that file!

This is good for interactive use, but there is, however, a caveat to this:  The `buildFHSUserEnv` mechanism is something like NixOS' last resort, emulating an FHS system during the existence of the environment, used e.g. for closed-source binary-only software, but creating problems when used in conjunction with other packages.

In fact, only for dealing with the dynamic libraries, the same behavior should be possible just by creating a shell environment which sets `LD_LIBRARY_PATH` during it's lifetime (using `makeLibraryPath`) without resorting to `buildFHSUserEnv`. 

This would be somewhat similar to what I had in mind in https://github.com/factor/factor/issues/2180#issuecomment-539515834 for the solution to 2.iii

PR #2196, opened 2019-10-16
Simplify nix shell environment

timor:
`buildFHSUserEnv` uses `chroot` to create a kind of sandbox environment.  This
change uses a simpler mechanism, and also provides the necessary libraries for
running all the demos.

Tested (on NixOS):
- check out master
- run `nix shell`
- in the shell:
  - run `./build.sh net-bootstrap`
  - run `./factor -i=factor.image`
  - enter `USE: demos demos`
  - play around with the demos

Not Tested:
- core unit tests (likely to fail until the corresponding issues in #2180 are addressed)

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,20 +1,27 @@
 { pkgs ? import <nixpkgs> {} }:
-(pkgs.buildFHSUserEnv {
-  name = "factor";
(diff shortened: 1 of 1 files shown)


timor:
@kwshi Could you try this on your NixOS system, if possible?  Thanks.

PR #2201, opened 2019-10-31
Don't use rm from the nmake build.

melted:


melted:
As one typically uses nmake from the Visual Studio CMD console, I think this makes sense.

mrjbq7:
Merged!

PR #2202, opened 2019-10-31
cache: Don't use in-place filter

### LINKED ISSUE #1978: UI gets corrupted after using Factor normally, cross-platform rendering bug of some kind

melted:
This solves a UI corruption problem by not reusing the hashtable when
purging the cache. The root cause of why the hashtable gets corrupted
when filtering in place hasn't been found.

Fixes #1978.

### DIFF
--- a/basis/cache/cache.factor
+++ b/basis/cache/cache.factor
@@ -38,9 +38,9 @@ M: cache-assoc dispose* clear-assoc ;
 PRIVATE>
 
 : purge-cache ( cache -- )
-    [ assoc>> ] [ max-age>> ] bi V{ } clone [
+    dup [ assoc>> ] [ max-age>> ] bi V{ } clone [
         '[
             nip dup age>> 1 + [ >>age ] keep
             _ < [ drop t ] [ _ dispose-to f ] if
-        ] assoc-filter! drop
+        ] assoc-filter >>assoc drop 
     ] keep [ last rethrow ] unless-empty ;


melted:
This might be a bit more inefficient, but assoc-each is very inefficient compared to what a word specialized for hashtables would be (>alist already extracts all the elements), so I don't feel too bad about it.

melted:
I'm testing this now. So far so good, but as the bug is appearing infrequently I guess it's best to hold off merging it a bit.

mrjbq7:
@melted I'm not sure why using ``assoc-filter!`` would break hashtables, but did this seem to fix your UI corruption?

mrjbq7:
@melted have you seen UI corruption since applying this patch?

melted:
It works fine! It still bugs that I didn't find the root cause for why the other goes off the rails sometimes.

melted:
I know what happens, it calls the finalizers for a bunch of textures, but they don't get deleted from the hashmap. Hence why random textures appear when the texture IDs are reused.

erg:
I pushed this to master. Let's try to figure out if it fixes everything or the root cause. Also, if we keep seeing the corruption, maybe we will revert this one--it's hard to trigger.

Thanks for tracking down a potential fix!

PR #2205, opened 2019-11-03
Linux kernel 5 is released

catb0t:
The test failed on e.g. `Linux $hostname 5.3.0-20-generic`

### DIFF
--- a/basis/system-info/linux/linux-tests.factor
+++ b/basis/system-info/linux/linux-tests.factor
@@ -11,5 +11,5 @@ tools.test kernel ;
 [ t ] [ domainname string? ] unit-test
 
 { t } [
-    release "." split1 drop { "2" "3" "4" } member?
+    release "." split1 drop { "2" "3" "4" "5" } member?
 ] unit-test


erg:
This test kind of sucks. Can we do a fix that won't break on Linux 6?

catb0t:
All the unit tests for this vocab are kind of contrived. I have no idea what this could be usefully changed to instead.

mrjbq7:
Maybe change it to get Linux version a second way and compare? Like uname or something.

timor:
I think this test only makes sense if there is a list of officially supported kernel version numbers.

catb0t:
@mrjbq7 This vocab uses the `(uname)` system call to get information. The `uname` command also uses that syscall of course, which would mostly test the consistency of `uname(1)` with `uname(2)` (or Factor's syscall ability, which is not the scope of this vocab).

However the man page `NOTES` says 
> Part of the utsname information is also accessible  
> via  `/proc/sys/kernel/{ostype, hostname, osrelease, version, domainname}`.

Reading those files directly could be a good test that Factor's invocation of the `uname(2)` syscall is correct?

mrjbq7:
Perhaps not worth worrying about.  Maybe we should just remove the test someday.

In any event, merged your patch!

Issue #2206, opened 2019-11-05
Adapt commandline parsing to support '--help' and '--version' flags


nomennescio:
It would be beneficial to have Factor's commandline parsing (vm_parameters::init_from_args (..)) support the common flags `--help` to show available options, as well as a flag `--version` to show the version of Factor.

nomennescio:
On Windows, running factor.exe -help does not show any help, because the commandline switches are parsed in Factor, and before that, stdio is redirected to the nullwriter. If issue #2204 would be implemented, it would show output. As for the version flag, in #2208 @catb0t  indicates he would like to work on commandline parsing, including aliases for some other switches. I leave this issue open for that.

Issue #2184, opened 2019-09-27
Command line arguments


mrjbq7:
I was playing around, adding ``-version`` and ``-q`` options to the listener:

```diff
diff --git a/basis/listener/listener.factor b/basis/listener/listener.factor
index b308ab5947..23df22b3f1 100644
--- a/basis/listener/listener.factor
+++ b/basis/listener/listener.factor
@@ -228,6 +228,10 @@ SYMBOL: interactive-vocabs
     ] with-interactive-vocabs ;
 
 : listener-main ( -- )
-    version-info print flush listener ;
+    "version" get [
+        "Factor " write vm-version print flush
+    ] [
+        "q" get [ version-info print flush ] unless listener
+    ] if ;
 
 MAIN: listener-main
```

And realized that there's no nice way to add "vocabulary-level arguments" to the VM-level help that's printed by ``./factor -h`` which is defined in ``command-line.startup``.

Perhaps we need an API to query a vocabulary for command-line documentation, or a hook to update it in a command-line vocab or something...

PR #2208, opened 2019-11-05
Feature commandline help

nomennescio:
Proof of concept of adding '--help' to commandline. Full help text to be supplied, see #2206.

Note that 'exit()' is required because the destructors assume malloc has already been called, which is not true when '--help' is finished.

### DIFF
--- a/vm/factor.cpp
+++ b/vm/factor.cpp
@@ -154,7 +154,7 @@ void factor_vm::factor_sleep(long us) {
 
 void factor_vm::start_standalone_factor(int argc, vm_char** argv) {
   vm_parameters p;
-  p.init_from_args(argc, argv);
+  if (!p.init_from_args(argc, argv)) return;
   init_factor(&p);
   pass_args_to_factor(argc, argv);
 
--- a/vm/factor.cpp
+++ b/vm/factor.cpp
@@ -154,7 +154,7 @@ void factor_vm::factor_sleep(long us) {
 
 void factor_vm::start_standalone_factor(int argc, vm_char** argv) {
   vm_parameters p;
-  if (!p.init_from_args(argc, argv)) return;
+  if (!p.init_from_args(argc, argv)) exit (0);
   init_factor(&p);
   pass_args_to_factor(argc, argv);
 
--- a/vm/image.cpp
+++ b/vm/image.cpp
@@ -49,7 +49,7 @@ vm_parameters::~vm_parameters() {
   free((vm_char *)executable_path);
 }
 
-void vm_parameters::init_from_args(int argc, vm_char** argv) {
+bool vm_parameters::init_from_args(int argc, vm_char** argv) {
   int i = 0;
 
   for (i = 1; i < argc; i++) {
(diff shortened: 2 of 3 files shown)


mrjbq7:
We do have ``--help`` equivalent:

```
$ ./factor -help
Usage: factor [Factor arguments] [script] [script arguments]

Factor arguments:
    -help               print this message and exit
    -i=<image>          load Factor image file <image> (default factor.image)
    -run=<vocab>        run the MAIN: entry point of <vocab>
        -run=listener   run terminal listener
        -run=ui.tools   run Factor development UI
    -e=<code>           evaluate <code>
    -no-user-init       suppress loading of .factor-rc
    -datastack=<int>    datastack size in KiB
    -retainstack=<int>  retainstack size in KiB
    -callstack=<int>    callstack size in KiB
    -callbacks=<int>    callback heap size in KiB
    -young=<int>        young gc generation 0 size in MiB
    -aging=<int>        aging gc generation 1 size in MiB
    -tenured=<int>      tenured gc generation 2 size in MiB
    -codeheap=<int>     codeheap size in MiB
    -pic=<int>          max pic size
    -fep                enter fep mode immediately
    -no-signals         turn off OS signal handling
    -console            open console if possible
    -roots=<paths>      a list of ":"-delimited extra vocab roots

Enter
    "command-line" help
from within Factor for more information.
```

It's printed not from the VM, but from the ``command-line.startup`` vocabulary currently.

nomennescio:
Ah I see! I didn't stumble upon it when trying all options to get help. The common ones usually tried are:
'/?', '-h', and '--help'

Together with the fact that commandline parsing already happens in the VM, I thought it was just missing. At least the '-help' option is missing from the '"command-line" help' documentation.

And currently the help message is only printed from 'factor.com', 'factor.exe' is silent.

Is it an idea to:
- Add the '-help' option to '"command-line" help'
- Add aliases '-h' and '--help' for '-help'

catb0t:
I have been long considering porting a "proper" un-trickable command line argument parser to Factor (that would make supporting `-h`/`--help`/`-help` easier/more automatic),  just haven't settled on the design details to actually work on it yet

nomennescio:
> I have been long considering porting a "proper" un-trickable command line argument parser to Factor (that would make supporting `-h`/`--help`/`-help` easier/more automatic), just haven't settled on the design details to actually work on it yet

Why not do something similar to 'getopt'? Given Factor's rich types that should be not too difficult.

catb0t:
Like Getopt, except Getopt (including e.g. Getopt::Long in Perl) is an antiquated artifact of C programming idioms, and really belongs in museums, that we should move beyond, for a more modern solution

nomennescio:
See also #2184

mrjbq7:
I made a change just now to allow "--help" to work...

67bc5a92891d51824ef6b4a896ff49ca107ae774

I agree that it's the more common long option, but I didn't want to make "-help" not work (yet) since that's what we've used for awhile.

PR #2214, opened 2019-11-08
Add description of '-help' switch to documentation.

nomennescio:
See #2206


erg:
I'm trying to make github the source of truth, but now when I merge nothing goes to master...

Issue #2199, opened 2019-10-25
Running build.sh net-bootstrap fails on Cygwin64 with CLANG


nomennescio:
When trying to run build.sh net-bootstrap from Cygwin64 with clang installed, I got the following error:
```
Building factor with 8 cores
make factor factor-ffi-test factor-lib CONFIG=vm/Config.windows.x86.64
make[1]: Entering directory '/d/project/build/factor'
windres.exe vm/factor.rs vm/resources.o
make[1]: *** No rule to make target 'vm/mvm-windows.hpp', needed by 'vm/master.hpp.gch'.  Stop.
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/d/project/build/factor'
make: *** [GNUmakefile:185: windows-x86-64] Error 2
make failed
```

The mvm-windows.hpp comes from
`vm/Config.windows:PLAF_MASTER_HEADERS += vm/os-windows.hpp vm/mvm-windows.hpp`

No such file exist in the repo, although there is a mvm-windows.cpp and mvm.hpp.
Even worse,
 git log --full-history  -- vm/mvm-windows.hpp
shows such a file has never existed.

How come?

mrjbq7:
We use ``build.cmd`` on windows, not sure if we've ever used ``build.sh`` in a cygwin environment like that?  @erg ?

nomennescio:
It seems Cygwin support is fully built in. It would be great to have a free alternative to the MS proprietary toolchain working again. It seems like a minor issue. The Nmakefile does not reference that header file either, so maybe it can just be removed from GNUmakefile.

nomennescio:
Just tried a little bit; it's more than just removing the vm/mvm-windows.hpp from the GNUmakefile. The current configuration only uses flags for the Cygwin g++ compiler, which are not suitable for the Cygwin clang c++ compiler. And it seems it misses some declarations. I will try to make it work, cannot be that difficult.

nomennescio:
Seem Cygwin does not support the no-cygwin flag since gcc 4.0. To 'cross-compile' Windows binaries something else is needed instead. This might actually be a bit more work than anticipated.

nomennescio:
Have successfully compiled both 32 bit and 64 bit versions of Factor under Cygwin using MinGW-W64 g++, which required only minor adaptations to the existing code. The only thing failing currently is running the 32 bit version due to missing DLLs, which I will try to fix.

nomennescio:
[...]

PR #2215, opened 2019-11-09
Support for MinGW compiler. Need to instal MinGW compiler and runtime.

nomennescio:
See #2199

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -92,7 +92,7 @@ check_ret() {
 set_downloader() {
     test_program_installed wget
     if [[ $? -ne 0 ]] ; then
-        DOWNLOADER=wget
+        DOWNLOADER="wget -nd"
         DOWNLOADER_NAME=wget
         return
     fi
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -92,7 +92,7 @@ check_ret() {
 set_downloader() {
     test_program_installed wget
     if [[ $? -ne 0 ]] ; then
-        DOWNLOADER=wget
+        DOWNLOADER="wget -nd"
```
mrjbq7: Why do you need ``--no-directories``?

vm/os-windows.cpp
```diff
@@ -185,12 +185,14 @@ uint64_t nano_count() {
 
 void sleep_nanos(uint64_t nsec) { Sleep((DWORD)(nsec / 1000000)); }
 
+#ifndef EXCEPTION_DISPOSITION
```
mrjbq7: Why is this ``ifndef`` added?

catb0t: It probably prevents an "illegal re-declaration of enum EXCEPTION_DISPOSITION" when MinGW pre-defines it

build.sh
nomennescio: Without it the binary ends up in the wrong place. Maybe because of different defaults on Unix or because of a ~/.wgetrc?

vm/os-windows.cpp
nomennescio: That's exactly right. MinGW predefines it using defines. This is the simplest way to adapt for that.

mrjbq7:
@erg want to look this over?

PR #2216, opened 2019-11-09
help.lint.coverage: fix for shadowing "empty" word; prevent the other…

catb0t:
… test-only words from being shadowed too

When `load-all test-all`ing, there would be at least 3 `empty` words in scope, and then the `help.lint.coverage` test fails as intended because of an `ambiguous-use-error`. This change makes them all more unique.

I could also have forced the test-only words to be used namespaced-only which would be much more robust, but getting `find-word` to work with namespaced words (like `a:b`) would be complicated for little benefit (right now).

### DIFF
--- a/extra/help/lint/coverage/coverage-docs.factor
+++ b/extra/help/lint/coverage/coverage-docs.factor
@@ -10,7 +10,7 @@ PRIVATE>
 ABOUT: "help.lint.coverage"
 
 ARTICLE: "help.lint.coverage" "Help coverage linting"
-"The " { $vocab-link "help.lint.coverage" } " vocabulary implements a very picky documentation completeness checker."
+"The " { $vocab-link "help.lint.coverage" } " vocabulary implements a very pedantic documentation completeness checker."
 $nl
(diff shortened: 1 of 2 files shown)


catb0t:
Actually, this shadowing does not happen with `load-all "help.lint.coverage" [ reload ] [ test ] bi`, it just happens because I think `test-all` has a singular vocabulary and word scope for every vocab it tests? That seems to be what's going on...

mrjbq7:
``find-word`` does work with qualified words:

```
IN: scratchpad "math:+" find-word .
+
```

But still, I like your change and merged it.

catb0t:
`"empty" find-word` finds  only `help.lint.coverage.tests.private:empty` (when that's loaded) before `load-all test-all`, but after `load-all test-all`, `"empty" find-word`'s ambiguous-use restart shows also definitions from `classes.tuple.tests` and `compiler.tests.simple`. Is this a bug in the way `test-all` works?

mrjbq7:
Not a bug with find-word, just a side effect of having multiple empty definitions after those vocabs are loaded.  I guess you could fix the test by doing a qualified name for it, but I think your fix is fine. At least until someone makes “an-empty-word” and it breaks again! ;-)

PR #2217, opened 2019-11-10
soundex: move to extra as it's unused; fix authors.txt filename

catb0t:
Pretty cool vocab but I don't think it needs to be in basis, especially since it's not used by any other vocab. Also, the `author.txt` file wasn't picked up by the help system, but is now.


mrjbq7:
Possibly a good idea.  I love having one big tree of vocabs and hope that a good tagging and search mechanism can make it easy to discover stuff (e.g., the “collections” tag).

The core/basis distinction is really only because of how we bootstrap right now.  The basis/extra distinction has been either 1) code quality and completeness, or 2) standard library versus extended library, or 3) arbitrary. 

What are your thoughts?

catb0t:
I know (or I thought) that all of `core` and `basis` are compiled into `factor.image` during bootstrap (or is it only as needed?). This being a toy vocab, having nothing depend upon it, and not being essential for basic or everyday programs, makes it an extended library.

Reducing the size of `factor.image` is a good thing, but I didn't realise there was any arbitrary-ness in the `basis`/`extra` dichotomy. 

Being extended-library alone, is enough to me that it belongs in extra.

mrjbq7:
Not all of basis is loaded currently, but a decent amount of it is.

mrjbq7:
I merged this, but aside from maybe ``roman``, I don't think there's other candidates for moving to ``extra`` right now.

catb0t:
I only noticed this because i was watching the `load-all` scroll by. I wasn't going to start on `roman`, and i think the rest of basis is pretty essential otherwise

PR #2218, opened 2019-11-13
grc-pass: new vocab

AlexIljin:
A small contribution: this vocab requests a new password from https://www.grc.com/passwords.htm.
Very inefficient, but I use it all the time, might as well automate it.
Plus, it's a little homage to the great Steve Gibson.

### DIFF
--- a/extra/grc-pass/authors.txt
+++ b/extra/grc-pass/authors.txt
@@ -1,0 +1,1 @@
+Alexander Ilin
--- a/extra/grc-pass/grc-pass.factor
+++ b/extra/grc-pass/grc-pass.factor
@@ -1,0 +1,33 @@
+! Copyright (C) 2019 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: accessors http.client kernel math sequences splitting ;
+
+IN: grc-pass
+
+<PRIVATE
+
+: get-pass ( key -- string )
+    "https://www.grc.com/passwords.htm" http-get swap code>> 200 assert=
+    string-lines [ dupd subseq? ] find 2nip
+    "</font></td></tr></table>" over subseq-start head
+    dup [ CHAR: > = ] find-last drop 1 + tail ;
+
+: replace-& ( string -- string' )
+    "&gt;" ">" replace "&lt;" "<" replace "&amp;" "&" replace ;
+
+! Replace &amp; last to avoid side-effects. For example, "&amp;lt;" should
+! become "&lt;", not "<".
+
+PRIVATE>
+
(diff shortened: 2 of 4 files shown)


bpollack:
Erm…not even from a Factor standpoint as such, but from an infosec standpoint, I'm really uncomfy with this. I'd be 100.00% on board with adding a password generating vocab to Factor, just…there are so many things here that make me nervous as hell.

Do you want to pair on a password generation vocab maybe? It'd work offline, we can make it arbitrarily secure, and I can think of places we could use it (e.g. in furnace)

AlexIljin:
@bpollack, will the `http-get` call fail if the SSL cert is not valid?

AlexIljin:
A good password generation vocab may be very involved and very simple, depending on how much you trust the source of entropy provided by the OS. Personally, I would probably use something well vetted, like libsodium.

bpollack:
@AlexIljin I genuinely forget re. certs; I had written that, but I think @erg had to back at least _some_ bits of that out due to compat issues. My goal at the time had been to move us entirely to libtls/libressl, but I never quite got there. I can dig in if you want (and try to restore the functionality without breaking things if desired).

As far as password generation: I think our random number generation should be okay; did you see something in `secure-random-generator` that you felt was wanting? Linux weirdness aside, `/dev/urandom` for unixes should be fine as an entropy source, and Windows is correctly using `CryptGenRandom`, so building directly off that (over e.g. the built-in Mersene twister, which still seeds from those) should be _plenty_ secure. The bigger part of the vocab would just be deciding what to put in it; I'd love to have a 1Password-style config where we can do words or random data, and I don't _think_ that'd be all that hard.

Thoughts/concerns/preferred desert toppings?

AlexIljin:
@bpollack, I don't think there is anything wrong with `secure-random-generator`, this PR is about creating a web service like we have in `extra/google/*` vocabs around the thing I use often to get my passwords. Currently it's a manual process for me: open the web site, select some text with the mouse (ugh!), then copy-paste somewhere (typically, a 7-zip archive creation dialog), so I wanted to automate it and at the same time provide an example vocab.

Creating a password generator based on `secure-random-generator` is a great idea, but I would still like to add this web service, because why not have both?

Basically, for a password I'd like to generate an N-character string, where each character is "secure-randomly" selected with equal probability from the set of [33,126] ASCII character codes. Maybe something like this:
```
: ascii-password ( n -- string )
    [
        33 126 [a,b] [ random ] curry "" replicate-as
    ] with-secure-random ;
```

AlexIljin:
Submitted PR #2219 for the password generation thing.

PR #2222, opened 2019-11-22
Basic Tensor Functionality

nandeeka:
This PR includes the code for a basic scalable matrix manipulation vocabulary. All public words are accompanied by tests and documentation.


extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+    0 repetition ;
+
+! Construct a tensor of ones
+: ones ( shape -- tensor )
+    1 repetition ;
+
+! Construct a one-dimensional tensor with values start, start+step,
+! ..., stop (inclusive)
+: arange ( a b step -- tensor )
+    <range> [ length 1array ] keep [ ] float-array{ } map-as <tensor> ;
```
mrjbq7: ```[ ] float-array{ } map-as``` can also be ``alien.c-types:float >c-array``

mrjbq7: Also, you have ``>float-array`` up above that contains the same definition.

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+    [ drop ] dip ; ! remove accumulator
+
+! helper for transpose: given shape & flat index, gives nd index for transpose
+:: trans-index ( ind shape -- seq )
+    ! what we use to divide things
+    shape reverse :> S
+    ! what we multiply by
+    shape get-mults :> mults
+    ! accumulator
+    { }
```
mrjbq7: Might be better to make the accumulator a ``vector`` (e.g., ``V{ } clone``, and then you can ``push`` onto it).

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+    ! loop thru amnts we want to slice off of S
+    0 [ dup S length < ]
```
mrjbq7: Maybe easier to do something like:

```
S [| i n |
    ...
] each-index
```

Since you seem to use the index for ``head-slice product`` and the value for ``m``.

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+    { shape array }
+    { vec float-array } ;
+
+! Errors
+ERROR: shape-mismatch-error shape1 shape2 ;
+
+<PRIVATE
+
+! Construct a tensor of zeros
+: <tensor> ( shape seq -- tensor )
```
mrjbq7: May want some basic assertions that the shape has positive (or at least non-negative?) values.

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+    tensor boa ;
+
+: >float-array ( seq -- float-array )
+    [ ] float-array{ } map-as ;
+
+:: repetition ( shape const -- tensor )
```
mrjbq7: This could be a one-line definition if you wanted to:

```
: repetition ( shape const -- tensor )
    [ dup product ] dip <repetition> >float-array <tensor> ;
```

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+! ..., stop (inclusive)
+: arange ( a b step -- tensor )
+    <range> [ length 1array ] keep [ ] float-array{ } map-as <tensor> ;
+
+! Construct a tensors with vec { 0 1 2 ... } and reshape to the desired shape
+: naturals ( shape -- tensor )
+    [ ] [ product [0,b) >float-array ] bi <tensor> ;
+
+! Reshape the tensor to conform to the new shape
+: reshape ( seq tensor -- tensor )
```
mrjbq7: You might change the stack effect to be ``( shape tensor -- tensor )``, and maybe think about re-ordering stack effects, and extracting the shape checking out:

```
: check-shapes ( shape1 shape2 -- shape1 shape2 )
    2dup [ product ] bi@ = [ shape-mismatch-error ] unless ;

: reshape ( tensor shape -- tensor )
    '[ _ check-shapes nip ] change-shape ;
```

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+    [ swap >>shape ]
+    [ shape>> swap shape-mismatch-error ] if ;
+
+! Flatten the tensor so that it is only one-dimensional
+: flatten ( tensor -- tensor )
+    dup shape>>
+    product { } 1sequence >>shape ;
+
+! Outputs the shape of a tensor
+: shape ( tensor -- shape )
```
mrjbq7: If the accessor is sufficient (meaning this is just a simple alias), we should probably just use that for now.

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+! Turn into Factor ND array form
+! Source: shaped-array>array
+TYPED: tensor>array ( tensor: tensor -- seq: array )
+    [ vec>> >array ] [ shape>> ] bi
+    [ rest-slice reverse [ group ] each ] unless-empty ;
+
+<PRIVATE
+
+! Apply the binary operator bop to combine the tensors
+TYPED:: t-bop ( tensor1: tensor tensor2: tensor bop: ( x y -- z ) -- tensor: tensor )
```
mrjbq7: I might rename ``bop`` and ``uop`` to just ``quot:`` for consistency with some of our other style, or make it ``binary-quot:`` and ``unary-quot`` maybe.

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+        ! Push n and p and multiply the clices
+        n p 2d-matmul
+        0
+    ] map drop
+    tensor3 ;
+
+<PRIVATE
+! helper for transpose: gets the turns a shape into a list of things
+! by which to multiply indices to get a full index
+: get-mults ( shape -- seq )
```
mrjbq7: This could use a better name maybe.

Other ideas for the implementation:

```
: get-mults ( shape -- seq )
    rest-slice <reversed> cum-product { 1 } prepend ;
```

I suspect this and ``trans-index`` are pretty slow since you're making these multiplier arrays for each index operation.

mrjbq7: And the multiplier arrays are constant with shape, not index.

extra/tensors/tensors.factor
```diff
@@ -0,0 +1,257 @@
...
+            1array append
+        ] dip
+        ! increment iterator
+        1 +
+    ] while drop ! drop iterator
+    reverse ;
+PRIVATE>
+
+! Transpose an n-dimensional tensor
+TYPED:: transpose ( tensor: tensor -- tensor': tensor )
```
mrjbq7: Have you done any benchmarking or profiling of the code to see how it performs roughly against something like ``flip`` for a two-dimensional array, or against ``ndarray.T``?

extra/tensors/tensors.factor
nandeeka: This has now been included in the functions that call `<tensor>`, because it sometimes needs to happen well before `<tensor>` is called.

extra/tensors/tensors.factor
nandeeka: I updated this, but the reshape implementation is not quite what you suggested. Let me know if there are any further changes.

extra/tensors/tensors.factor
nandeeka: No benchmarking has been done yet. That would be a goal for next semester.

PR #2227, opened 2019-12-13
Improve Tensor Operation Performance

nandeeka:
This pull request contains a number of updates to the tensors vocabulary that improve the performance of matrix operations.


mrjbq7:
I merged this manually since the github button wasn't working, so it got a new gitref on commit and might require you to do some git shenanigans locally.

Issue #2230, opened 2019-12-31
System V ABI non-compliance?


kusumotonorio:
Because of a doubt, I wanted to see what happens when C invokes the Factor function, but I didn't know how to experiment with it. Instead, I decided to experiment with the behavior of calling C's function from Factor.

I created the following dynamic library:
```C
ffitest.c:

typedef struct ST1 {
  unsigned long mem1;
  unsigned long mem2;
} st1;
  
unsigned long test_func1(unsigned long a, unsigned long b, unsigned long c, st1 d, st1 e) {
    unsigned long x;
	
    x = a + b + c;
	return x;
}

unsigned long test_func2(unsigned long a, unsigned long b, unsigned long c, st1 d, st1 e) {
    unsigned long x;
	
    x = a + b + c + d.mem1 + d.mem2;	
	return x;
}

unsigned long test_func3(unsigned long a, unsigned long b, unsigned long c, st1 d, st1 e) {
    unsigned long x;
	
    x = a + b + c + d.mem1 + d.mem2 + e.mem1;
	return x;
}

unsigned long test_func4(unsigned long a, unsigned long b, unsigned long c, st1 d, st1 e) {
    unsigned long x;
	
    x = a + b + c + d.mem1 + d.mem2 + e.mem1 + e.mem2;
	return x;
}
```

Next, I wrote the code for Factor like this:
```Factor
ffitest.factor:

USING: alien alien.libraries alien.c-types alien.syntax classes.struct ;
IN: ffitest

<< "libdylibx64" "/Applications/factor/ffitest.dylib" cdecl add-library >>

STRUCT: st1 { mem1 ulong } { mem2 ulong } ;

LIBRARY: ffitest
FUNCTION: ulong test_func1 ( ulong a, ulong b, ulong c, st1 d, st1 e )
FUNCTION: ulong test_func2 ( ulong a, ulong b, ulong c, st1 d, st1 e )
FUNCTION: ulong test_func3 ( ulong a, ulong b, ulong c, st1 d, st1 e )
FUNCTION: ulong test_func4 ( ulong a, ulong b, ulong c, st1 d, st1 e )
```

And, I experimented with them.

```
IN: scratchpad USING: classes.struct ffitest ;
Loading resource:work/ffitest/ffitest.factor
IN: scratchpad 1 2 3 S{ st1 f 4 5 } S{ st1 f 6 7 } test_func1

--- Data stack:
6
[...]

PR #2235, opened 2020-01-19
Improved System V AMD64 ABI compliance

kusumotonorio:
This PR fixes an issue when passing a structure as a value argument. This became a problem in System V AMD64 ABI.
The details of this problem can be found in #2230.

### DIFF
--- a/basis/compiler/cfg/builder/alien/boxing/boxing.factor
+++ b/basis/compiler/cfg/builder/alien/boxing/boxing.factor
@@ -13,21 +13,17 @@ SYMBOL: struct-return-area
 SYMBOLS: int-reg-reps float-reg-reps ;
 
 : record-reg-reps ( reps -- reps )
-    dup [
-        dup second not [  ! on-stack?: f 
-            first int-rep? int-reg-reps float-reg-reps ?
-            dup get [ inc ] [ drop ] if
-        ] [ drop ] if
-    ] each ;
+    dup  ! reps: { { reg-rep on-stack? odd-register? } ... }
+    [ [ [ second not ] [ first int-rep?     ] bi and ] count int-reg-reps   +@ ]
+    [ [ [ second not ] [ first int-rep? not ] bi and ] count float-reg-reps +@ ]
+    bi ;
 
 : unrecord-reg-reps ( reps -- reps )
-    dup [
-        dup second not [  ! on-stack?: f 
-            first int-rep? int-reg-reps float-reg-reps ?
-            dup get [ dec ] [ drop ] if
-        ] [ drop ] if
-    ] each ;
-
+    dup
+    [ [ [ second not ] [ first int-rep?     ] bi and ] count -1 * int-reg-reps   +@ ]
+    [ [ [ second not ] [ first int-rep? not ] bi and ] count -1 * float-reg-reps +@ ]
+    bi ;
+    
 GENERIC: flatten-c-type ( c-type -- pairs )
 
 M: c-type flatten-c-type
--- a/basis/compiler/cfg/builder/alien/boxing/boxing.factor
+++ b/basis/compiler/cfg/builder/alien/boxing/boxing.factor
@@ -12,18 +12,22 @@ SYMBOL: struct-return-area
 
 SYMBOLS: int-reg-reps float-reg-reps ;
 
+<PRIVATE
+
 : inc-not-f ( variable -- ) dup get [ inc ] [ drop ] if ; inline
 
 : dec-not-f ( variable -- ) dup get [ dec ] [ drop ] if ; inline
 
-: record-reg-reps ( seq -- seq )
+PRIVATE>
+
+: record-reg-reps ( reps -- reps )
     dup [
         dup second not [  ! on-stack?: f 
             first int-rep? int-reg-reps float-reg-reps ? inc-not-f
         ] [ drop ] if
     ] each ;
 
-: unrecord-reg-reps ( seq -- seq )
+: unrecord-reg-reps ( reps -- reps )
     dup [
         dup second not [  ! on-stack?: f 
             first int-rep? int-reg-reps float-reg-reps ? dec-not-f
--- a/basis/compiler/cfg/builder/alien/boxing/boxing.factor
+++ b/basis/compiler/cfg/builder/alien/boxing/boxing.factor
@@ -12,18 +12,15 @@ SYMBOL: struct-return-area
 
 SYMBOLS: int-reg-reps float-reg-reps ;
 
-: record-reg-reps ( reps -- reps )
-    dup  ! reps: { { reg-rep on-stack? odd-register? } ... }
-    [ [ [ first int-rep?     ] [ second not ] bi and ] count int-reg-reps   +@ ]
-    [ [ [ first int-rep? not ] [ second not ] bi and ] count float-reg-reps +@ ]
-    bi ;
+: reg-reps ( reps -- int-reps float-reps )
+    [ second ] reject [ [ first int-rep? ] count ] [ length over - ] bi ;
+
+: record-reg-reps ( reps -- reps )            
+    dup reg-reps [ int-reg-reps +@ ] [ float-reg-reps +@ ] bi* ;
 
 : unrecord-reg-reps ( reps -- reps )
-    dup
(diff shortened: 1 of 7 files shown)


basis/compiler/cfg/builder/alien/boxing/boxing.factor
```diff
@@ -10,19 +10,33 @@ IN: compiler.cfg.builder.alien.boxing
 
 SYMBOL: struct-return-area
 
+SYMBOLS: int-reg-reps float-reg-reps ;
+
+: record-reg-reps ( reps -- reps )
+    dup  ! reps: { { reg-rep on-stack? odd-register? } ... }
+    [ [ [ first int-rep?     ] [ second not ] bi and ] count int-reg-reps   +@ ]
```
mrjbq7: Maybe these are simpler like this:

```factor
: record-reg-reps ( reps -- reps )
    dup [
        dup second [ drop ] [ first int-rep? int-reg-reps float-reg-reps ? inc ] if
    ] each ;

: unrecord-reg-reps ( reps -- reps )
    dup [
        dup second [ drop ] [ first int-rep? int-reg-reps float-reg-reps ? dec ] if
    ] each ;
```

Or maybe (although this has an intermediate array created):

```factor
: record-reg-reps ( reps -- reps )
    dup [ second ] reject [ first int-rep? int-reg-reps float-reg-reps ? inc ] each ;

: unrecord-reg-reps ( reps -- reps )
    dup [ second ] reject [ first int-rep? int-reg-reps float-reg-reps ? dec ] each ;
```

At the least, ``-1 *`` should be ``neg``, but it would be nice to iterate once, or simplify these similar expressions below.

mrjbq7: Or even, doing a single namespace increment:

```factor
: reg-reps ( reps -- int-reps float-reps ):
    [ second ] reject [ [ first int-rep? ] count ] [ length over - ] bi

: record-reg-reps ( reps -- reps )
    dup reg-reps int-reg-reps float-reg-reps [ +@ ] bi-curry@ bi* ;

: unrecord-reg-reps ( reps -- reps )
    dup reg-reps [ neg ] bi@ int-reg-reps float-reg-reps [ +@ ] bi-curry@ bi* ;
```

basis/cpu/x86/64/unix/unix.factor
```diff
@@ -24,16 +24,27 @@ M: x86.64 reserved-stack-space 0 ;
...
-    struct-types&offset split-struct [
+:: flatten-small-struct ( c-type -- seq )
+    c-type struct-types&offset split-struct [
         [ lookup-c-type c-type-rep reg-class-of ] map
         int-regs swap member? int-rep double-rep ?
         f f 3array
-    ] map ;
+    ] map :> reps
+    int-reg-reps get float-reg-reps get and [
+        reps [ first int-rep? ] count :> int-mems
```
mrjbq7: If you made ``reg-reps`` like above, this could just be:

```factor
reps reg-reps :> ( int-mems float-mems )
```

Looking at it in more detail, in ``reg-reps`` you check for ``second not`` but here you don't... 

Bug?  Or not possible code path?

basis/cpu/x86/64/unix/unix.factor
```diff
@@ -24,16 +24,27 @@ M: x86.64 reserved-stack-space 0 ;
...
+            reps [ first t f 3array ] map
+        ] [ reps ] if
+    ] [ reps ] if ;
 
 M: x86.64 flatten-struct-type ( c-type -- seq )
     dup heap-size 16 <=
-    [ flatten-small-struct ] [ call-next-method [ first t f 3array ] map ] if ;
+    [ flatten-small-struct record-reg-reps ] [
+        call-next-method [ first t f 3array ] map
+        unrecord-reg-reps
```
mrjbq7: Isn't this doing no work since ``second`` is always ``t``?  (Due to ``[ first t f 3array ] map``)? 

So ``unrecord-reg-reps`` will not decrement anything?

basis/compiler/cfg/builder/alien/boxing/boxing.factor
mrjbq7: On a related question, is it true that these are either "int-rep" or "float-rep" for the purposes of your change?  Or at least that practically that's fine when you have other c-types in a struct?

basis/cpu/x86/64/unix/unix.factor
kusumotonorio: Indeed, this is obviously strange. I ponder what I wanted to do. 
Please wait for a while to consider other points too.

basis/compiler/cfg/builder/alien/boxing/boxing.factor
kusumotonorio: > On a related question, is it true that these are either "int-rep" or "float-rep" for the purposes of your change? Or at least that practically that's fine when you have other c-types in a struct?

The names of the variables `int-reg-reps` and `float-reg-reps` are directly related to CPU register types, rather than C type names.
For example, in X86-64:

```Factor
IN: cpu.x86.64.unix

M: x86.64 param-regs
    drop {
        { int-regs { RDI RSI RDX RCX R8 R9 } }
        { float-regs { XMM0 XMM1 XMM2 XMM3 XMM4 XMM5 XMM6 XMM7 } }
    } ;
```
The word that Factor uses to allocate the actual registers for its arguments is `next-reg-param` :
```Factor
IN: compiler.cfg.builder.alien.params
GENERIC: next-reg-param ( odd-register? rep -- reg )

USING: compiler.cfg.builder.alien.params cpu.architecture kernel
math namespaces sequences ;
M: int-rep next-reg-param
    [ nip ?dummy-stack-params ] [ nip ?dummy-fp-params ] [
        drop
        [ int-regs get last even? [ int-regs get pop* ] when ]
        when
    ] 2tri int-regs get pop ;

USING: compiler.cfg.builder.alien.params cpu.architecture kernel
namespaces sequences ;
M: object next-reg-param
    nip [ ?dummy-stack-params ] [ ?dummy-int-params ] bi
    float-regs get pop ;
```

Here, `int-rep` is allocated from the registers of `int-regs`, while others are allocated from `float-regs`.

Factor provides `int-rep` for C type arguments that use integer registers.

```Factor
IN: scratchpad USING: alien.c-types compiler.cfg.builder.alien.boxing classes.struct ;
IN: scratchpad int base-type flatten-c-type .
{ { int-rep f f } }
alien.c-types:short base-type flatten-c-type .
{ { int-rep f f } }
IN: scratchpad long base-type flatten-c-type .
{ { int-rep f f } }
IN: scratchpad STRUCT: s1 { m1 long } { m2 ulong } ;
IN: scratchpad s1 base-type flatten-c-type .
{ { int-rep f f } { int-rep f f } }

File: basis/cpu/x86/64/unix/unix.factor (hunk 2 above)
Comment (kusumotonorio) [reply basis/cpu/x86/64/unix/unix.factor]: The second element of the `reps` element is all `f`, so it was not checked here.
Even if the check is done, I don't think it will affect the operation. (Strictly speaking, it affects the execution speed.)

File: basis/compiler/cfg/builder/alien/boxing/boxing.factor (hunk 1 above)
Comment (kusumotonorio) [reply basis/compiler/cfg/builder/alien/boxing/boxing.factor]: I think all three suggestions are better than the original. The third, in particular, allows us to use `reg-reps` elsewhere, making the code cleaner. 
However, the name may indicate that `reg-reps` does not include the reps of `on-stack?`.

**Comment by mrjbq7** [comment]
I have no idea what the best name should be, I leave that to you!

File: basis/cpu/x86/64/unix/unix.factor (hunk 3 above)
Comment (kusumotonorio) [reply basis/cpu/x86/64/unix/unix.factor]: If the structure is larger than 16 bytes, Factor doesn't need to `record-reg-reps`, so change the above code.
```Factor
M: x86.64 flatten-struct-type ( c-type -- seq )
    dup heap-size 16 <=
    [ flatten-small-struct record-reg-reps ] [
        call-next-method unrecord-reg-reps
        [ first t f 3array ] map
    ] if ;
```

kusumotonorio:
After all, I couldn't think of it either. :-)

PR #2237, opened 2020-01-29
Fix linux find-so failure if no old ld entries exist

doublec:
If '/etc/ld.so.cache' does not contain any old entries
using the 'HeaderOld' struct then it fails with a bad-magic
error before looking for the HeaderNew entries.

My Ubuntu 19.10 system doesn't have any old entries so always
fails here when using 'load-all', which results in the
'extra/llvm' vocab failing due to using 'find-so'.

The fix implemented here is to catch the error and recover from
it by seeking back to the start of the header before looking for
the new header entries.

### DIFF
--- a/extra/ldcache/ldcache.factor
+++ b/extra/ldcache/ldcache.factor
@@ -1,8 +1,9 @@
 ! Copyright (C) 2017 Björn Lindqvist.
 ! See http://factorcode.org/license.txt for BSD license.
 USING: accessors alien.c-types alien.strings assocs byte-arrays
-classes.struct fry io io.encodings.binary io.encodings.string io.files
-kernel math math.bitwise sequences system ;
+classes.struct continuations fry io io.encodings.binary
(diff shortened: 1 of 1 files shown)


mrjbq7:
Nice!!

timor:
Since this is one of the issues I had to work around on NixOS: What is the rationale behind parsing/interpreting `/etc/ld.so.cache` from factor?

mrjbq7:
I wanted a cross-platform way to find the actual library (absolute path) that would be used for a given name. 

It morphed a little bit and lost the absolute path ness and needs to be adjusted a bit. And it shouldn’t be relied on to bake absolute paths into images or the image might get stale versus the OS or library version. 

It wasn’t (originally) intended to do this:

“Foo” find-so add-library 

And I think it’s not great for that use case, since you should just let the OS find it.

timor:
@mrjbq7 Thanks!

PR #2245, opened 2020-03-01
math.matrices: fix/rename mnorm, update all norms

### LINKED ISSUE #2244: Using mnorm on matrix with negative-only entries causes div-by-zero error
If i run
```factor
USE: math.matrices
{ { -1 0 } { 0 0 } } mnorm
```
I get a division by zero error.  Is this word not supposed to be defined for matrices having negative values?

catb0t:
closes #2244

- `mnorm` has been renamed to `normalize-matrix`
	to reflect what it actually does, which
	is normalize a matrix, not find a norm
	of a matrix.

- `mnorm` is no longer a word defined here.

- bugfix: previously, `normalize-matrix` found
	the supremum of a matrix (`mmax`),
	before taking the supremum's absolute
	value (`abs`) and dividing the matrix
	by it (`m/n`).
	for matrices containing only negative
	values and 0, the supremum is 0, and
	a `div-by-zero` error was thrown.

	`normalize-matrix` has been fixed to
	first `abs` all the matrix elements,
	and then find the supremum and divide,

	it also receieved a zero-matrix? guard
	for optimization and preventing
	`div-by-zero`.

- new alias: `hilbert-schmidt-norm` for
	`frobenius-norm`,  to go along with
	`math.matrices.extras.<hilbert-matrix>`
	and improve searchability by physicists.

- new word: `matrix-p-norm`, written as an
	analogue of `math.vectors.p-norm`.

- new word: `matrix-p-q-norm`, which generalizes
	entrywise matrix norm over the L^p,q
	vector space.

- new word: `matrix-p-norm-entrywise`:
	`matrix-p-norm`'s fallback
	for p =/= 1, 2, inf; analogue of
	`math.vectors.p-norm-default`.

- all norm words have gotten new docs,
	`zero-matrix?` guards as an optimisation,
	and most have gotten new tests.

### DIFF
--- a/basis/math/matrices/matrices-docs.factor
+++ b/basis/math/matrices/matrices-docs.factor
@@ -1,4 +1,4 @@
-! Copyright (C) 2005, 2010, 2018 Slava Pestov, Joe Groff, and Cat Stevens.
+! Copyright (C) 2005, 2010, 2018, 2020 Slava Pestov, Joe Groff, and Cat Stevens.
 USING: accessors arrays assocs generic.single formatting locals help.markup help.markup.private help.syntax io
 kernel math math.functions math.order math.ratios math.vectors opengl.gl prettyprint
 sequences sequences.generalizations urls ;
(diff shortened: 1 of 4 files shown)


catb0t:
I also removed some redundant matrices tests that I had accidentally duplicated previously.

PR #2246, opened 2020-03-01
factlog, subset-Prolog-like DSL running on Factor

kusumotonorio:
factlog is an embedded language that runs on Factor with the capabilities of a subset of Prolog.

You can read about it on the following page:
https://github.com/kusumotonorio/factlog


catb0t:
Do you plan to make the Factor repo the main home for factlog, and updates to it will be contributed here directly? 

Or will you maintain it as a separate repo which you periodically make a PR like this?

I hope you maintain it  out-of-tree, just because the `README.md` is brilliant documentation. Though it could be written as an article in factlog-docs...

kusumotonorio:
I will probably make many experimental changes. For this reason, it is better for the users of factlog to keep the code in Factor separate from my repository.

As you may have noticed, the factlog-doc contains almost as good as the README. ;-)

mrjbq7:
Do you think we should call it the “prolog” vocabulary? 

Our historical approach to naming has been to use the most general/descriptive name for things.

mrjbq7:
Or maybe "logic" or something similar noun?

kusumotonorio:
I don't object to renaming the factlog in the standard library if the Factor community wants to. But there are several reasons why I have avoided calling it "prolog".

- I think "prolog" already existed in the past.
https://code-factor.blogspot.com/search?q=prolog
> Improve the Prolog implementation in extra/prolog

- If you really want to do practical logic programming in Factor, I think it's better to interface with a real Plolog and communicate with it. And it is more suitable for the name of "prolog".

- factlog is not as powerful as Prolog, and it is similar to Prolog but requires a different description.

timor:
How about `logic.unification`?

kusumotonorio:
Isn't it a little longer?😀

So, do we call the vocabulary "logic"? I like the name "factlog" @ArnautDaniel gave it, I wish I could leave it in some form ...

kusumotonorio:
I'll call this DSL "factlog" and rename the vocab to "logic" to make it happen.
In principle, remove the word "factlog" from the "logic" library. That way it is "logic" for the user and my project name can remain "factlog".

Please give me some time for such modification.

kusumotonorio:
I did it! ;-)

mrjbq7:
Thanks!!

Merged, and made a couple small fixes in ab81a957c9524e8cc991d3ae915c9f44366ff61b.

kusumotonorio:
Thank you!

PR #2248, opened 2020-03-04
Rename and genericise norm-related words in vectors and matrices

catb0t:
this rename-only changeset causes the following test to fail in `math.vectors.simd`: https://github.com/factor/factor/blob/3efe23cc7171dbb6d6308b0103c338c487e6ae82/basis/math/vectors/simd/simd-tests.factor#L18
it fails because `[ { float-4 } declare norm ] final-classes` is `{ object }` (even after a rebuild) with my changes, even though I didn't change anything type-related, just making it a generic word `l2-norm`. 

But `v+` is declared the same way, `GENERIC: v+ ( v x -- w ) M: object v+ ... ;` and it passes a similar `final-classes` test.

---

After modernising the matrices interface for matrix norms, I realised a rename and genericising of `math.vectors.norm` and related words is sort of an obvious next step. This commit is still kinda silly anyway.

This changeset, when finished, will make `l1-norm`, `l2-norm`, `l-infinity-norm`, and `p-norm` into a set of generic words defined for `object` (in `math.vectors`, as with its other generics), `matrix`, and `zero-matrix` (in `.matrices`).

It also makes the matrix-specific implementations public, so they can be used directly, as `matrix-p-norm` etc.

Finally, the vectors and matrices docs have been updated, and as `norm` is now `l2-norm`, it has been renamed everywhere. `matrix-normalize` is also renamed to match the growing suite of `matrix-` words (such as `matrix-except`).


mrjbq7:
Is it possible we shouldn’t rename math.vectors::norm?

http://mathonline.wikidot.com/the-norm-of-a-vector

catb0t:
It's very possible, but it is a little ambiguous that we (and math as a whole) have dozens of different calculations meant by the simple name "norm" without any qualifiers.

If it's not renamed, should `l2-norm` in matrices be called/aliased to `norm` too?

erg:
Thank you for working on the matrices vocabularies!

catb0t:
@mrjbq7 I've essentially just added `ALIAS: norm l2-norm` so that `norm` is an alias for the generic `l2-norm`, which makes the behaviour / appearance the same as before (other than that test still failing because `norm` is generic now, but wasn't before). 

I also changed all the uses of `norm` back, from `l2-norm` everywhere. Does that make sense, or would you prefer it to be organised in a more plain-`norm` centric way?

mrjbq7:
Well that didn't work.  Sorry @catb0t I seem to have resolved a conflict in your branch with a merge commit.  Silly github.

Let me try again.

mrjbq7:
It looks like the generic definition for ``matrix`` (which is "all sequences with a predicate") is messing up inlining dispatch on ``simd-128`` types for ``l2-norm``.

I wonder if we should make matrix a mixin-class or a hard type like this?

```
TUPLE: matrix < array ;
```

Thoughts @catb0t ?

catb0t:
> It looks like the generic definition for `matrix` (which is "all sequences with a predicate") is messing up inlining dispatch on `simd-128` types for `l2-norm`.
> 
> I wonder if we should make matrix a mixin-class or a hard type like this?
> 
> ```
> TUPLE: matrix < array ;
> ```
> 
> Or even a wrapper class:
> 
> ```
> TUPLE: matrix { underlying sequence read-only }
> ```
> 
> Thoughts @catb0t ?

I never replied because I don't really know the tradeoffs, I think a wrapper class well-written could keep this vocab and user code as simple as possible while keeping SIMD support / performance.

catb0t:
Thanks for pulling this!

PR #2255, opened 2020-03-12
mirrors: delete-at and clear-assoc are an error, fix #1757

### LINKED ISSUE #1757: Difference between mirror clear-assoc and delete-at
I just noticed that in the ``mirrors`` vocabulary, the``M\ mirror delete-at`` sets a slot to ``f``:

```factor
M: mirror delete-at ( key mirror -- )
    [ f ] 2dip set-at ;
```

But  the ``M\ mirror clear-assoc`` sets the slot to its initial value:

```factor
M: mirror clear-assoc ( mirror -- )
    [ object-slots ] [ object>> ] bi '[
        [ initial>> ] [ offset>> _ swap set-slot ] bi
    ] each ;
```

Probably those should be consistent?

catb0t:
`M\ mirror delete-at` and `M\ mirror clear-assoc` have been made to throw a new `mirror-slot-removal` error, because it doesn't make sense to remove a tuple slot, and this behaviour should not have been relied on.

I don't think this breaks anything else but I'm currently loading and testing all vocabs.

closes #1757

### DIFF
--- a/basis/mirrors/mirrors-tests.factor
+++ b/basis/mirrors/mirrors-tests.factor
@@ -49,9 +49,8 @@ TUPLE: color
 { green integer }
 { blue integer } ;
 
-{ T{ color f 0 0 0 } } [
-    1 2 3 color boa [ <mirror> clear-assoc ] keep
-] unit-test
+[ \ + make-mirror clear-assoc ] must-fail
+[ \ + make-mirror [ "name" ] dip delete-at ] must-fail
 
 ! Test reshaping with a mirror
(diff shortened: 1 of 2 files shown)


mrjbq7:
Would it be better to throw an error, or to not implement the words?

catb0t:
leaving it unimplemented doesn't make it as obvious why it's not implemented. I suppose I could just document that mirror slots can't be removed but the error seems like a  good specific explanation.

mrjbq7:
Merged, thanks!

PR #2260, opened 2020-03-15
fix various typos; cleanup and fully document boyer-moore

catb0t:
Some old uncommitted changes I found lying around: fixing minor typos and using the Factor convention of full-word-names instead of abbreviations like `bc` to look nicer.


mrjbq7:
Nice!  Thanks!

PR #2271, opened 2020-04-01
rosetta-code.multisplit: new vocab

AlexIljin:
A rosetta-code solution created for https://rosettacode.org/wiki/Multisplit


mrjbq7:
Thanks!

jonenst:
Ideally, we would scan the input string only once instead of once per separator.
Of course, we could also use a fancy string search algorithm like Boyer Moore but one that supports multiple patterns..

PR #2279, opened 2020-04-13
Some contributions

AlexIljin:
Not sure how `windows.version` could be meaningfully unit-tested without adding a file with known VersionInformation resource to the repository.

### DIFF
--- a/extra/drive-strings/drive-strings.factor
+++ b/extra/drive-strings/drive-strings.factor
@@ -1,0 +1,9 @@
+! Copyright (C) 2019 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: byte-arrays io.encodings.string io.encodings.utf16
+kernel math sequences splitting windows.kernel32 ;
+IN: drive-strings
+
+: logical-drive-strings ( -- seq )
+    30 4 2 * * dup <byte-array> [ GetLogicalDriveStrings ] keep
+    utf16le decode swap head "\0" split harvest ;
--- a/basis/alien/data/data-docs.factor
+++ b/basis/alien/data/data-docs.factor
@@ -257,4 +257,4 @@ ARTICLE: "c-out-params" "Output parameters in C"
 { $code
   "1234 { c-string } [ do_frob ] with-out-parameters"
 }
-"which would put the functions return value and error string on the stack." ;
+"which would put the function's return value and error string on the stack." ;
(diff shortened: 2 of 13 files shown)


extra/drive-strings/drive-strings.factor
```diff
@@ -0,0 +1,9 @@
+! Copyright (C) 2019 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: byte-arrays io.encodings.string io.encodings.utf16
+kernel math sequences splitting windows.kernel32 ;
+IN: drive-strings
```
mrjbq7, line 5: I'm not sure about this vocab name, or if it shouldn't be part under ``windows`` vocabulary...

mrjbq7:
Merged, thanks!

We can figure out ``drive-strings`` later.

extra/drive-strings/drive-strings.factor
AlexIljin: Good idea, I'll move it under `windows.`.

AlexIljin: Ah, you've already merged it.

mrjbq7: It's all good -- being in ``extra`` makes me stress less about it.

mrjbq7: I could move it to ``windows.drive-arrays`` if you think that's better.

AlexIljin: Yes, it totally makes sense. There are no drive letters on other platforms, and this is specifically a demo/wrapper for a WinAPI function.
I should have treated it same as the `windows.version` vocab here.

AlexIljin: Except I don't really like `drive-arrays`, better `drive-strings` or simply `drives`. But I leave it up to you.

mrjbq7: I meant ``windows.drive-strings``, that was a typo!

Moved in 1e81dbdf17281dfcacaad7a1e685b3cdaf147fd3.

AlexIljin: You mean "copied", not "moved" : ))

mrjbq7: Oh geez.  Maybe my brain is on lockdown too.

Non-atomic move completed in a89474786edba391abd6da6e5d3ec9e7a4daacb6.

PR #2280, opened 2020-04-13
Doc fixes

AlexIljin:
There is one commit that fixes the "delete to Recycling Bin" function on Windows, the rest are documentation beautifications.

### DIFF
--- a/extra/io/files/trash/windows/windows.factor
+++ b/extra/io/files/trash/windows/windows.factor
@@ -2,8 +2,9 @@
 ! See http://factorcode.org/license.txt for BSD license
 
 USING: accessors alien.c-types alien.data alien.strings
-alien.syntax classes.struct destructors io.files.trash kernel
-libc literals math sequences system windows.types ;
+alien.syntax classes.struct destructors io.files.trash
+io.pathnames kernel libc literals math sequences system
+windows.types ;
 
 IN: io.files.trash.windows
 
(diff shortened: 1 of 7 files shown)


extra/io/files/trash/windows/windows.factor
```diff
@@ -51,6 +52,7 @@ PRIVATE>
 
 M: windows send-to-trash ( path -- )
     [
+        absolute-path [ dup CHAR: / = [ drop CHAR: \\ ] when ] map
```
mrjbq7, line 55: Perhaps we should call ``normalize-path`` here instead?

AlexIljin: I avoided `normalize-path`, because it returns a UNC path, and the documentation says that "SHFileOperation fails on any path prefixed with "\?"."

Actually, please skip this commit: it crashes for me at the moment with a memory protection fault - the original version, the one I submitted in this PR and the one with `normalize-path`, they all crash.

AlexIljin: https://docs.microsoft.com/en-us/windows/win32/api/shellapi/ns-shellapi-shfileopstructa

mrjbq7:
I cherry-picked all the commits except the ``io.files.trash`` one.

If you could fix the memory protection fault with it and the normalize-path or absolute-path that would be great!  I thought it worked at some point.  It's maybe a little hard to test.

AlexIljin:
> I cherry-picked all the commits except the `io.files.trash` one.
> 
> If you could fix the memory protection fault with it and the normalize-path or absolute-path that would be great! I thought it worked at some point. It's maybe a little hard to test.

Submitted in PR#2281.

PR #2281, opened 2020-04-15
io.files.trash.windows: fix the `send-to-trash` word

AlexIljin:


mrjbq7:
Won’t that test fill up your recycle bin every time it’s run?

AlexIljin:
Yes, but the files are zero length, and it's just one file per test.
Frankly, I never run the full test suite, so I don't mind, and the build machines can clear the Bin as part of their work.
Maybe we could somehow tag this test as manual-run only.

mrjbq7:
Ok. Could you make that test for io.files.trash? Nothing about it is windows specific. 

What does windows do if you trash a file and then hard delete the directory It was in? If that’s handled well, then we could use with-test-directory and run your test code inside it.  

Thoughts?

AlexIljin:
It is handled well in the sense that the file will stay in the Trash Bin after the folder is deleted.
`with-test-directory` is a good idea, updated.

extra/io/files/trash/trash-tests.factor
```diff
@@ -0,0 +1,10 @@
+! Copyright (C) 2020 Alexander Ilin.
+! See http://factorcode.org/license.txt for BSD license.
+USING: io.directories io.files.temp io.files.trash kernel tools.test ;
+IN: io.files.trash.tests
+
+{ } [
+    ! temp-file is not used here, because it returns the absolute path, and we
+    ! want to ensure send-to-trash works without giving it the full path.
+    [ "io.files.trash-tests" dup touch-file send-to-trash ] with-temp-directory
```
mrjbq7, line 9: I think you mean ``with-test-directory``, which cleans up after itself?

AlexIljin: Right! Updated.

mrjbq7:
Merged, thanks!

PR #2289, opened 2020-05-14
Add imenu tags to factor-mode for Emacs

Mercerenies:
[Imenu](https://www.emacswiki.org/emacs/ImenuMode) is a useful feature in Emacs for jumping to definitions in a given source file. This PR adds support for Imenu tags to `factor-mode`, enabling users to jump to various sorts of definitions in Factor code.


mrjbq7:
Cool!

mrjbq7:
Thanks!!

PR #2290, opened 2020-05-16
io.standard-paths: fix standard-login-paths for fish shell

sfesenko:
Fish shell automatically split variables whose name ends in "PATH" into lists,
and uses space as separator for output
Colons force fish to use standard $PATH representation


mrjbq7:
Nice, thanks!

PR #2293, opened 2020-05-19
compiler.tree.propagation.slots: remove unused word

timor:
The last use of `length-accessor?` has been removed in
8e227bc874e356f1e292ab18d1bad1d48966746a, which obsoleted the `length` slot.


mrjbq7:
Thanks!

PR #2294, opened 2020-05-20
shell.nix: supply `wrapFactor` helper to make standalone factor binary

timor:
This adds the shell function `wrapFactor`.  This function is intended to wrap
the result of calling `build.sh` in the shell environment so it can be executed
outside of the nix shell.

Example:

```
$ nix-shell
[nix-shell] $ ./build.sh bootstrap
...build factor vm and image...
[nix-shell] $ wrapFactor .
exit
$ ./factor
```

`wrapFactor` takes the path to the factor root dir as argument, and expects the
binary `factor` and the image file `factor.image` there and uses Nixpkgs'
`makeWrapper` to wrap the `factor` executable in-place with the correct
`LD_LIBRARY_PATH`.  Afterwards, the factor executable can be called outside of
the nix-shell environment.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -15,13 +15,25 @@ let
     udis86 # available since NixOS 19.09
     openal
   ];
+  runtimeLibPath = lib.makeLibraryPath runtimeLibs;
 in
 (mkClangShell {
   name = "factor-shell-env";
-  LD_LIBRARY_PATH = "/run/opengl-driver/lib:${lib.makeLibraryPath runtimeLibs}" ;
+  LD_LIBRARY_PATH = "/run/opengl-driver/lib:${runtimeLibPath}" ;
   buildInputs = runtimeLibs ++ [
     # for building factor
     git
     curl
+    makeWrapper
   ];
+  shellHook = ''
+    wrapFactor () {
+    [ -n "$1" ] || { printf "Usage: wrapFactor <factor-root>" ; return; }
+    local root="$(realpath $1)"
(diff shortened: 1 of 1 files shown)


timor:
cc @kwshi @spacefrogg

shell.nix
```diff
@@ -15,13 +15,25 @@ let
...
+    makeWrapper
   ];
+  shellHook = ''
+    wrapFactor () {
+    [ -n "$1" ] || { printf "Usage: wrapFactor <factor-root>" ; return; }
+    local root="$(realpath $1)"
+    local binary="''${root}/factor"
+    wrapProgram "$binary" --prefix LD_LIBRARY_PATH : ${runtimeLibPath} \
+      --argv0 factor
+    ln -sf "''${root}/factor.image" "''${root}/.factor-wrapped.image"
```
mrjbq7, line 36: What's this symlink for?

spacefrogg: As far as I know, the factor binary searches for an image with the same basename next to it. `wrapProgram` moves the original `$binary` to `.${binary}-wrapped` and establishes a wrapper with the name `$binary`. Thus, I expect the wrapped factor binary to search for a wrapped-named factor image.

I'm just supplying the intent of this code. I don't know if any of my assumptions are correct.

mrjbq7:
Thanks!

shell.nix
timor: Yes, that is the reason for the symlink.

PR #2299, opened 2020-06-06
Vim syntax highlighting overhaul

bb010g:
Mostly, I wanted comments to work properly. Relevant notes from writing & rewriting these patches were left as comments in `syntax/factor.vim` for future work.

Of note is the generated syntax content moving to `syntax/factor/generated.vim`, which makes iterating on everything else in `factor.vim` a lot easier, and also clarifies how users can extend `factorWord` keywords out-of-tree. This file is also now properly marked as generated to GitHub's [linguist](https://github.com/github/linguist).

Also, the minimum Vim version was bumped up to 6, but I think that's reasonable. Pre-6 patterns couldn't match multiple lines.

---
This change is [Reviewable](https://reviewable.io/reviews/factor/factor/2299)


misc/vim/ftdetect/factor.vim
```diff
@@ -1 +1,3 @@
-autocmd BufRead,BufNewFile *.factor,{,.}factor*-rc set filetype=factor
+" autocmd BufRead,BufNewFile *.factor setf factor " upstreamed
```
mrjbq7, line 1: why is this commented out?

misc/factor.vim.fgen
```diff
@@ -1,333 +1,98 @@
-<%
-USING: accessors io kernel multiline prettyprint sequences sorting vocabs ;
+USING: accessors
+    calendar calendar.format
+    io kernel multiline prettyprint sequences sorting
+    splitting tr vocabs ;
 IN: factor.vim.fgen
+ALIAS: wr write
```
mrjbq7, line 6: Why not just use ``write``?  Is there a name conflict somehow?

mrjbq7:
If you want a couple more things that would be great to add:

1) Binary number literals:

```
0b10101
```

2) Multiline comments:

```
/* comment */
/* multline ( x -- y )
  2  comment */
 6 /* something else */ 2 +
```

3) Named unicode characters:

```
CHAR: snowman
```

4) Regexp strings (including with options)

```
R/ USING: [^;]* ;/s 
```

5) Predicate classes

```
PREDICATE: class < superclass predicate... ;
```

6) Alien FFI constructs:

```
    LIBRARY: name
    TYPEDEF: old new
    ENUM: type words... ;
    ENUM: type < base-type words...
    FUNCTION: return name ( parameters ) ;
    FUNCTION-ALIAS: factor-name return name ( parameters ) ;
```

mrjbq7:
Thanks!!

misc/vim/ftdetect/factor.vim
bb010g: https://github.com/vim/vim/commit/520470a9d6afa273294848bc31e9ddfabe130434#diff-44a099e40ae628aef90a28c6408a4c61R546-R548

https://github.com/vim/vim/blob/975a880a1389e8ce6dea8d66a7c109140b2f94ec/runtime/filetype.vim#L583-L584

I figured to leave it in though for demonstrating where you'd handle new (non-`factor-doc`) Factor file extensions, if necessary.

misc/factor.vim.fgen
bb010g: It's used a lot, and helps keep a non-trivial amount of lines `textwidth=64` compliant.

bb010g:
1. Binary number literals are `factorBin`. They don't have an error class, though, like `factorHex`.
2. Multiline comments are `@factorMultilineComment`; their caveats are documented and handled properly (mainly, they're words, not lexer syntax). This is somewhere where syntax highlighting being reliably correct is really valuable—someone just starting out probably would expect `USE: /* foo */ bar` to use a vocabulary and not complain that `Vocab name cannot contain ":/\ "`.
3. `CHAR:` I believe handles better than before now. Unicode should work fine just by Vim's multibyte compliance in patterns.
4. `R/` regexp strings aren't in there. Those look like fun. 😅
5. `PREDICATE:` sounds decently easy.
6. Planning to drop those in soon too; I actually started on this because I wanted to comment after some alien `STRUCT:` body lines and the lack of proper coloring (the *whole* struct declaration was bright, bold green for me, *including* comments) drove me crazy.

mrjbq7:
For some reason binary number literals aren't highlighting in my VIM.  Do they work for you?

The character "long names" don't highlight (only the first character does for me):

```
CHAR: snowman
```

not

```
CHAR: ☃
```

mrjbq7:
Here's some things that highlight as a number and shouldn't:

```
,0.1
.
-.
```

And these stack effects don't highlight properly:

```
    ( :integer -- :float )
    ( quot: ( a -- b ) -- )
    ( x quot: ( a -- b ) -- y )
    ( ..a quot: ( ..a x -- ..b ) -- ..b )
```

mrjbq7:
If you have any energy to work on these, I'd appreciate it!

But no worries, if you don't!

bb010g:
> The character "long names" don't highlight (only the first character does for me):
>
> ```
> CHAR: snowman
> ```

Ah. Didn't realize those existed.

Numbers are fixed (overeager `factorFloat`s).

Stack effects:
* `( :integer -- :float )`: This has the problem of running face-first into our prevalent use of `<` at the beginning of patterns to ensure the beginnings of words are at the beginning of words. Placing a `:` before one means that you're now not at the beginning of a word. We can dodge this with stack effects thanks to them not being…everything else, but they're an exception to the rule.

  https://github.com/factor/factor/blob/14b1418f6a0c76397d0030c548408c56b4031163/misc/vim/syntax/factor.vim#L282-L285

* `( quot: ( a -- b ) -- )`: Wrote `/v` instead of `/\v`. This was annoyingly long to catch.
* `( x quot: ( a -- b ) -- y )`: Same bug.
* `( ..a quot: ( ..a x -- ..b ) -- ..b )`: Same bug.

I'll open a new PR in a bit after updating `/misc/vim/README.md`.

mrjbq7:
Fixed a couple things in ddf498d5ad1db39137c7892ff32d7795678943a6.

I added a ``misc/syntax-test.factor`` which has a bunch of random syntax to verify highlighting.

Feel free to add forms or modify if it strikes your fancy.

mrjbq7:
Should we move ``misc/factor.vim.fgen`` to something like ``misc/vim/generate.factor``?

mrjbq7:
(I think I might be wrong to re-order the factorBin, factorHex forms, but it fixed an error with NAN: highlighting.)

bb010g:
Followed up in #2300; still have to rebase.

PR #2301, opened 2020-06-07
fixup! [misc] vim/syntax: Fixups

bb010g:
i'm bad at titling pull requests edition

---
This change is [Reviewable](https://reviewable.io/reviews/factor/factor/2301)

PR #2303, opened 2020-06-09
vim/syntax: Even more fixups.

bb010g:
In which the issues brought up in #2301 are addressed. I really don't see numbers having any more issues from here onwards.

`|:syn-priority|` is respected now, `:syn-skip` & `:syn-keepend` are used when appropriate, newlines don't jank stuff up, comments don't extend match regions, numbers are much more reliable, and stack effect error highlights return.

A feature request has even been sent to Bram.  
https://github.com/vim/vim/issues/872#issuecomment-641025231

---
This change is [Reviewable](https://reviewable.io/reviews/factor/factor/2303)

### DIFF
--- a/misc/syntax-test.factor
+++ b/misc/syntax-test.factor
@@ -49,16 +49,40 @@
 ! Definitions
 
     : word ( x -- ) drop ;
+    : word error drop ;
     :: word ( x -- ) x drop ;
     TYPED: word ( a b: class ... -- x: class y ... ) body ;
     TYPED:: word ( a b: class ... -- x: class y ... ) body ;
     MACRO: word ( inputs... -- ) definition... ) ;
     MACRO:: word ( vars... -- outputs... ) definition... ) ;
-    M: class generic (definition) ... ;
+    M: object explain drop "an object" print ;
+    M: class generic definition... ;
     M:: class generic ( vars... -- outputs... ) body... ;
+    M:: class generic error body... ;
     GENERIC: word ( stack -- effect )
+    GENERIC: word
+        ( stack -- effect )
+    GENERIC: word
+( stack -- effect )
+    GENERIC: word ! comment
+        ( stack -- effect )
+    GENERIC: word drop ! 3rd token wrong
+    GENERIC: word ! next line wrong
+        drop
+    GENERIC: word
+drop ! wrong
     HOOK: word variable ( stack -- effect )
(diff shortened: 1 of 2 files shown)


misc/syntax-test.factor
```diff
@@ -226,12 +250,25 @@ PRIVATE>
...
     0x1AB4p30
-    0b10101
-    0o1234567
-    NAN: CAFE1234
+    0B10101
+    0O1234567
+    NAN: CAFE1234 0,. ! third token wrong
+    0,. ! wrong, next line also wrong
+    0,.
+    NAN: ! ff 0xff comment
```
bb010g, line 260: I'm glad to see that GitHub's syntax highlighting has been thrown for a loop.

bb010g:
Umm, don't merge this yet, just noticed a regression from rebasing.

mrjbq7:
Thanks!

mrjbq7:
Do you know if there's a way in Vim to change all Typedef colors inside of <PRIVATE PRIVATE> so we don't have to repeat ourselves for all the possible syntax that might exist but be in a private vocab?

mrjbq7:
A couple of minor questions, if you don't mind:

1. I used to be able to hit `*` and have it jump to the next instance of the word under the cursor.  Now it seems to do some kind of "head" search, so if you are at (for example) `gl-line` it will jump to `gl-color`.

2. Tuple definitions seem to be highlighting every other slot differently:

```factor
TUPLE: foo < bar every other slot name ;
```

bb010g:
I've got a fix for tuples locally that I'll merge after cleaning it out of my current WIP commit.

---

As for `*`, that's probably due to the syntax file setting `:syn-iskeyword` instead of `'iskeyword'` now.

```vimhelp
	It is recommended when writing syntax files, to use this command to
	set the correct value for the specific syntax language and not change
	the 'iskeyword' option.
```

So, we could set this in the ftplugin or leave it for the user to customize? It looks like normally this is set in ftplugin, if it's set.

mrjbq7:
@bb010g just checking in on the fixes you mention above?

mrjbq7:
Hi @bb010g I pushed a fix for '*', want to send me your fix for tuples?

bb010g:
Sorry about the huge delay. (Suddenly stopped working with Factor for a while.) Just pushed the relevant patch from my current working directory: #2333

PR #2318, opened 2020-07-29
TOTP

AlexIljin:
Updated the Time-based One-Time Passwords vocab to fully mimic the [Google Authenticator](https://github.com/google/google-authenticator-android) app. Now if I don't have the app on my phone for some reason, I can use simply use Factor.


mrjbq7:
Thanks!

PR #2319, opened 2020-07-29
Checksums

AlexIljin:
I'm not sure why the `checksum-state` was not inherited from the `disposable` in the first place, but it seems to make a lot of sense, especially since implementing `dispose` is a requirement.

I didn't run the full test suite on this, so feedback and comments are welcome.


mrjbq7:
Looks good to me, thanks!

PR #2327, opened 2020-08-23
contributors: merge the known contributor aliases

AlexIljin:
What do you think about this?
Without this change the `contributors` vocab is kind of generic and can be used on any git repository, and with this change it's much more specific to the Factor repo. I'm not sure it the change is in line with the original intention behind the vocab.
On the other hand, if it's used on an unrelated repository, there is very little chance that the antialiasing will make any difference at all. It'll essentially be a no-op on almost any other repo in the world.

### DIFF
--- a/core/assocs/assocs-docs.factor
+++ b/core/assocs/assocs-docs.factor
@@ -409,6 +409,8 @@ HELP: delete-at*
 { $description "Removes an entry from the assoc and outputs the previous value together with a boolean indicating whether it was present." }
 { $side-effects "assoc" } ;
 
+{ delete-at delete-at* ?delete-at } related-words
+
 HELP: rename-at
 { $values { "newkey" object } { "key" object } { "assoc" assoc } }
 { $description "Removes the values associated to " { $snippet "key" } " and re-adds it as " { $snippet "newkey" } ". Does nothing if the assoc does not contain " { $snippet "key" } "." }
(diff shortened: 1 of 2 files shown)


mrjbq7:
Not sure! I can see it's usefulness for our repo though.

Btw, instead of ``extract-keys``, you should maybe use ``delete-at*`` or ``?delete-at``.  I'm not quite sure why we need both of those, but one of them should serve your purpose.

Using something like (untested):

```factor
unclip '[ over delete-at* [ _ pick at+ ] with when ] each
```

AlexIljin:
I most certainly **would** use one of those, if they were a bit more discoverable. Reading through the docs I didn't see those words exist, because the help topic lists `delete-at`, and there are no pointers to the other variants!

mrjbq7:
Yeah, I didn’t realize they existed either although I thought they *should* exist. Not sure why both variants are necessary.

AlexIljin:
I see the only difference seems to be in using `at*` vs. `?at`, the latter outputs the key if it was not found in the assoc.

AlexIljin:
Tested and pushed the newer version.

AlexIljin:
It seems that in the code you suggested there is a weird interaction between `fry` and `with`.

AlexIljin:
See issue #2328 about the weird interaction.

AlexIljin:
I have nothing to add to this PR.

mrjbq7:
Thanks!

PR #2331, opened 2020-09-13
google.charts: use quickchart.io instead of defunct google chart api

sfesenko:


sfesenko:
it seems, that https://quickchart.io/ is not 100% compatible with google chat api, so it cannot be used as drop in replacement, and it would be better to have separate integration

PR #2352, opened 2020-10-17
compiler.propagation.call-effect: handle class-info-only value-infos

### LINKED ISSUE #2351: Compilation error for declared inputs to call(
Problem:
```
IN: scratchpad USE: kernel.private : blub ( x -- ) { composed } declare call( -- ) ;
```
fails with
```
Cannot create slice: from < 0
from -2
to   0
seq  f
```

I would expect this to be simply uninferable.  This fails because during propagation.call-effect, `(infer-value)` assumes literal or cloned value info with actual slot info for `curried` and `composed`, and fails if only class info is given.

timor:
Checks if the `slots>>` value-info slot is non-empty before trying to infer
effect from non-literal value-info.

Fixes #2351


mrjbq7:
Thanks!!

PR #2353, opened 2020-10-23
JSON reader improvements

Abtin-Molavi:
Some changes to the json.reader vocabulary in order to throw errors in more cases where the input is invalid JSON syntax. Also additional tests were added to reader-tests in order to test this functionality.

Note: these changes cause an existing test on an empty input (which has been removed as part of this pull request) to fail because the parser now throws an error rather than returning an empty hash table. This is consistent with the behavior of other implementations (e.g. Python).


mrjbq7:
Awesome, thanks!!  Merged!

PR #2354, opened 2020-10-29
Extending Url Parsing

DMind44:
Added parsing for missing authority, as well as empty ports, usernames, and passwords.

Added tests for the above features.


erg:
Thank you!

PR #2358, opened 2020-11-03
urls: Don't escape the path and then url-encode it.

erg:
Add a test case where it breaks without this patch.

Basically %2f and / are different and we can't encode/decode every one we see.


erg:
This isn't the right fix.

PR #2359, opened 2020-11-03
peg.ebnf: Allow ] in character classes.

erg:


erg:
It's not quite right.

```
[]]   ! match ] only
[^]] ! match not ]
[^]  ! syntax error
[^]\]] ! match ] only
[^]a] ! not ] and not a
[^a\]] ! not a and not ]
```

erg:
Pushed.

PR #2365, opened 2020-11-07
Doc fixes

AlexIljin:
I always forget how to use the `backtrack` vocab, so after rediscovering that today for the third time I've decided to update its documentation.


mrjbq7:
Thanks!!

PR #2366, opened 2020-11-08
images.viewer: Use superclass implementation for model activation/deactivation

timor:
`image-contol` seems to duplicate functionality from `notify` in their `graft*`/`ungraft*` methods with `start-control` and `stop-control`.  If I understood the source correctly, this was supposed to be public API, and `activate-control`/`deactivate-control` look like they implement that same behavior with better updating, so made those public API instead.

### DIFF
--- a/basis/ui/gadgets/gadgets-docs.factor
+++ b/basis/ui/gadgets/gadgets-docs.factor
@@ -205,6 +205,18 @@ HELP: ui-notify-flag
 { $var-description "A " { $link flag } " raised to notify the UI thread that there is work to do." }
 { $see-also notify-ui-thread } ;
 
+HELP: activate-control
+{ $values
+  { "gadget" gadget }
+}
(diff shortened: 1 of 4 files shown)


mrjbq7:
I merged your first commit, but I'm unsure about making activate-control public.

For now, I remove stop-control and start-control with the logic that if you don't want to refresh the image, you don't have the change the model's value... (3655a069db81d27b970a5d498effbd523f239efb).

mrjbq7:
Thanks!

timor:
Activation/deactivation is coupled to grafting right now.  If you have a couple of connected models, they will always update as long as the gadget at the end of the chain is grafted.  I think the idea behind `start-control` and `stop-control` was to have a way to manually enable/disable this, so you would have a general way to manually disable all computations associated with a gadget.

The simple case where you only set the model to an image directly is indeed no problem.

PR #2368, opened 2020-11-13
ui.images: support image tuples as well as image-names

timor:
Allows objects created by `<image>` to be used with `<image-pen>`, using the image-texture-cache.

Example:
```
IN: scratchpad auto-use 1234 { 50 20 } normal-noise-image
IN: scratchpad auto-use <image-pen>
IN: scratchpad auto-use "foo" <label>
IN: scratchpad auto-use swap >>interior
IN: scratchpad auto-use gadget.
```


mrjbq7:
Thanks!

timor:
> Thanks!

Wow, that was quick!  Did you have a look at it? There aren't any tests for the affected vocabs yet...

PR #2370, opened 2020-11-14
models: Add word to force recomputation of model values

timor:
`compute-model` can be used to force evaluation of a model.  This is useful during development, but also if you want to use models outside of gadget implementations.

One use I found for me is when using models to compute things like gadget positions, but needing the value before they are grafted.


mrjbq7:
It seems like you only need to activate/deactivate if ref-count is zero?  Otherwise you can just grab the value?

mrjbq7:
I guess, the only difference from what I just mentioned and your approach, is that I was thinking check the current ref value for zero before calling activate/deactivate, and you approach increments the ref value and decrements and only does work in the originally zero case.

mrjbq7:
Thanks @timor !

timor:
> I guess, the only difference from what I just mentioned and your approach, is that I was thinking check the current ref value for zero before calling activate/deactivate, and you approach increments the ref value and decrements and only does work in the originally zero case.

I guess checking `ref>>` is quicker, while the version I PR'd only uses public interface.  Would probably make sense to check `ref>>`, actually.

PR #2371, opened 2020-11-15
Add <?arrow> and <?smart-arrow> for model updates only when quotation values change

timor:
Like `<arrow>` and `<smart-arrow>`, but using `?set-model` instead of `set-model`.

This can save computation if a value data-flow model is needed instead of an event data-flow model.


mrjbq7:
Thanks!

Issue #641, opened 2012-08-23
Implement a more modular module system


jckarter:
The module system should be improved to better handle development outside of the Factor core and enable automated package management. There is some discussion on a proposal at https://gist.github.com/3440892.

lolbinarycat:
I think this seems pretty good, but `.factor` is already a long enough file extension, couldn't we just use `package.yaml` and `module.yaml`? This would also have the benefit of editors like emacs automatically  using the appropriate syntax highlighting.

mrjbq7:
I mean, I'd rather use a simpler format that YAML, like just INI (sort of like the java properties format maybe)...

lolbinarycat:
toml?

mrjbq7:
🤷 

Simpler is better!

Java properties are maybe too simple, but they satisfy the non-nested INI file basically.

Easy to read, easy to write.

lolbinarycat:
toml is basically standardized INI, and a toml parser for factor already exists

AlexIljin:
I'm also for simpler syntax. Ideally, the minimal syntax that does the job would be best, plus a format version indicator in case we need to adjust for something unforeseen.

AlexIljin:
I like the section on the TOML criticism here https://en.wikipedia.org/wiki/TOML
Ini-files are the easiest to support in practise, and TOML overspecifies stuff in my opinion. This may be important for interoperability, but that's not our case.

mrjbq7:
And non-nested simple INI files are basically the simplest to parse without
worrying about specs.

lolbinarycat:
[Here's my draft revision](https://gist.github.com/lolbinarycat/8c22cbfc16b981ec3ebd4d842c950a29)

odiferousmint:
S-expressions!

lolbinarycat:
I actually recently wrote a simple parser for an lisp-like s-expression syntax (although it was in haskell, so would have to be ported). I could try and do that if it's something people are interested in.

PR #2372, opened 2020-11-21
shell.nix: Add more runtime dependencies, fix opengl dependencies and wrapper

timor:
Required for vocabs:
- ogg
- ogg.vorbis
- compression.zlib.ffi

Theoretically, we would need to include every library that has been declared with `LIBRARY:` to make sure all vocabs with ffi work out-of-the box.  This may result in an unacceptable nix closure size though.
Possibly related: #641 

cc @spacefrogg

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -35,6 +35,11 @@ in
     [ -n "$1" ] || { printf "Usage: wrapFactor <factor-root>" ; return; }
     local root="$(realpath $1)"
     local binary="''${root}/factor"
+    local wrapped="''${root}/.factor-wrapped"
+    # Remove the wrapped binary if a new VM has been compiled
+    ${lib.getBin file}/bin/file $binary |grep ELF >/dev/null && rm -f "$wrapped"
+    # Restore the factor binary if it was already wrapped
+    [ -e "$wrapped" ] && { mv "$wrapped" "$binary" ; }
     wrapProgram "$binary" --prefix LD_LIBRARY_PATH : ${runtimeLibPath} \
       --argv0 factor
     ln -sf "''${root}/factor.image" "''${root}/.factor-wrapped.image"
--- a/shell.nix
+++ b/shell.nix
@@ -9,7 +9,8 @@ let
     gdk_pixbuf
     gnome2.gtkglext
     pcre
-    mesa_glu
+    libGL
+    libGLU
     freealut
     openssl
     udis86 # available since NixOS 19.09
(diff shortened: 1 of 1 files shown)


timor:
@spacefrogg
I basically messed up every line of code twice on this PR.  Maybe if you have some time you could take a look at this.

spacefrogg:
As I see it right now, this asks for an interface.
Just like for python or other languages, you should define a set of "packages" that the user can activate in some kine of `factorWithPackages`. Right now, they would not actually draw in a factor vocab but set up the necessary dependencies such that the resulting factor can work with those vocabs. This solves 2 1/2 problems:
- It gives the user a list of viable options and resolves her dependency issues by creating the relation between the factor vocab she wants to use and the necessary setup that has to be performed.
- It prepares factor to actually integrate off-source vocabs.

spacefrogg:
It also shifts the responsibility for the closure size to the user, where it belongs.

timor:
> As I see it right now, this asks for an interface.
> Just like for python or other languages, you should define a set of "packages" that the user can activate in some kind of `factorWithPackages`.
That we already have in the not-yet-upstreamed new factor derivation. The build helper code there still has to be changed to use this repo's shell.nix.

The other question is whether the changes here look correct in fixing the opengl package dependency issues and the changes to the `wrapFactor` helper...

spacefrogg:
It works.

timor:
I used this for a while now.  There is still an impurity problem, which happens when Factor is started from an Emacs that has been built against a different version of glib/gtk.

Nevertheless, it would be nice if this could get merged to fix compatibility with recent NixOS versions.

spacefrogg:
LGTM.

shell.nix
```diff
@@ -31,6 +35,11 @@ in
     [ -n "$1" ] || { printf "Usage: wrapFactor <factor-root>" ; return; }
     local root="$(realpath $1)"
     local binary="''${root}/factor"
+    local wrapped="''${root}/.factor-wrapped"
+    # Remove the wrapped binary if a new VM has been compiled
+    ${lib.getBin file}/bin/file $binary |grep ELF >/dev/null && rm -f "$wrapped"
```
mrjbq7, line 40: does ``$binary`` need to be in quotes in case ``root`` has spaces in it?

timor: I think you are correct.

mrjbq7:
Merged and fixed in 1928e6094ea644b24e511e5c21036729b017f98d.

timor:
Thanks!

PR #2373, opened 2020-11-21
Specialize some of gadget's default implementations on object instead

timor:
This allows implementing mixins which then can specify default behavior for
layouting and grafting for their instances.

Specifically concerns these generics:
- `model-changed`
- `layout*`
- `graft*`
- `ungraft*`

### DIFF
--- a/basis/models/models.factor
+++ b/basis/models/models.factor
@@ -58,6 +58,7 @@ DEFER: remove-connection
     [ activate-model ] [ deactivate-model ] [ value>> ] tri ;
 
 GENERIC: model-changed ( model observer -- )
+M: object model-changed 2drop ;
 
 : add-connection ( observer model -- )
     dup connections>>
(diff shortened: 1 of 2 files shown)


timor:
A different solution for the actual problem (define `MIXIN:`s which specialize grafting/layouting), would be if there was some way of specifying a least class when defining a `MIXIN`, so that subclasses of `gadget` don't always look to `gadget`'s for implementations.

timor:
For me, after running `./build.sh bootstrap` and running Factor, the definitions on `gadget` still existed.  Only after `refresh-all` the new ones were picked up.  Does that make sense or is there a quirk in my setup?

mrjbq7:
Hmm, doing a ``bootstrap`` should result in recompiling the ``ui`` vocabulary which should pull in those changes.

timor:
I tried it again with a clean checkout, and after bootstrap, `M\ gadget graft*` is undefined as expected.  Don't know what that was.  Sorry for the noise.

mrjbq7:
Thanks!

PR #2376, opened 2020-11-22
logic-docs: fix string

kusumotonorio:
Separated the `"` and `{` by a space.


mrjbq7:
Thanks! I think I had already patched that.

It's funny that ``"hello"length`` and ``"hello"{ }`` worked, and recently getting ready for some of the new parser techniques that @erg has been working on had us fixing that, and I was surprised there were so many (8? 12?) places which we had to fix!

PR #2397, opened 2020-12-15
sets: rename combine/refine to union-all/intersect-all

bpollack:
This renames `sets:combine` and `sets:refine` to `sets:union-all` and `sets:intersect-all`, respectively. Renaming `assoc-combine` and `assoc-refine` are left to a separate exercise.


core/sets/sets-docs.factor
```diff
@@ -46,7 +46,7 @@ ARTICLE: "set-operations" "Operations on sets"
 "Operations on groups of sets:"
 { $subsections
     combine
```
mrjbq7, line 48: I still see ``combine`` is some of these diffs?

bpollack: ...that is correct, because I forgot to amend the diff prior to pushing. Fixed.

mrjbq7:
Looks great to me!

@erg do you have any last words for the ``combine/refine`` nomenclature?

PR #2399, opened 2020-12-16
assocs: rename assoc-combine/refine to assoc-union-all/intersect-all

bpollack:
This is the second half of 2abb328052. Now that the first half has landed, I think this second part makes sense.

NB, I'm not really thrilled with the simple words being `assoc-union` and `assoc-intersect`, since it's really more like `assoc-merge` and `assoc-common-keys-only` or something, but at least it's now all consistent.


mrjbq7:
I sort of agree with your name criticism.

We have an ``assoc-merge`` in assocs.extras that does ``push-at`` to combine, where ``assoc-combine`` does ``set-at``...

mrjbq7:
It might be we need to change ``assoc-union`` and ``assoc-intersect`` to something else and then make ``-all`` versions of them...

bpollack:
I prefer the name `assoc-merge` over `assoc-union`, and `assoc-merge-all` makes complete sense as well. That's just leave `assoc-refine`. Need to think on that one.

For what it's worth, annotating the files shows you wrote `assoc-merge` two years before @erg wrote `assoc-union`, and their implementations are so close that I think we should axe one or the other. I don't even think the performance should be meaningfully different, though I'd have to check.

bpollack:
Ooh, what about `assoc-merge` and `assoc-overlap`, with corresponding `assoc-merge-all` and `assoc-overlap-all`? (I appreciate I'm bikeshedding names here, but since you also felt like `assoc-union` and `assoc-intersect` felt weird...)

mrjbq7:
Or maybe ``assoc-collect`` and ``assoc-collect-all``?

mrjbq7:
Merged in 950ead9bb9b24506989910463a4616d93ec5c7ed, thanks!

PR #2409, opened 2020-12-30
Added UTExportedTypeDeclarations to Info.plist

tgunr:
Permit system to identify file extension .factor as text

### DIFF
--- a/Factor.app/Contents/Info.plist
+++ b/Factor.app/Contents/Info.plist
@@ -34,7 +34,7 @@
 	<key>CFBundleVersion</key>
 	<string>0.99</string>
 	<key>NSHumanReadableCopyright</key>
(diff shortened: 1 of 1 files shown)


mrjbq7:
What benefit does this provide?

Is this for the "Open With..." right click menu on factor files?

tgunr:
Factor should claim .factor files the same as other applications on the Mac. By declaring .factor files conform to source files, plain-text files, and text files allows other applications to open and view content. By declaring .factor files as public.source-factor will permit Factor to use editors which conform to source-factor type files. 

Also will permit using mdimporter to search for Factor specific data.

Note: This only affects source files on the Mac as far as I know, *inx and Windows do not have similar constructs.
—
Dave Carlton
davec@polymicro.net

mrjbq7:
Thanks!

PR #2421, opened 2021-01-14
graphs: add closure-as, base closure on that

timor:
Motivation: `V{ } closure-as` returns elements in reverse post-order.


erg:
Merged, thanks!

PR #2426, opened 2021-01-30
xkcd: fix urls

alesguzik:
`random-xkcd.` was broken because the url has changed. I also took a liberty to update urls to use https, as that's what xkcd.com uses by default.

P.S. I haven't looked at factor for a couple of years. I see that it still has regular activity, but the last release was 2.5 years ago. Are there any plans to cut a new release? Any plans for 1.0?

### DIFF
--- a/extra/xkcd/xkcd.factor
+++ b/extra/xkcd/xkcd.factor
@@ -14,7 +14,7 @@ IN: xkcd
 : comic-image ( url -- image )
     http-get nip
     R/ \/\/imgs\.xkcd\.com\/comics\/[^\.]+\.(png|jpg)/
-    first-match >string "http:" prepend load-http-image ;
+    first-match >string "https:" prepend load-http-image ;
 
 : comic-image. ( url -- )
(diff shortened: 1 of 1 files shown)


mrjbq7:
Lots of plans for new features. I’m hoping for 0.99 near mid-2021. 

I kind of want to see 0.999, 0.9999, 0.9999, maybe we never reach 1.0!

(More likely we will switch to calendar releases like other projects, but it would be funny to make 0.100 the next release after 0.99)

mrjbq7:
Thanks!

alesguzik:
Thank you!

odiferousmint:
> I see that it still has regular activity

Yup, Factor is under active development by these awesome guys :). I saw that there is a Discord server, too. While I do not like Discord as I do not like GitHub, it is still one of the best ways to reach an audience. That, and the IRC channel, of course.

Discord invite link: https://discord.gg/BcKCjqPr7e
IRC channel on irc.freenode.net: #concatenative

PR #2429, opened 2021-02-09
VM: fix comments for inline_cache_jit

timor:
I think the two comments need to be swapped to describe the correct thing happening in the code?


mrjbq7:
Thanks!

PR #2430, opened 2021-02-14
Extend inline cache mechanism to support multiple dispatch

timor:
# Motivation
This is intended to provide a backend implementation for #2364 which re-uses the
existing inline caching and generics infrastructure as much as possible.

# Modifications
## Covariant Tuples
This is a new classoid which captures the notion of covariant subtyping
(inspired by Julia) of call-site parameter tuples.  This is also allows re-using
the existing method definition infrastructure and class-algebra functionality,
because internally, multi-methods are specialized on an instance of this.

This has nothing to do with Factor's tuple classes, and can be thought of an
instance of a `dispatch-specifier`-kind-of-like type.

## `generic.multi`
This vocab implements a new method combination `multi-combination`, and some
very basic Syntax support, but mostly the dispatch code generation logic.

This is inspired by amongst others, by [this
paper](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.49.5769&rep=rep1&type=pdf),
but for the generation of the multiple dispatch decision tree.  However, there
is zero optimization implemented yet.

The basic idea is to insert a step which converts the "methods" word property,
which contains specializers of type `covariant-tuple` into nested regular
dispatch decision trees, inserting indirections (similar to predicate engine
words), which have their own inline cache though.  For this purpose, a new
dispatch engine is provided which is basically the same as the
`tag-dispatch-engine**, but compiles intermediate tag dispatchers instead of only
top-level ones.

## Optimizing Compiler
Based on the code for inlining `single-generics`, the same thing is done for
multi-method call sites, so that the most specific method is also selected if
possible.  This is probably not completely correct yet.

# What works

## Defining Multimethods
`single-generic` generic words are turned into `multi-generic` words by defining
multi-methods on them, so there is no separate `MGENERIC:` right now.  They can
be mixed with regular methods.  The classes are specified in the effect of the
method definition like the `TYPED:` syntax.

e.g.
```factor
GENERIC: foo ( x x -- x )
MM: foo ( x: fixnum y: tuple -- x ) ... ;
MM: foo ( x: object y: object -- x ) ... ;
! Is the same as
M: object foo ... ;
```

## Generation of Dispatch logic
Naive successive lookup on the arguments using the megamorphic inline
cache and PIC mechanism.  Note that the nested dispatchers by default do not
generate a PIC stub, only a `mega-cache-lookup`.  This can be enabled, I have
only done 2 tests on this, and for one of them the additional indirection
actually seemed to have a negative impact on performance.

Note that most of the benefits only apply to tuple and built-in class dispatch,
and multiple dispatch on predicate classes (including singletons) will generate nested 
predicate engine words like before.

# What does not work
- This branch does not contain anything in the direction of partial inlining.
- There is no attempt to simplify the decision tree.
- There is no support for hook combinations
- Bootstrapping, probably due to circular dependencies
- probably a lot more... this needs to be tested more thouroughly

# Some more ideas
## More explicit dispatch specification
I think one of the things that is actually even more suitable for reducing
redundancy is Julia's notion of parametric dispatch. [This
paper](https://dl.acm.org/doi/pdf/10.1145/320384.320407), although quite old,
also describes an interesting mechanism of specifiying dispatch based on
predicates.

If, for example, singletons and mixins were based on tuples, they could take advantage of
the existing tuple dispatch engines due to the inheritance mechanism.  Also, it would be interesting to either
infer or provide special types of predicate classes (on tuples) which can be
proven to be subtypes. A predicate class which tests a certain tuple slot, for
example, will always be a subtype of the corresponding tuple class.

This kind of thinking was also one motivation for adding the `covariant-tuple`
classoid, as a device to separate dispatch specification from parameter data
types a bit.
## Get rid of nested dispatch tests for tuple-only participants
This is related to the previous point.  If all participating classes can be
ordered at compile time, it should be possible to generate the fast-hash
structure over the combination of all the objects involved at a call site
instead of successive tests.  That way, there would be no overhead at all when
dispatching only on a limited number of combinations of tuple classes.  This
would probably involve some modifications to the lookup-method code inside the VM.


timor:
Edited description to reflect changes in detection of ambiguous methods.

timor:
I am confused about bootstrapping:
What does work is:
1. Check out master, build master image
2. Switch to this branch
3. `refresh-all`, without dependency problems
4.  execute `make-my-image`
5. `build.sh bootstrap` with that image

What does not work is:
1. Check out this branch
2. `build.sh net-bootstrap`

The second variant dies with
```
*** Stage 2 early init... done
Loading resource:basis/bootstrap/stage2.factor
Loading resource:basis/command-line/command-line.factor
Loading resource:basis/debugger/debugger.factor
Loading resource:basis/compiler/errors/errors.factor
Loading resource:basis/grouping/grouping.factor
Loading resource:basis/io/styles/styles.factor
Loading resource:basis/colors/colors.factor
Loading resource:basis/delegate/delegate.factor
Loading resource:basis/delegate/protocols/protocols.factor
Loading resource:basis/deques/deques.factor
Loading resource:basis/io/streams/string/string.factor
Loading resource:basis/present/present.factor
Loading resource:basis/alien/c-types/c-types.factor
Loading resource:basis/cpu/architecture/architecture.factor
Loading resource:basis/strings/tables/tables.factor
Loading resource:basis/libc/libc.factor
Loading resource:basis/alien/destructors/destructors.factor
Loading resource:basis/functors/functors.factor
Loading resource:basis/functors/backend/backend.factor
Loading resource:basis/interpolate/interpolate.factor
Loading resource:basis/multiline/multiline.factor
Loading resource:basis/alien/syntax/syntax.factor
Loading resource:basis/alien/enums/enums.factor
Loading resource:basis/alien/libraries/libraries.factor
Loading resource:basis/alien/libraries/unix/unix.factor
Loading resource:basis/alien/parser/parser.factor
Loading resource:basis/alien/arrays/arrays.factor
Loading resource:basis/libc/linux/linux.factor
Loading resource:basis/prettyprint/prettyprint.factor
Loading resource:basis/colors/constants/constants.factor
Loading resource:basis/ascii/ascii.factor
Loading resource:basis/hints/hints.factor
Loading resource:core/generic/multi/multi.factor
Loading vocab:bootstrap/bootstrap-error.factor
Loading resource:basis/debugger/debugger.factor
Loading resource:basis/prettyprint/prettyprint.factor
Loading resource:basis/colors/constants/constants.factor
Loading resource:basis/ascii/ascii.factor
Loading resource:basis/hints/hints.factor
Loading resource:core/generic/multi/multi.factor
You have triggered a bug in Factor. Please report.
critical_error: The die word was called by the library.: 0
Starting low level debugger...
Basic commands:
  q ^D             -- quit Factor
  c                -- continue executing Factor - NOT SAFE
  t                -- throw exception in Factor - NOT SAFE
  .s .r .c         -- print data, retain, call stacks
  help             -- full help, including advanced commands

>

```

I noticed this behavior whenever I had a vocab dependency problem, but that was usually accompanied by `refresh-all` failing with the fresh image in the first variant above.  Is there some way to debug vocab dependencies from the low-level debugger?

mrjbq7:
If the first works then your process is pretty clean. That implies you added code that requires new boot images. And so a self-bootstrap works but a net-bootstrap doesn’t.

timor:
> If the first works then your process is pretty clean. That implies you added code that requires new boot images. And so a self-bootstrap works but a net-bootstrap doesn’t.

Thanks.  Are there certain criteria as to what kind of changes require new boot images?

mrjbq7:
Typically, for anything that changes in core/.

It is uncommon for basis/ changes to require new boot images. Basically boot image is a pre-compilation step that packages up core/.

timor:
I factored out the whole covariant tuple stuff into its own logic with an abstract `dispatch-type` type.

This PR basically now contains two separate concepts:

1. Generating nested inline caches from the class tuples of multi-methods for dispatch
2. Proof of Concept for a modular approach of parametrizing dispatch.

timor:
I added `classes.dispatch.class` which illustrates how this can be used to implement class methods.  Updated the description to incorporate an example using some quickly invented `CM:` word for simple constructor inheritance.

timor:
@kusumotonorio I just pushed a fix in 1890f5a8b, which you would need to apply to the corresponding code in your branch, if you want to stay synchronized.

kusumotonorio:
@tomor Roger that. I'm going to differentiate between those that actively follow your changes and those that don't by creating a new branch.

timor:
@kusumotonorio I really should split this PR into the dispatch-type abstraction stuff and the inline-cache implementation stuff, just haven't had time yet.

PR #2431, opened 2021-02-17
multi-generic

kusumotonorio:
This is one of the implementations that adds multi-methods to Factor.
This is an extension of the already existing `multi-methods`, which includes the `generic` functionality.
It declares methods with a unified syntax for generic words that dispatch with a single stack parameter, dispatch with a single hook variable, mathematical double-dispatch, and dispatch with a combination of stack parameters and hook variables.

Other features include
- Dispatching with a single stack parameter, dispatching with a single hook variable, and mathematical dispatching are handled with the same logic as `generic`, so their execution speed is expected to be the same.
- Even mathematical generic words can be dispatched by non-numeric objects.
- The `call-next-multi-method` can be used to allow differential programming through class inheritance.
- If the output of a method stack-effects declaration has class-specific information, the `typed` feature guarantees the class of the output and adds optimization information.
- If the class of some parameters used for dispatching can be estimated, it has a partial inline feature to generate compile-time dispatched code for the known parameters if possible.


kusumotonorio:
The commit I just made made it impossible to boot again. I was able to get through it the last time, but this time it seems to be difficult to solve...

timor:
(ref https://github.com/factor/factor/issues/2364#issuecomment-783372006)

@kusumotonorio
I pushed a [branch](https://github.com/timor/factor/commits/multi-generic) which contains two fixes.  The faulty code resulted in effectively generating only single-dispatch code.
This is how the benchmarks look on my machine now: 

```
100,000 repetitions of the showdown of the all combinations of No.001 to No.005 (singleton version)
(All methods/words are wrapped in words.)
single-dispatch:            0.010678 seconds (reference)
multi-dispatch:             0.070656 seconds (6.62 times slower)
cached-multi-dispatch:      0.050337 seconds (4.71 times slower)
single spec multi-dispatch: 0.010992 seconds (1.03 times slower)

100,000 repetitions of the showdown of the all combinations of No.001 to No.005 (tuple version)
(All methods/words are wrapped in words.)
single-dispatch:            0.009080 seconds (0.85 times slower)
multi-dispatch:             0.063536 seconds (5.95 times slower)
cached-multi-dispatch:      0.012376 seconds (1.16 times slower)
single spec multi-dispatch: 0.008915 seconds (0.83 times slower)
```

Makes a lot more sense now to me.

kusumotonorio:
I also imported the changes from the @timor's branch into my code and got the following result.

```
IN: scratchpad bm

100,000 repetitions of all combinations of rock-paper-scissors (singleton version)
no-dispatch:                0.003467 seconds (reference)
single-dispatch:            0.002193 seconds (0.63 times slower)
multi-dispatch:             0.002011 seconds (0.58 times slower)
cached multi-dispatch:      0.001851 seconds (0.53 times slower)
multi-dispatch (typed):     0.001905 seconds (0.55 times slower)
single spec multi-dispatch: 0.002154 seconds (0.62 times slower)
single-hook-dispatch:       0.054641 seconds (15.76 times slower)
multi-hook-dispatch:        0.039409 seconds (11.37 times slower)
single-hook-multi-dispatch: 0.055286 seconds (15.95 times slower)
wrapped sd:                 0.004904 seconds (1.41 times slower)
wrapped md:                 0.007514 seconds (2.17 times slower)
wrapped cached md:          0.007742 seconds (2.23 times slower)  <----
wrapped smd:                0.004811 seconds (1.39 times slower)

100,000 repetitions of all combinations of rock-paper-scissors (tuple version)
no-dispatch:                0.005495 seconds (1.59 times slower)
single-dispatch:            0.002925 seconds (0.84 times slower)
multi-dispatch:             0.001844 seconds (0.53 times slower)
cached multi-dispatch:      0.002134 seconds (0.62 times slower)
multi-dispatch (typed):     0.001849 seconds (0.53 times slower)
single spec multi-dispatch: 0.002985 seconds (0.86 times slower)
single-hook-dispatch:       0.055749 seconds (16.08 times slower)
multi-hook-dispatch:        0.043225 seconds (12.47 times slower)
single-hook-multi-dispatch: 0.058808 seconds (16.96 times slower)
wrapped sd:                 0.004979 seconds (1.44 times slower)
wrapped md:                 0.008840 seconds (2.55 times slower)  <----
wrapped cached md:          0.004584 seconds (1.32 times slower)
wrapped smd:                0.004645 seconds (1.34 times slower)
IN: scratchpad bm2

100,000 repetitions of the showdown of the all combinations of No.001 to No.005 (singleton version)
(All methods/words are wrapped in words.)
single-dispatch:            0.012252 seconds (reference)
multi-dispatch:             0.082302 seconds (6.72 times slower)
cached-multi-dispatch:      0.058202 seconds (4.75 times slower)  <----
single spec multi-dispatch: 0.011945 seconds (0.97 times slower)

100,000 repetitions of the showdown of the all combinations of No.001 to No.005 (tuple version)
(All methods/words are wrapped in words.)
single-dispatch:            0.009626 seconds (0.79 times slower)
multi-dispatch:             0.062025 seconds (5.06 times slower)
cached-multi-dispatch:      0.015438 seconds (1.26 times slower)  <----
single spec multi-dispatch: 0.009700 seconds (0.79 times slower)
```

kusumotonorio:
It would be great to have caching work in multi-method dispatch, and I would love to have such a mechanism.
The code I ported from @timor's PR seems to work well according to the benchmarks above. However, I would have to think harder to use that code in combination with multi-generic for purposes other than experimentation.

In `multi-generic`, which is derived from `multi-methods`, when generating code to determine methods, it performs a topological sort, sorting the methods in order of their specialization in the inheritance relation. By examining the results in order, the appropriate method is selected. There should be no difference between caching dispatch mechanisms and non-caching ones. I looked at the code I had ported and couldn't be sure that it would be sorted in an order that would not cause an inversion in the inheritance relationship. Of course, that may just be because I don't really understand how Factor orders its classes.

Perhaps giving Factor the newly introduced `covariant-tuple` algebra is necessary to make the cache mechanism work. So, it is also necessary to understand the relationship between the algebra for `covariant-tuple` and the comparison criteria for topological sorting.

timor:
> In `multi-generic`, which is derived from `multi-methods`, when generating code to determine methods, it performs a topological sort, sorting the methods in order of their specialization in the inheritance relation. By examining the results in order, the appropriate method is selected. There should be no difference between caching dispatch mechanisms and non-caching ones. I looked at the code I had ported and couldn't be sure that it would be sorted in an order that would not cause an inversion in the inheritance relationship.

Do you have an example in which the dispatch order is different for the cached and the non-cached variant?

> Perhaps giving Factor the newly introduced `covariant-tuple` algebra is necessary to make the cache mechanism work.

It should not be "necessary" per se, but it allows to re-use the single-dispatch and class algebra code already present.

kusumotonorio:
> Do you have an example in which the dispatch order is different for the cached and the non-cached variant?

I've never found it. However, I have a question as to why the author of `multi-methods` defined `classes<` instead of simply using `<=>`.

timor:
> I've never found it. However, I have a question as to why the author of `multi-methods` defined `classes<` instead of simply using `<=>`.

I have thought about the same thing.  I came to the conclusion that is because classes form a partially ordered set.  The only relation that you can query is `class<=`.  The `<=>` operator, as I understand it, presupposes a total order on the set elements, and relies on the fact that all pairs of a set are either `=`, or if not equal then `<` or `>`.  It does not contain the notion of `+incomparable+`.
If `a <= b` is `false`, that implies that `b <= a`  must be `true` in a totally ordered set.
However, for `a class<= b` resulting in `false`, concluding that `b class<= a` is `true` is wrong if `a` and `b` are incomparable.  Thus, you cannot use the same approach to sorting.

Note that there is an occurrence of `<=>` when sorting classes for method applicability to handle the case of incomparable classes.  So the dispatch is ordered by something which I think of as `method<=` or `dispatch<=`, which actually is guaranteed to be defined for each pair of the set of classes to dispatch on.

This is why I put the comment [here](https://github.com/factor/factor/pull/2431/commits/8984e64c562b3b7e3d89d10b4a6efb2654f43615#diff-04af266dcb753a0e8fb1ce56419fa19d2a1fc35217c7bf0723f4e9e8a1abe09eR193-R197), because technically the `<=>` is not defined for classes without tie breakers.

kusumotonorio:
I decided to look into method sorting. It turns out that there is a difference between sorting without cache derived from `multi-methods` and sorting by co-variant tuple algebra, but rather, I became suspicious of the sorting result without cache. I'll explain using `multi-methods`.

I have prepared a group of methods using several tuples with inheritance relations as follows:
``` factor
IN: scratchpad USE: multi-methods
Loading resource:extra/multi-methods/multi-methods.factor
IN: scratchpad 
TUPLE: t1 ;
TUPLE: t2 ;
TUPLE: t3 ;

TUPLE: t4 < t1 ;
TUPLE: t5 < t2 ;
TUPLE: t6 < t3 ;

TUPLE: t7 < t4 ;
TUPLE: t8 < t5 ;

IN: scratchpad 
GENERIC: mm ( o1 o2 o3 -- no )

METHOD: mm { t1 object object } 3drop 1 ;
METHOD: mm { t7 t5 object } 3drop 2 ;
METHOD: mm { t4 object object } 3drop 3 ;
METHOD: mm { object t8 t6 } 3drop 4 ;
METHOD: mm { object object object } 3drop 5 ;
METHOD: mm { object t2 object } 3drop 6 ;
METHOD: mm { object object t3 } 3drop 7 ;
METHOD: mm { t4 t5 t6 } 3drop 8 ;
METHOD: mm { object t5 object } 3drop 9 ;

IN: scratchpad 
```
I have sorted those methods using a word for topological sorting.
``` factor
IN: scratchpad \ mm methods sort-methods [ second ] map .
{
    mm-{ t7 t5 object }
    mm-{ t4 t5 t6 }
    mm-{ t4 object object }
    mm-{ t1 object object }
    mm-{ object t8 t6 }
    mm-{ object t5 object }
    mm-{ object t2 object }
    mm-{ object object t3 }
    mm-{ object object object }
}
```
The last of the parameters should be the top of the stack. The deepest ones in the stack with the highest specificity are at the top, and the last in the line are the methods with the lowest specificity.
If you look at the actual `def` of the generic word `mm`, it looks like this:
``` factor
IN: scratchpad \ mm def>> ...
[
    over t5? [ pick t7? ] [ f ] if
    [ mm-{ t7 t5 object } ] [
        dup t6? [ over t5? [ pick t4? ] [ f ] if ] [ f ] if
        [ mm-{ t4 t5 t6 } ] [
            pick t4?
            [ mm-{ t4 object object } ] [
                pick t1?
                [ mm-{ t1 object object } ] [
                    dup t6? [ over t8? ] [ f ] if
                    [ mm-{ object t8 t6 } ] [
                        over t5?
                        [ mm-{ object t5 object } ] [
                            over t2?
                            [ mm-{ object t2 object } ] [
                                dup t3?
                                [ mm-{ object object t3 } ] [
                                    t [
                                        mm-{ object object object }
                                    ] [
                                        3
                                        \ mm
                                        [ narray ]
                                        dip
                                        no-method
                                    ] if
                                ] if
                            ] if
                        ] if
                    ] if
                ] if
            ] if
        ] if
    ] if
]
```
It seems to be inspecting the result of the sort as I tried earlier.

I am uncomfortable with this result. Since the stack parameters are being examined in order from the top, shouldn't the methods be sorted in order of specificity from the top side?
I have never done anything more advanced than rock-paper-scissors with `multi-methods`. I would like to know if there is any operational experience of it for anyone.

kusumotonorio:
So does this mean that my understanding that `multi-methods` is identified from the top side of the stack is wrong?

timor:
> So does this mean that my understanding that `multi-methods` is identified from the top side of the stack is wrong?

It it is intended to be ordered right-to-left (top-to-bottom considering the stack), then I think it is wrong, yes.  Is the intended dispatch precedence order documented somewhere?

kusumotonorio:
>  Is the intended dispatch precedence order documented somewhere?

I've never seen it. Perhaps there is no such thing.
I saw multi-methods examine the stack from the top, and misunderstood it to be a sort that gives priority to the top side.

`multi-generic` needs to be fixed, but what needs to be decided is what policy to use to do the sorting. I can make it conform to the current sort, but to make it easier to trigger partial inline, it is better to decide the top of the stack first. I'm going to try to implement hook variables first, then stack variables near the top.

kusumotonorio:
I changed the sort order of methods in `multi-generic` to be different from `multi-methods`.

``` factor
IN: scratchpad USE: multi-generic
IN: scratchpad 
TUPLE: t1 ;
TUPLE: t2 ;
TUPLE: t3 ;

TUPLE: t4 < t1 ;
TUPLE: t5 < t2 ;
TUPLE: t6 < t3 ;

TUPLE: t7 < t4 ;
TUPLE: t8 < t5 ;

MGENERIC: m-nc ( o1 o2 o3 -- no )
MM:: m-nc ( :t1 :object :object -- No.1 ) 1 ;
MM:: m-nc ( :t7 :t5 :object -- No.2 ) 2 ;
MM:: m-nc ( :t4 :object object -- No.3 ) 3 ;
MM:: m-nc ( :object :t8 :t6 -- No.4 ) 4 ;
MM:: m-nc ( :object :object :object -- No.5 ) 5 ;
MM:: m-nc ( :object :t2 :object -- No.6 ) 6 ;
MM:: m-nc ( :object :object :t3 -- No.7 ) 7 ;
MM:: m-nc ( :t4 :t5 :t6 -- No.8 ) 8 ;
MM:: m-nc ( :object :t5 :object -- No.9 ) 9 ;
```
``` factor
IN: scratchpad \ m-nc methods sort-methods [ second ] map .
{
    MM\ m-nc ( :object :t8 :t6 -- No.4 )
    MM\ m-nc ( :t4 :t5 :t6 -- No.8 )
    MM\ m-nc ( :object :object :t3 -- No.7 )
    MM\ m-nc ( :t7 :t5 :object -- No.2 )
    MM\ m-nc ( :object :t5 :object -- No.9 )
    MM\ m-nc ( :object :t2 :object -- No.6 )
    MM\ m-nc ( :t4 :object object -- No.3 )
    MM\ m-nc ( :t1 :object :object -- No.1 )
    MM\ m-nc ( :object :object :object -- No.5 )
}
```
Sorting with `sort-single-methods` (equivalent to `sort-methods` in `generic`) whitch uses `<=>` in covariant-tuple dispatch looks like this:
``` factor
MGENERIC: m-c ( o1 o2 o3 -- no ) cached-multi
MM:: m-c ( :t1 :object :object -- NO.1 ) 1 ;
MM:: m-c ( :t7 :t5 :object -- No.2 ) 2 ;
MM:: m-c ( :t4 :object object -- No.3 ) 3 ;
MM:: m-c ( :object :t8 :t6 -- No.4 ) 4 ;
MM:: m-c ( :object :object :object -- No.5 ) 5 ;
MM:: m-c ( :object :t2 :object -- No.6 ) 6 ;
MM:: m-c ( :object :object :t3 -- No.7 ) 7 ;
MM:: m-c ( :t4 :t5 :t6 -- No.8 ) 8 ;
MM:: m-c ( :object :t5 :object -- No.9 ) 9 ;
```
``` factor
IN: scratchpad \ m-c "dispatch-type" word-prop methods>> sort-single-methods [ second ] map .
{
    MM\ m-c ( :object :object :object -- No.5 )
    MM\ m-c ( :t1 :object :object -- NO.1 )
    MM\ m-c ( :t4 :object object -- No.3 )
    MM\ m-c ( :object :t2 :object -- No.6 )
    MM\ m-c ( :object :t5 :object -- No.9 )
    MM\ m-c ( :t7 :t5 :object -- No.2 )
    MM\ m-c ( :object :object :t3 -- No.7 )
    MM\ m-c ( :t4 :t5 :t6 -- No.8 )
    MM\ m-c ( :object :t8 :t6 -- No.4 )
}
```
They are the same in that they are focused on the specificity of the top of the stack, but the order is reversed.

mrjbq7:
Is this changing the order of dispatch? Or is this an implementation detail of what order it checks the stack arguments?

kusumotonorio:
Since the method is checked for conformity in the sorted order, how it is sorted may be important for execution.

For example, let's define a set of generics and methods like the above in `multi-generic` and `multi-methods`, and put `t1`, `t5`, and `t6` tuples on the stack to call those generics.
``` factor
IN: scratchpad t1 new t5 new t6 new m-nc .
7
```
In `multi-generic`, the method that returns 7, `MM\ m-c ( :object :object :t3 -- No.7 )`, was chosen. It is the third from the top in the sort order. Since `t6` is a subclass of `t3` and is at the top of the stack, it was given more weight in the choice of methods.

Next, we will see if this is the case with `multi-methods`.
``` factor
 t1 new t5 new t6 new mm .
1
```
It chose the method `mm-{ t1 object object }`, which returns 1.
It is the fourth method in the `multi-methods`. The emphasis here is on `t1`, which is the deepest in the stack.

mrjbq7:
Is that eventually resolved via call-next-multi-method?

Meaning is it a question of ordering, or is one preventing the other from being called?

kusumotonorio:
> Meaning is it a question of ordering, or is one preventing the other from being called?

What exactly do you mean by this?

I believe that the choice of methods from either side of the stack is a matter of "specification promise", not "semantic certainty". Once that rule is set, it shouldn't matter which way we go, but we need to be consistent.

kusumotonorio:
I noticed that the function I implemented in `multi-generic` as `call-next-multi-method` sometimes does not match its name.
The definition of "*next* multi-method" in languages with multi-methods is "the method that would have been called if a method did not exist". My implementation of it is "the method that would have been called if the generic word had been called in the method's signature when the method did not exist", and it is possible that they would not be the same when called with the actual parameters. So if we were to give it a more appropriate name, it would be something like `call-super-muti-method` (if there is such a term).

I made it to mimic the `call-next-method` in `generic`. The method it replaces is statically determined at compile time of the method that calls it, but I think to achieve a true `call-next-multi-method`, the dispatching code would have to be embedded inline and dispatched at runtime.

When I imagine the situation where `call-multi-method` will actually be used, I feel that the current implementation is fine, but it is also undesirable to be different from other languages. I am not sure how it is handled in CLOS, but I found an explanation in R.

http://web.mit.edu/r/current/lib/R/library/methods/html/NextMethod.html

> The ‘next’ method (i.e., the first inherited method) is defined to be that method which would have been called if the current method did not exist. This is more-or-less literally what happens: The current method (to be precise, the method with signature given by the defined slot of the method from which callNextMethod is called) is deleted from a copy of the methods for the current generic, and selectMethod is called to find the next method (the result is cached in the method object where the call occurred, so the search typically happens only once per session per combination of argument classes).
> 
> The next method is defined from the signature of the current method, not from the actual classes of the arguments. In particular, modifying any of the arguments has no effect on the selection. As a result, the selected next method can be called with invalid arguments if the calling function assigns objects of a different class before the callNextMethod() call. Be careful of any assignments to such arguments.

It seems that it has a similar character to my implementation in R.

I'm trying to figure out what I should do. Here's what I came up with.

1. use the current implementation with the name `call-next-multi-method`. 
2. Rename the current implementation to `call-super-multi-method`. 
3. Stop the current implementation and create a true `call-next-multi-method` that runs slower.
4. Rename the current implementation to `call-super-multi-method` and create a new true `call-next-multi-method` so that users can use both.

Which one do you think is better?

kusumotonorio:
@timor I noticed that the cache dispatch of covariant-tuple does not make the correct method selection in `multi-generic` at the moment. I had checked this before and it felt normal then, so I reverted a change ( https://github.com/factor/factor/pull/2431/commits/2c37179926a9eb1c136bce807dad5c7b6720ae09 ) that mimicked your branch and the problem was fixed.

What was happening before this change was that the default method was chosen at any time. In the rock-paper-scissors test, the result was always a loss.
Does your branch dispatch correctly? If so, then my changes are not appropriate.

timor:
> What was happening before this change was that the default method was chosen at any time. In the rock-paper-scissors test, the result was always a loss.
Yes, I noticed that and that should be fixed in the current branch.

> Does your branch dispatch correctly? If so, then my changes are not appropriate.
Can you give an example which currently fails?

kusumotonorio:
I have previously made changes based on your changes https://github.com/timor/factor/commit/3c0c6ed8a055a1eec451e977f316dd5d2d84b80f and https://github.com/timor/factor/commit/cee950d9f6e2a282fb69be773c1c3657bdd1253a .

The one that disabled those changes passes the next test.

``` factor
MGENERIC: cached-md-beats? ( obj1 obj2 -- ? ) cached-multi

MM:: cached-md-beats? ( obj1: paper obj2: scissors -- ?: boolean ) obj1 obj2 2drop t ;
MM:: cached-md-beats? ( :scissors :rock -- ?: boolean ) t ;
MM:: cached-md-beats? ( :rock :paper -- ?: boolean ) t ;
MM:: cached-md-beats? ( :thing :thing -- ?: boolean ) f ;


{ f t f  f f t  t f f } [
    paper paper       cached-md-beats?
    paper scissors    cached-md-beats?
    paper rock        cached-md-beats?

    scissors paper    cached-md-beats?
    scissors scissors cached-md-beats?
    scissors rock     cached-md-beats?

    rock paper        cached-md-beats?
    rock scissors     cached-md-beats?
    rock rock         cached-md-beats?
] unit-test
```

Enabling these changes does not give the expected result.
``` factor
resource:extra/multi-generic/multi-generic-tests.factor: 233

Unit Test: {
    { f t f f f t t f f }
    [
        paper paper cached-md-beats?
        paper scissors cached-md-beats?
        paper rock cached-md-beats?
        scissors paper cached-md-beats?
        scissors scissors cached-md-beats?
        scissors rock cached-md-beats?
        rock paper cached-md-beats?
        rock scissors cached-md-beats?
        rock rock cached-md-beats?
    ]
}

=== Expected:
f
t
f
f
f
t
t
f
f
=== Got:
f
f
f
f
f
f
f
f
f
```
I left `inline` in the definition of the generic for that test, so I was late to notice this problem.

kusumotonorio:
@timor 
Hmmm, I downloaded the branch you made ( https://github.com/timor/factor/tree/multi-generic ), recreated the image and tried it, but still `cached-md-beats?` returns `f` in any combination.
Now I'm trying to make my branch equivalent to yours.

timor:
> @timor
> Hmmm, I downloaded the branch you made ( https://github.com/timor/factor/tree/multi-generic ), recreated the image and tried it, but still `cached-md-beats?` returns `f` in any combination.
> Now I'm trying to make my branch equivalent to yours.

Oh sorry, that was not what I meant with "my latest code", I was referring to the other PR (#2430).  If you point me to a commit to check out, I am happy to take a look what could be the difference.

kusumotonorio:
I have always wanted to run your PR under myself, and have tried several times, but unfortunately have not yet succeeded in doing so.
Maybe I should make some changes to master and replicate that branch of yours.

kusumotonorio:
https://github.com/factor/factor/pull/2431#issuecomment-791846923
> I noticed that the function I implemented in multi-generic as call-next-multi-method sometimes does not match its name.

 Right now, the method selection priority of `multi-generic` is (probably temporarily) set to left-to-right (bottom-to-top), which makes it easier to compare with CLOS. So, I did an experiment to see what differences there are between the two, and recorded it here.
``` lisp
(defclass t0 () ())
(defclass t1 (t0) ())
(defclass t2 (t1) ())
(defclass t3 (t2) ())

(defmethod m-cl ((c1 t3) (c2 t0) (c3 t2)) (format t "No.1-") (call-next-method))
(defmethod m-cl ((c1 t2) (c2 t2) (c3 t2)) (format t "No.2-") (call-next-method))
(defmethod m-cl ((c1 t2) (c2 t0) (c3 t0)) (format t "No.3-") (call-next-method))
(defmethod m-cl ((c1 t0) (c2 t0) (c3 t0)) (format t "No.4~%"))

* (m-cl (make-instance 't3) (make-instance 't2) (make-instance 't0))
No.3-No.4
NIL
* (m-cl (make-instance 't2) (make-instance 't2) (make-instance 't2))
No.2-No.3-No.4
NIL
* (m-cl (make-instance 't3) (make-instance 't2) (make-instance 't2))
No.1-No.2-No.3-No.4
NIL
```
``` factor
USE: multi-generic

TUPLE: t0 ;
TUPLE: t1 < t0 ;
TUPLE: t2 < t1 ;
TUPLE: t3 < t2 ;

MGENERIC: mg ( o1 o2 o3 -- )
MM: mg ( :t3 :t0 :t2 -- ) "No.1-" write call-next-multi-method ;
MM: mg ( :t2 :t2 :t2 -- ) "No.2-" write call-next-multi-method ;
MM: mg ( :t2 :t0 :t0 -- ) "No.3-" write call-next-multi-method ;
MM: mg ( :t0 :t0 :t0 -- ) 3drop "No.4" write nl ;
IN: scratchpad t3 new t2 new t0 new mg
No.3-No.4
IN: scratchpad t2 new t2 new t2 new mg
No.2-No.3-No.4
IN: scratchpad t3 new t2 new t2 new mg
No.1-No.3-No.4
```
In the last example of `multi-generic`, the second method was not included in the `call-next-multi-method` chain because the second parameter of the second method is not a superclass of the corresponding parameter of the first method.

timor:
@kusumotonorio Wow, that's really interesting! I always thought that Factor and CLOS have the same semantics there.  The documentation of `call-next-method` states that it is an error to call a next method with something that would not have been able to dispatch on the current method, and inserts corresponding checks before calling the next method.  I wonder whether this kind of restriction is consistent in the multiple-dispatch case.

I think the difference is checking the objects at the statically determined next-method call vs. performing something like successive dispatch, deciding anew what the next applicable method is at the next-method call site, but kind of excluding dispatching to the present one again?  Would that mean that it was possible to create infinte loops in the CLOS case?

timor:
I just checked the behavior with my covariant tuple branch.  This is one of the cases where the dispatch is ambiguous.  A compiler "warning" error is generated after defining all methods: `T{ ambiguous-method-specializations
    { classes { D{ t3 t2 t2 } } }
}` which informs about the problematic case.  It still compiles successfully, and when run, exhibits the same behavior as `multi-generic`.

kusumotonorio:
> Factor does not do that, since it does not have the information of what the generic function was called with when dispatching again in a nested call-next-method.

I think you're right.
In a single parameter dispatch with `generic`, it would work just fine as a `call-next-method`, but in a multi-dispatch, there could be a case where this would make a difference. However, it is doubtful that we will ever see such a combination of methods in real-world programming as I have experimented with.

kusumotonorio:
In [Common Lisp Cookbook](https://lispcookbook.github.io/cl-cookbook/clos.html), `call-next-method` is described as follows:

> During the execution of a method, the remaining applicable methods are still accessible, via the local function call-next-method. This function has lexical scope within the body of a method but indefinite extent. It invokes the next most specific method, and returns whatever value that method returned.

If we take the phrase "the next most specific method" to mean "the method with the next most specific specializer for the current method's specializer", we can use `call-next-method` or `call-next-multi-metod` could be used.

kusumotonorio:
The current version of `multi-generic` does what I want it to do, but unfortunately it is not complete enough for practical use. What I consider to be the biggest problem at the moment is the redefinition of generic words and the behavior of `multi-generic` when reloading.

I need to understand what I have to do when I redefine a Factor word, but since `multi-generic` has two origins, `multi-method` and `generic`, it seems to me that they are treated differently in the vocabularies I'm not sure what to do.

timor:
> The current version of `multi-generic` does what I want it to do, but unfortunately it is not complete enough for practical use. What I consider to be the biggest problem at the moment is the redefinition of generic words and the behavior of `multi-generic` when reloading.

Can you give an example of what the problematic behavior is?

kusumotonorio:
@timor Thank you for your interest. I will answer with a better idea of what the problem is. Please give me some time.

core/classes/algebra/algebra.factor
```diff
@@ -195,6 +244,18 @@ M: anonymous-intersection (classes-intersect?)
 M: anonymous-complement (classes-intersect?)
     class>> class<= not ;
 
+: 2any? ( ... seq1 seq2 quot: ( ... elt1 elt2 -- ... ? ) -- ... ? )
```
mrjbq7, line 247: this is now in sequences vocab

core/classes/algebra/algebra.factor
```diff
@@ -195,6 +244,18 @@ M: anonymous-intersection (classes-intersect?)
...
+     [ not ] compose 2all? not ; inline
+
+ ! TODO: Check if semantics are correct here!
+ ! It is stated that the intersection is empty, if no object can be an instance
+ ! of both classes at the same time.  With dispatch in mind, I would say that if
+ ! there is any position that can not be an instance of both, two dispatches are
+ ! exclusive.
+ M: covariant-tuple (classes-intersect?)
+     covariant-classes
+     [ classes-intersect? ] 2all? ;
```
mrjbq7, line 257: is this supposed to be ``2any?``?

PR #2432, opened 2021-02-18
sequences: Add 2any? combinator

timor:
Missing member of the pairwise sequence combinator family.


timor:
CI fails on linting docs.  Is this because it is a change to a core vocab?

mrjbq7:
It appears that our Travis CI just builds from the existing boot image, so changes to core aren't incorporated until the boot image is updated.  That's less than helpful!

mrjbq7:
Btw, thanks @timor for the contribution!

We set Travis CI up as a quick check, but rely on our own build infrastructure to run all the tests (post-commit).  With Travis CI transitioning to a paid service or something soon, we might need to adjust our approach with pre-commit checks.  Github Actions seems pricey, but maybe we can use the "run on our own computers" version of that service.

timor:
Good to know, thanks.

PR #2434, opened 2021-03-04
Add base16 themes

inivekin:
Allows theming as in this repo, but fixes the theming setup process and adds help docs: https://github.com/inivekin/base16-factor
See https://github.com/chriskempson/base16 for more details


mrjbq7:
Awesome, thanks!!

PR #2435, opened 2021-03-12
Fix apropos bug in theme switching

inivekin:
The header on apropos output, restart-required-on-vocab-reload output and generally any output relying on title-style now adheres to the theme correctly.

Also changed the base16 theming to reduce highlight of borders around the next/previous buttons at the bottom of the help browser to improve aesthetics.


mrjbq7:
Awesome, thanks!  

It looks like you are trying to merge from ``master`` but also tracking ``master`` and our branches have diverged a little -- I recommend forcing your ``master`` to track ours and then make a branch for each pull request, so it's a little easier to merge automatically!

Committed as fe2e10bda51b4b80d215d3eb39f176476a48e9c8.

PR #2436, opened 2021-03-12
http: Make parse-set-cookie matching case insensitive

### LINKED ISSUE #2176: Unexpected "Cookie value must not be f" error
I'm trying to make a connection to a server, using session cookies. I have a running example with working curl parameters. which I'm trying to translate to Factor. Upon my first GET http-request, I receive three cookies, and I want to set these for the subsequent POST http-request. However, just storing the cookies from the response object into the request object raises an unexpected "Cookie value must not be f" error, Upon inspection, I received this cookie (prettyprinted):

T{ cookie { name "Secure" } }

(raw)

T{ cookie f "Secure" f f f f f f f f f }

What's the problem here? How to solve?

gifti258:
Fixes #2176

The matching is case sensitive but per RFC 6265 clients should match case-insensitively. When an attribute has uppercase letters, it is interpreted as a cookie name instead. This causes the cookie list to be wrongly parsed, cookies can be torn into multiple ones. When a separated attribute without a value is fed back to a request, it causes the problem mentioned in the issue.

I'm a Factor newbie and I guess this could be written better, without clever deep stack shuffling and forbidden shuffle words.

### DIFF
--- a/basis/http/http.factor
+++ b/basis/http/http.factor
@@ -1,6 +1,6 @@
 ! Copyright (C) 2003, 2010 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors arrays assocs base64 calendar calendar.format
+USING: accessors arrays ascii assocs base64 calendar calendar.format
 calendar.parser combinators fry hashtables http.parsers io io.crlf
 io.encodings.iana io.encodings.utf8 kernel make math math.parser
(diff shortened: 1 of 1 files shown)


mrjbq7:
Do you have some test cases that you can contribute also?

mrjbq7:
I added a test case from the RFC in 7fe89f8a1e88ebb34dffafee900f61138f6ac7a5.

Thanks!

PR #2447, opened 2021-03-31
HPack for HTTP/2

DMind44:
Introducing a vocabulary for HPack header compression, including static huffman encoding. This is for providing HTTP/2 protocol support. Based on RFC 7541 [https://tools.ietf.org/html/rfc7541](https://tools.ietf.org/html/rfc7541). Provide unit tests and manually checked using round trip testing against python hyper library [https://github.com/python-hyper/hpack](https://github.com/python-hyper/hpack).


basis/http2/hpack/hpack-docs.factor
```diff
@@ -0,0 +1,36 @@
+! Copyright (C) 2021 Your name.
```
erg, line 1: Please add a name/organization here.

basis/http2/hpack/hpack-tests.factor
```diff
@@ -0,0 +1,304 @@
+! Copyright (C) 2021 Your name.
```
erg, line 1: Please add a name/organization here. (Same as above)

PR #2449, opened 2021-04-04
math.intervals: Add [0,b] and [0,b), use in a couple of places

timor:
This replicates the `[0,b]` and `[0,b)` words from `math.ranges`.


mrjbq7:
Thanks!

PR #2450, opened 2021-04-05
assocs.extras: Combinator for merging two assocs

timor:
Combine two assocs, specifying a quotation in case there is a value for a
certain key in both.

### DIFF
--- a/extra/assocs/extras/extras-tests.factor
+++ b/extra/assocs/extras/extras-tests.factor
@@ -1,4 +1,4 @@
-USING: assocs.extras kernel math sequences tools.test ;
+USING: arrays assocs.extras kernel math math.order sequences tools.test ;
 
 { f } [ f { } deep-at ] unit-test
 { f } [ f { "foo" } deep-at ] unit-test
@@ -12,7 +12,7 @@ USING: assocs.extras kernel math sequences tools.test ;
 { H{ { "a" V{ 2 5 } } { "b" V{ 3 } } { "c" V{ 10 } } } }
 [
     { H{ { "a" 2 } { "b" 3 } } H{ { "a" 5 } { "c" 10 } } }
-    [ ] [ assoc-merge ] map-reduce
+    [ ] [ assoc-collect ] map-reduce
 ] unit-test
 
 { H{ } } [ H{ { 1 2 } } 2 over delete-value-at ] unit-test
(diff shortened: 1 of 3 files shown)


mrjbq7:
We have several potential merge methods:

* replace (assocs.extras:assoc-union)
* push-at (assocs.extras:assoc-merge)
* at+

Is it possible to make some of these use your implementation?

timor:
> Is it possible to make some of these use your implementation?

Could you point me to the exact words you mean? (I can't find `assoc-infimum`)

One point to note is that this uses `assoc-map`, which duplicates the first assoc, while `assoc-union` e.g. pre-allocates an assoc which is the sum size of both assocs, if I understand the code correctly. So there may be a performance/memory impact depending on how big the overlap (number of "conflicting" entries) between the two assocs is.

mrjbq7:
assoc-infimum doesn’t exist, I was just thinking about your min and max example in the PR description.  

assoc-merge exists in assocs.extras, perhaps it needs a better name though for collecting values for a given key.

timor:
> assoc-merge exists in assocs.extras, perhaps it needs a better name though for collecting values for a given key.

`assoc-collect` ?

mrjbq7:
That might be better naming!

It would be cool if we had definitions something like:

```
: assoc-collect ( seq -- assoc ) [ ?push ] assoc-merge-with ;

: assoc-infimum ( seq -- assoc ) [ min ] assoc-merge-with ;

: assoc-supremum ( seq -- assoc ) [ max ] assoc-merge-with ;

: assoc-union-all ( seq -- assoc ) [ nip ] assoc-merge-with ;
```

timor:
`assoc-collect` has different semantics, I fear.
That would be equivalent to
```factor
[ [ 1vector ] assoc-map ] bi@ [ append ] assoc-merge-with
```
And the current `assoc-merge` looks way more efficient than that.

The others look sound at first glance.

timor:
Naming-wise `assoc-infimum` and `assoc-supremum` suggest a scalar result to me. I'd suggest `assoc-min` and `assoc-max`, corresponding to `vmin` and `vmax` semantics.

mrjbq7:
Sounds good

mrjbq7:
One caveat, we have a pattern of words that apply to two assocs and ones that apply to a sequence of assocs. Those might have different names (like assoc-union and assoc-union-all).

timor:
Then `assoc-union-all` in your example above should only be `assoc-union`, I think.  `assoc-merge-with` expects a binary operation between assoc elements.  You can of course then use `reduce` to apply that to a sequence of assocs.

timor:
Like this?
```factor
: assoc-collect! ( assoc1 assoc2 -- assoc1 )
    over [ push-at ] with-assoc assoc-each ;

: assoc-collect ( assoc1 assoc2 -- newassoc )
    [ [ [ assoc-size ] bi@ + ] [ drop ] 2bi new-assoc ] 2keep
    [ assoc-collect! ] bi@ ;

! iterate over assoc2, replace conflicting values
! Modifies assoc1
: assoc-merge-with! ( assoc1 assoc2 quot: ( value1 value2 -- new-value ) -- assoc1' )
    [| key2 val2 quot | val2 key2 pick
     at* [ swap quot call ] [ drop ] if
     key2 pick set-at ] curry assoc-each ; inline

: assoc-merge-with ( assoc1 assoc2 quot: ( value1 value2 -- new-value ) -- new-assoc )
    pick [ [ clone ] 2dip assoc-merge-with! ]
    [ drop nip ] if
    ; inline

: assoc-min ( assoc1 assoc2 -- assoc )
    [ min ] assoc-merge-with ;

: assoc-max ( assoc1 assoc2 -- assoc )
    [ max ] assoc-merge-with ;

: assoc-sum ( assoc1 assoc2 -- assoc )
    [ + ] assoc-merge-with ;

: new-assoc-union! ( assoc1 assoc2 -- assoc1 )
    [ nip ] assoc-merge-with! ;

: new-assoc-union ( assoc1 assoc2 -- union )
    [ nip ] assoc-merge-with ;

: new-assoc-union-as ( assoc1 assoc2 exemplar -- union )
    swap [ assoc-clone-like ] dip new-assoc-union ;

: reduce-assocs ( seq quot: ( assoc1 assoc2 -- new-assoc ) -- assoc )
    [ ] swap map-reduce ; inline

: assoc-collapse-with ( seq quot: ( value1 value2 -- new-value ) -- assoc )
    [ assoc-merge-with ] curry reduce-assocs ; inline

: assoc-union-all1 ( seq -- assoc )
    [ new-assoc-union ] reduce-assocs ;

: assoc-union-all2 ( seq -- assoc )
    [ nip ] assoc-collapse-with ;

: assoc-sum-all ( seq -- assoc )
    [ + ] assoc-collapse-with ;
```

mrjbq7:
Yeah, maybe!

With perhaps the sequence based ones having a capacity heuristic that creates a ``new-assoc`` of the appropriate type (type of first assoc in the sequence?)... maybe the sum of assoc-size of all assocs?  That would not be correct if it was a lot of overlapping keys though...

I can also see having a key-quot and a value-quot for merging to make it more general, but perhaps that's not worth it until we have use-cases that need it.

And for things like assoc-union-all, it would be better to use a destructive merge probably.

If you think this is all not worth it right now, we can focus on the PR itself.

timor:
Renamed `assoc-merge` → `assoc-collect`, and shortened `assoc-merge-with` to `assoc-merge`.  Reducing version is called `assoc-collapse`, and uses the destructive operation after cloning the first assoc.  Changed `sequences.abbrev` to use that instead of duplicate `assoc-merge` private implementation.

mrjbq7:
Pretty cool!

PR #2452, opened 2021-04-08
classes.predicate: Implement <CLASSIFY ... CLASSIFY>

timor:
This allows imposing a mutually-exclusive order on predicate classes, which helps with dispatch on ambiguous predicate classes.

Example:
```factor
USING: classes.predicate ;
IN: scratchpad
<CLASSIFY
PREDICATE: b-range < number 50 < ;
PREDICATE: a-range < number 100 < ;
CLASSIFY>
GENERIC: smallest-range ( number -- limit )
M: a-range smallest-range drop 100 ;
M: b-range smallest-range drop 50 ;
40 smallest-range .
50
```

This is directly inspired by

Ernst, Michael, Craig Kaplan, and Craig Chambers. "Predicate dispatching: A unified theory of dispatch." European Conference on Object-Oriented Programming. Springer, Berlin, Heidelberg, 1998.

(Section 2.5)


timor:
I fixed an issue with inheritance and also simplified the syntax.

PR #2466, opened 2021-05-16
Adds a keyboard-driven link click gesture

inivekin:
This is a more keyboard driven method to select links from the focused window - it displays a pop-up window with fuzzy selection of the labelled, clickable buttons of the current window.

I chose fuzzy selection over a method like enumerating the links to a number/letters (what tools like vimium does) to select because this way you know what to type before initiating the selection by using the button label. I have also mapped it to alt+f only because it is a similar mapping to vimium.

Let me know if I've not implemented the mixin and graft/ungraft stuff correctly - I'm not going to pretend to I understand factor's object-oriented side yet.


basis/ui/tools/button-list/button-list.factor
```diff
@@ -0,0 +1,56 @@
...
+    children>> [ label? ] find swap [ text>> ] [ drop f ] if ;
+
+: store-labelled-button ( button -- str/f )
+    dup label-from-button [ [ active-buttons get set-at ] keep ] [ drop f ] if* ;
+
+: remove-labelled-button ( button -- str/f )
+    label-from-button [ dup active-buttons get delete-at ] [ f ] if* ;
+
+M: clickable graft* [ store-labelled-button drop ] [ call-next-method ] bi ;
+M: button ungraft* [ remove-labelled-button drop ] [ call-next-method ] bi ;
```
mrjbq7, line 30: Why is graft defined on ``clickable`` but ingraft on ``button``?

basis/ui/tools/tools.factor
```diff
@@ -34,6 +33,28 @@ tool "common" f {
     { T{ key-down f f "F11" } toggle-fullscreen }
 } ? prepend define-command-map
 
+
+tool "tool-switching" f {
```
mrjbq7, line 37: Is this section duplicative of the above code?

basis/ui/tools/button-list/button-list.factor
inivekin: Is fixed on the next commit, just a massive blunder on my part - static typing systems have made me complacent about these things

basis/ui/tools/tools.factor
inivekin: Not even sure how I wrangled this - sorry for making your job difficult, will be more careful next time. Should be fixed on the next commit.

mrjbq7: I don't mind difficult!

It's nice to get contributions and new features!

mrjbq7:
Merged, thanks!

I can't use ``A+f``, so made it ``A+t`` temporarily.

PR #2470, opened 2021-05-23
Raylib fixes/improvements + mesh picking demo

timor:
I was playing around with raylib 3D functionality by trying to get the [models/mesh-picking example](https://www.raylib.com/examples.html) to work, so I thought I just add it to the demos.  I found a couple of bugs in the FFI along the way, and implemented some utility stuff:

- `Vector{2,3,4}` types implement the sequence protocol
- Defined an `ARRAY-SLOT:` helper to deal with the pattern in raylib structs to have length and pointer fields


mrjbq7:
Cool!

PR #2471, opened 2021-05-24
[RFC] Use glfw explicitly in raylib mesh-picking demo to allow context switching

timor:
This shows how to use `yield` inside raylib render loops using glfw API calls to allow running raylib applications in separate Factor threads cooperatively.

However, there are a couple of issues with this, so I don't know it it is the right way to solve it:
- raylib does not document any way of being incorporated to other multi-threading applications as far as I could find
- relies on implementation details, specifically on raylib using `glfw`
- requires `glfw` as another runtime library dependency
- While this works for this example, I have no idea whether it will work with the other parts of raylib, like audio

Also, I don't have the means to test this on Windows or Mac.

To test:
```factor
USING: raylib.demo.mesh-picking threads ;
[ main ] "raylib-demo" spawn
```


mrjbq7:
Should ``raylib.glfw`` just be ``glfw`` since it's not depending on any raylib stuff?

mrjbq7:
Added the ``glfw.ffi`` vocab, separate from this PR.

Perhaps we should make the rest of the changes, although there are conflicts to resolve now that I updated raylib to 4.2 and changed the mesh-picking demo slightly.

mrjbq7:
Added ``glfw-yield`` in ab69a4f171ca48d6d53d81cfd52a42108da1c08b

PR #2472, opened 2021-05-26
Add command to run unit tests after refreshing

timor:
This saves a bit of typing when one uses `F2` and `"long-vocab.name" test "other-vocab" test` a lot during test-driven development. 

Added a default key binding on `Shift-F2` for the UI Tools windows, and the corresponding `fuel` bindings as `C-c M-r` (open to suggestions on this one, not too happy...)

### DIFF
--- a/basis/tools/test/test-docs.factor
+++ b/basis/tools/test/test-docs.factor
@@ -1,4 +1,4 @@
-USING: help.markup help.syntax kernel quotations io ;
+USING: help.markup help.syntax kernel quotations io strings vocabs.refresh ;
 IN: tools.test
 
 ARTICLE: "tools.test" "Unit testing"
@@ -82,6 +82,16 @@ HELP: test
 
 HELP: test-all
 { $description "Runs unit tests for all loaded vocabularies." } ;
+
+HELP: refresh-and-test
+{ $values { "prefix" string } }
+{ $description "Like " { $link refresh } ", but runs unit tests for all reloaded vocabularies afterwards." } ;
+
+HELP: refresh-and-test-all
+{ $values { "prefix" string } }
+{ $description "Like " { $link refresh-all } ", but runs unit tests for all reloaded vocabularies afterwards." } ;
+
+{ refresh-and-test refresh-and-test-all } related-words
 
 HELP: :test-failures
 { $description "Prints all pending unit test failures." } ;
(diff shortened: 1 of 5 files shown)


basis/vocabs/refresh/refresh.factor
```diff
@@ -1,8 +1,8 @@
 ! Copyright (C) 2009 Slava Pestov.
 ! See http://factorcode.org/license.txt for BSD license.
-USING: accessors assocs checksums checksums.crc32
-combinators.short-circuit io.encodings.utf8 io.files kernel
-namespaces sequences sets source-files vocabs vocabs.loader ;
+USING: accessors assocs checksums checksums.crc32 io.encodings.utf8 io.files
+kernel namespaces sequences sets source-files tools.test.private vocabs
```
mrjbq7, line 4: I'm not sure vocabs.refresh can depend on tools.test.

It is likely that ui.tools can depend on tools.test if you want to move these words to tools.test.

timor: > I'm not sure vocabs.refresh can depend on tools.test.
> 
> It is likely that ui.tools can depend on tools.test

I tried boostrapping, and everything seemed to be fine.  Is there another reason why it should not depend on it?
I can move these words to `tools.test` nevertheless.

mrjbq7: Right now both tools.test and vocabs.refresh are part of "bootstrap.tools", so that might currently be the case, but I'm not sure but it seems tools.test is the more optional vocabulary.

timor: I moved the code over to `tools.test`.  Please have a look if it looks correct.

timor:
Actually there probably two cases which I frequently encounter:
1. Modify vocab until tests succeed/Implement new stuff and see whether it breaks: Here the actions are `refresh-all` and `"interesting-vocab" test`.
2. Modify unit tests until they succeed/Add more failing unit tests: Here the actions are repeated `"interesting-vocab" test` invocations, where I switch between the test vocabulary file and the "Errors" window.  Maybe a kind of `retest-failed` or `retest-last` command would be useful there?

mrjbq7:
If we want your second idea, we probably need to change how we define and run tests, @erg and I have discussed similar idea, basically moving from:

{ t } [ 1 1 = ] unit-test

to something like:

TEST: [ 1 1 = ] { t }

where the second creates a test case than can be independently run later instead of top-level code.  There is some issue with defining setup/teardown code for these test cases, which are often not (in practice) independent of the previous test cases defined in that file.

PR #2480, opened 2021-07-21
hashtables: Preserve access semantics for alist >hashtable conversion

timor:
Alists have a left-to-right first key occurrence wins access strategy.  This
changes the iteration order when converting an alist via `>hashtable`, so that
the first entry will dominate the resulting assoc.


mrjbq7:
Thanks!

PR #2482, opened 2021-07-22
assocs.extras: Add assoc-collapse! and assoc-collapse-as

timor:
`assoc-collapse!` is the destructive version of `assoc-collapse`, but takes the
target assoc as extra element.

For `assoc-collapse-as`, the size of resulting assoc is based on the first assoc
in the input sequence.


timor:
I personally find these useful, but maybe they don't justify having their own words.  Feel free to close the PR if you think they don't add enough value...

mrjbq7:
Thanks!

PR #2484, opened 2021-08-12
Create test-word.factor

Bruno-366:


Bruno-366:
was unsure as to where to put it, so I put it in `extra`. Also the word could be renamed to something better...

timor:
Just out of curiosity, what is your use case for this?  It looks similar to the interface for writing unit tests in a way.

Bruno-366:
Yeah, I agree it’s similar to a unit test but not quite. It could be used
as a building block for property testing if factor supported that.
Right now it’s just a useful tool that I often use when writing words.

Here’s another [example], I have 5 different implementations of `xor` and
want to make sure they all work as expected. So I generate all possible
inputs with `all-selections` and then use `test-word` on the different
implementations.

[example]:
https://forward-lang.github.io/forward/blog/X-VHDL-forth.html#_test_script

```
{ f t } 2 all-selections
[ xor1 ] test-word .
```

timor:
Maybe this should be added to the test vocab then?

Bruno-366:
Well... perhaps, if that’s ok with you :)

mrjbq7:
If it doesn’t test, we shouldn’t call it test-word, but I’m not opposed to
more words that are useful for people.

Bruno-366:
Yeah it doesn’t test in the conventional sense that it compares with an
expected output, but I couldn’t come up with a better name haha

timor:
Ah sorry. I was under the impression that you use it for unit testing in the end.

Bruno-366:
I'm going to add one more word that I wrote some time ago, its a bit similar.
Given an array of words (for example: produced by `\ palindrome? definition`)
it will execute those words and show any intermediary values, as a form of super simple printf debugging.

jonenst:
The help still has "before-word" but the actual word is "before-after".
What about "visualize-results" or "gather-results" instead of "test-word" ?
And "gather-intermediates" for "before-after" ?

To me, this is a good tool for visualizing results when you don't really know what you want and are exploring (otherwise, unit tests are superior because they are silent when they work and check results)

Bruno-366:
Yeah that actually sounds good.
I’m going to try to see if I can implement some kind of property testing
using the stack-checker
<https://docs.factorcode.org/content/vocab-stack-checker.html> and perhaps
math.combinatorics
<https://docs.factorcode.org/content/vocab-math.combinatorics.html>.
The idea is to generate a decent subset of the inputs a word could possibly
accept.
I think the usefulness of gather-results would increase by a lot then.

Bruno-366:
Hmm, the property-based testing seems to be harder than I expected.

```factor
[ xor ] infer in>> length
```
I can use the above to figure out if a word is unary, binary or ternary (ie how many inputs it takes).

```factor
{ 1 2 3 4 5 } 1 all-selections . ! for unary number words
{ 1 2 3 4 5 } 2 all-selections . ! for binary number words
```
There's three obstacles I don't know how to solve:
1) In the above example, for words with 2 inputs, we expect the word to have 2 inputs of the same class
   If we instead have for example `remove-nth` `(  n seq -- seq' )` which expects 2 inputs with different classes,
  then the following would not work, because sometimes we would get `n seq` and sometimes `seq n` (also `n n` and `seq seq`)

```factor
{ 0 "a"  } 2 all-selections
```
So the above problem was how to generate possible inputs, while taking into account situations where stack order matters
The second problem is:

2) I don't know if its possible to figure out from `infer`'s output the appropriate class for the inputs of a word.

```factor
[ xor ] infer in>> ! outputs an array containing the strings "x" "x"
```
"x" could be anything, not necessarily a number, and I think this has to do with the fact that factor is dynamically typed.

3) There's a random numbers generator, but in true property testing you would have a random array generator, random string generator, etc.

---

Summary: my guess is that from the stack effect of a word, there isn't an easy way to generate possible inputs for that word,
this is something that has to be done by hand, although I'll mention in the docs a few examples as how to do this :)

timor:
Could you replace "quote" with "quotation" in the documentation?

razetime:
these seem to be extension of `with-datastack`, so with that in mind i have these ideas:
- change vocab to `continuations.extras`
- rename `gather-results` to `with-datastacks`
- rename `gather-intermediates` to `datastack-states`, or something similar.

mrjbq7:
@razetime sounds good to me!

PR #2486, opened 2021-09-07
Dispatch on Words like on Singleton Classes

timor:
Automatically treat words as singletons with regards to method dispatch.  This uses the wrapper syntax `M: \ some-word some-generic ... ;` to get/generate a predicate class which dispatches as if the word had been defined as a singleton class itself.

As an example usage, I added a commit which shows how this could be used in the front-end compiler to implement some properties in a more abstract way.

I imagine this could also be used instead of word properties for a number of cases, possibly making dispatch a bit more efficient because of inline caching.

### DIFF
--- a/basis/compiler/tree/propagation/info/info.factor
+++ b/basis/compiler/tree/propagation/info/info.factor
@@ -86,36 +86,29 @@ UNION: fixed-length array byte-array string ;
         [ [ interval>> empty-interval? ] [ class>> real class<= ] bi and ]
     } 1|| ;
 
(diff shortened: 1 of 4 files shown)


mrjbq7:
That's kinda cool!

The terminology of singleton-class seems to conflict with SINGLETON: syntax, maybe we can find a way to merge or modify that.

timor:
Also, current version breaks `last-word`...

timor:
#2499 is a more thorough approach on this.

PR #2500, opened 2021-12-05
Fix off-by-one error in rolling-map

gifti258:
`{ 1 2 3 4 5 6 7 8 9 } 3 [ >array ] rolling-map ...`
gives
```
{
    { }
    { 1 }
    { 1 2 }
    { 1 2 3 }
    { 2 3 4 }
    { 3 4 5 }
    { 4 5 6 }
    { 5 6 7 }
    { 6 7 8 }
}
```
The first slice is zero elements long, should be one; the last slice doesn't reach the last element of the sequence.
With this change
`{ 1 2 3 4 5 6 7 8 9 } 3 [ >array ] rolling-map ...`
gives
```
{
    { 1 }
    { 1 2 }
    { 1 2 3 }
    { 2 3 4 }
    { 3 4 5 }
    { 4 5 6 }
    { 5 6 7 }
    { 6 7 8 }
    { 7 8 9 }
}
```
which I think is the expected behaviour.


erg:
Looks good, can you please add a unit test that would have failed too?

erg:
Thanks for the fix!

PR #2504, opened 2021-12-20
math.matrices: Fix documentation

gifti258:


erg:
The docs look correct as they are?

The link is to the opposite operation, so diagonal links to anti-diagonal, and vice versa

gifti258:
`<diagonal-matrix>` and `<anti-diagonal-matrix>` create matrices, `main-diagonal` and `anti-diagonal` take diagonals out of matrices.

```
5 [1,b] <diagonal-matrix> dup .
{
    { 1 0 0 0 0 }
    { 0 2 0 0 0 }
    { 0 0 3 0 0 }
    { 0 0 0 4 0 }
    { 0 0 0 0 5 }
}
anti-diagonal .
{ 0 0 3 0 0 }
```
These are opposite operations???

erg:
Got it. Thanks!

catb0t:
I think this was just an error from my copy-pasting originally. Reading those lines again, I'm pretty sure I meant to write "This word is the opposite of **`<diagonal-matrix>`**" for **`<anti-diagonal-matrix>`** and vice versa.

`main-diagonal` and `anti-diagonal` are opposite words to me.

I don't think `anti-diagonal` and `<diagonal-matrix>` were meant to be compared, because one is a constructor and one is a normal function - they are not even inverses in a mathematical sense.

erg:
The docs here could use a ``{ foo bar baz } related-words`` section.

PR #2542, opened 2022-01-04
pcg[-vec]: new vocabs to choose from

alex-ilin:
Please don't merge this PR yet, it's only for review and discussion.

I have implemented the [PCG MWC RNG](https://www.pcg-random.org/) using [this source](https://github.com/tkaitchuck/Mwc256XXA64/blob/main/impl/src/gen64.rs) as a reference. I'm not sure I've made enough mentions in the source to give credit to the original author, please comment. (@tkaitchuck)

There are two alternative implementations in this PR, we must decide which is better, and only difference is the following. The `random.pcg-mwc` vocab is using a `STRUCT:` to keep the state data, while the `random.pcg-mwc-vec` variant keeps the state in an `array`, so its variables are not named and thus may be difficult to trace and compare with the original author's code.

Negatives:
- arrays are allocated in dynamic memory, and should in theory be slower than updating fields of a preallocated `STRUCT:`;
- individual values are not named.

Positives:
- the code with the array is more stacky and flows better (see `:: rot-state`, which shows that in the original code the struct is used a bit like a queue, shifting the values across fields [here](https://github.com/tkaitchuck/Mwc256XXA64/blob/f668e18bdb055c6ae2c3ca74ff587732ff64532c/impl/src/gen64.rs#L90));
- despite the memory concerns, the code with the array runs faster.

Here are my time measurements:
```
USING: random random.pcg-mwc random.pcg-mwc-vec ;
gc 0 0 random.pcg-mwc:<Mwc256XXA64> [ 10,000,000 [ dup random-32* drop ] times ] time drop
gc 0 0 random.pcg-mwc-vec:<Mwc256XXA64> [ 10,000,000 [ dup random-32* drop ] times ] time drop
Running time: 8.5137255 seconds
Running time: 6.3718603 seconds
```
So, the same number of random numbers is generated _faster_ by the `random.pcg-mwc-vec` that uses array for the state. Could it be that I have used suboptimal code in the `STRUCT:` approach? Maybe using `locals` was a mistake? Please, let me know.

Lastly, I did not yet find any reference data to verify the implementation. Theoretically, if someone has Rust installed, they can get the Crate with this RNG and generate some data for a few different seeds. This would give me the test vectors to verify the implementation(s).


alex-ilin:
By the way, I got a bit confused by the unfamiliar terminology, so I'd like to ask: Is `pcg-mwc` a good name for this vocab?

mrjbq7:
Nice!!

I like the ``pcg-mwc`` version and it's faster if you make it a tuple and make ``next-u64`` inline:

```
TUPLE: Mwc256XXA64 x1 x2 x3 c ;
```

mrjbq7:
I might just call it "pcg" since thats what they call their family of random number generators.

mrjbq7:
It seems like since we only need 32 bits of randomness, we could use:

mrjbq7:
Or alternatively save half the generated output for the next random-32* call... that would be twice as fast.

alex-ilin:
> It seems like since we only need 32 bits of randomness
This reminded me of something I wanted to write in the description. Since the code is supposed to be optimized for 32- or 64-bit architecture, shouldn't I somehow detect the current architecture and use the appropriate code? Since we are wanting to use this generator because of its speed, don't we want to avoid 64-bit multiplications in the 32-bit Factor builds?

alex-ilin:
I have replaced `STRUCT:` with `TUPLE:` and added caching of the unused 32 bits between calls and renamed the vocabs. Now the performance is:
- 3.1945688 seconds for `random.pcg`,
- 3.461146 seconds for `random.pcg-vec`.

The two remaining questions are:
- should I use a different implementation for 32-bit Factor builds (PCG-XSH-RS),
- how can I test whether my implementation actually matches the Rust version (produces the same output given the same seeds)?

mrjbq7:
Numpy seems to recommend using PCG over other options:

https://numpy.org/doc/stable/reference/random/performance.html

alex-ilin:
I have added 32-bit implementation, here are my timings (lower value is better):
- Factor x32, 32-bit PCG: 6.3746483 seconds
- Factor x32, 64-bit PCG: 4.1127481 seconds
- Factor x64, 32-bit PCG: 1.8983185 seconds
- Factor x64, 64-bit PCG: 3.277875 seconds

I wonder why non-native implementations are faster. Seeing this, if I didn't make any implementation mistakes, and if we care only about the speed, then we may want to use the 32-bit implementation on 64-bit Factor, especially since we only use the 32 bits per call. But I do like the idea of the larger state behind the 64-bit generator.

erg:
git clone the source repo, `cd tests`,

```rust
use std::error::Error;
use std::io;

use rand_core::RngCore;

use pcg_mwc::Mwc256XXA64;
use pcg_mwc::Mwc128XXA32;

fn main() -> () {
    let mut mwc = Mwc256XXA64::new(1, 1);
    let mut i = 0;
    loop {
        let value = mwc.next_u64();
        if i % 1000000 == 0 {
            println!("i: {}, value: {}", i, value);
        }
        i = i + 1;
    }
}
```

`cargo run test`
```
i: 0, value: 14344626294072084364
i: 1000000, value: 4085314307712647575
i: 2000000, value: 2556289222880332383
i: 3000000, value: 17291746359613858494
i: 4000000, value: 2677539082771571770
i: 5000000, value: 16707736736687949691
i: 6000000, value: 9574364111318242804
i: 7000000, value: 8238894912038295993
i: 8000000, value: 114777213669366784
i: 9000000, value: 14114186591998470906
i: 10000000, value: 4129028049885190726
i: 11000000, value: 2440413087927987898
i: 12000000, value: 6347975883826306324
i: 13000000, value: 16112977322411635731
i: 14000000, value: 14873977780267919144
i: 15000000, value: 15791915102032839566
i: 16000000, value: 13739613705663953241
i: 17000000, value: 11435591774245052903
i: 18000000, value: 2266152047643332339
i: 19000000, value: 6820222934002754768
i: 20000000, value: 7697050046534177581
i: 21000000, value: 12969517220670525441
i: 22000000, value: 12150430512965632859
i: 23000000, value: 15817417532232030388
i: 24000000, value: 13058400660995283575
```

erg:
```rust
erg@Dougs-MacBook-Air.local  tests git:(main) ✗ cat src/main.rs
use std::error::Error;
use std::io;

use rand_core::RngCore;

use pcg_mwc::Mwc256XXA64;
use pcg_mwc::Mwc128XXA32;

fn main() -> () {
    let mut mwc = Mwc128XXA32::new(1, 1);
    let mut i = 0;
    loop {
        let value = mwc.next_u32();
        if i % 1000000 == 0 {
            println!("i: {}, value: {}", i, value);
        }
        i = i + 1;
    }

/*
    let mut mwc = Mwc256XXA64::new(1, 1);
    let mut i = 0;
    loop {
        let value = mwc.next_u64();
        if i % 1000000 == 0 {
            println!("i: {}, value: {}", i, value);
        }
        i = i + 1;
    }
*/
}
erg@Dougs-MacBook-Air.local  tests git:(main) ✗ cargo run test 
warning: unused import: `std::error::Error`
 --> src/main.rs:1:5
  |
1 | use std::error::Error;
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `std::io`
 --> src/main.rs:2:5
  |
2 | use std::io;
  |     ^^^^^^^

warning: unused import: `pcg_mwc::Mwc256XXA64`
 --> src/main.rs:6:5
  |
6 | use pcg_mwc::Mwc256XXA64;
  |     ^^^^^^^^^^^^^^^^^^^^

warning: `pcg-mwc-test` (bin "pcg-mwc-test") generated 3 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/pcg-mwc-test test`
i: 0, value: 3182460383
i: 1000000, value: 1446378418
i: 2000000, value: 3916596035
i: 3000000, value: 1789925569
i: 4000000, value: 3421697034
i: 5000000, value: 3497726190
i: 6000000, value: 3514413066
i: 7000000, value: 2952283670
i: 8000000, value: 1725265268
i: 9000000, value: 2949705070
i: 10000000, value: 832695707
i: 11000000, value: 24882711
i: 12000000, value: 3291302106
i: 13000000, value: 3904013643
i: 14000000, value: 431009355
i: 15000000, value: 1432821264
i: 16000000, value: 2139033935
i: 17000000, value: 872393768
i: 18000000, value: 382832554
i: 19000000, value: 565283590
i: 20000000, value: 56616096
i: 21000000, value: 2362276094
i: 22000000, value: 2238460587
```

erg:
What I've done in the past is add unit tests for the 100,000th or millionth number for each test algorithm with a couple different seeds. Whatever doesn't take too long to compute but exercises most of the code paths should give the right results.

alex-ilin:
@erg, thanks for the data, now I can proof the implementation!

alex-ilin:
> should this be the same first number?

Looks like it should.

mrjbq7:
How's it looking @alex-ilin ?

alex-ilin:
Sorry, no time at the moment. I'll try to find time on the weekend to see what could be the issue.

odiferousmint:
Can I take `xoshiro`[1]? Not sure which versions, perhaps I end up implementing all of them. I will implement it with SplitMix64. I have a C implementation, but I want to go full Factor.

[1] https://prng.di.unimi.it

alex-ilin:
> What I've done in the past is add unit tests for the 100,000th or millionth number for each test algorithm with a couple different seeds. Whatever doesn't take too long to compute but exercises most of the code paths should give the right results.

Hey, @erg!
Could you please produce some test data for me, so that I could add unit-tests?
Above you already gave me the first and the millionth output for the seed (1, 1) of the 32-bit generator.
Could you please do the same for another seed, and then the same for a couple of seeds from the 64-bit generator?

PR #2546, opened 2022-01-09
unix.scheduler: small new vocab to get Unix scheduler info

catb0t:
I couldn't find this functionality in the standard library already. This way, the posix-spawn branch is not creating new indirectly-related vocabularies.

Mac OSX is neglected here, even though I know there must be a few relevant things to implement for it. I know that OSX does not implement the POSIX _Process Scheduling_ optional feature, and does not provide `SCHED_OTHER`, `POSIX_SPAWN_SETSCHEDULER` etc.

OSX `pthreads` do allow POSIX-style scheduling but that's not relevant here.

`$examples` are unchecked because I don't want macosx to break when building. :) 

Commit message follows:

The unix.scheduler vocabulary currently implements a subset of the functions specified by POSIX to exist in the `<sched.h>` header.

This subset is needed to provide important functionality to the upcoming posix-spawn backend for io.launcher and the `<process>` interface.


mrjbq7:
Awesome!

It's a little different in how it loads ``unix.scheduler.platform`` and patches all the words into the ``unix.scheduler`` namespace, but I can see why it's like that.

PR #2553, opened 2022-01-13
misc/vim: Further improve `:FactorVocab` completion

bb010g:
The approach to "vocab:" pathnames is more strict than Factor's, but it also is less likely to infinitely recurse.


mrjbq7:
Cool!

How would i test some of this to see what it does?

mrjbq7:
Awesome, thanks!!

Issue #2577, opened 2022-02-06
Optimizing size of factor.image?


nomennescio:
The current factor.image is about 10x bigger than the factor executable, at about 130MB (for 'core' only I think). Loading more roots and saving the image grows that size considerably. A simple gzip on such an image gives a 6x size reduction, showing the potential for size improvement. An inspection reveals many repeating 0x00, 0x7f, and 0xEA bytes in an uncompressed image.

Is it an idea to support an additional image format which is more optimized for size?
Ideally the image format would be detectable in a header (I didn't check this out yet), but alternatively different filename postfixes can be used to discern, to support multiple formats (at least during a transition period). Even supporting just loading .gz or .zip images could work.

nomennescio:
I noticed that all work is done by primitive `(save-image)`

mrjbq7:
You can use the tools.image.analyzer vocabulary to see what’s in the image.
It’s a lot of generated code, definitions, documentation.

It would be nice to have a smaller footprint.  Ideas are welcome. We
currently mmap into it and then jump into the specified assembly at
locations.

nomennescio:
I'm looking at image.cpp, and do indeed see the dumping/loading of raw memory.
Given that many people have SSDs, compressing data tend to nowadays give a slight performance hit, so startup times might be slighty affected by using a compressed image. Using a compression lib in the vm is extremely straightforward and would just be an optional pre- and postprocessing step.

nomennescio:
Factor already has binding to three native compression libraries:
```
compression.zstd.ffi
compression.zlib.ffi
compression.snappy.ffi
```
But the corresponsing dynamic libraries are not distributed.
Picking one of these and include them in the standard Factor distribution might be a good candidate for a compression lib to handle compressed factor images. I will check these out.

mrjbq7:
I’m not sure compression but itself is a net positive. I’d rather see if we can reduce the image footprint in fundamental ways. 

These images are stored on local file systems, rarely copied over the network, and already slow to mmap when they are large?

mrjbq7:
For example, word properties could be reduced to a bitmap, better string interning might be possible, we should see where the bloat really comes from. 

A load-all image is like 400MB. 

Intuitively, that seems high. But I haven’t spent much time in awhile looking detailed at the image contents.

nomennescio:
[...]

PR #2589, opened 2022-02-09
Feature vm compression support

nomennescio:
Created single file zstd library from v1.5.2 with
`./combine.sh -r ../../lib -o zstd.c zstd-in.c`

Integrated resulting `zstd.c` and `zstd.h` files into `zstd.cpp` and `zstd.hpp` by placing them in namespace `zstd`.
Added `zstd.hpp` to `master.hpp`. Adapted GNUMakefile.


nomennescio:
To support #2577

mrjbq7:
Cool, let us know how the experiments go -- i'm particularly concerned about adding startup delay since we are a lot slower than python/ruby/etc in that use-case due to the mmap image step.

nomennescio:
> Cool, let us know how the experiments go -- i'm particularly concerned about adding startup delay since we are a lot slower than python/ruby/etc in that use-case due to the mmap image step.

If we make compression optional, wouldn't that solve your concern?

mrjbq7:
If it's "not worth it", meaning it's a lot slower than what we have, then it adds surface area, testing, and complexity to just save some space on the file system, when it might make more sense to work on improving the image generation itself.

But I'd love to see what you come up with.

mrjbq7:
This isn't quite a direct comparison, but I want to be able to be used for scripting use-cases, and things like CGI scripts, etc.

```
$ time python3 -c ""    
python3 -c ""  0.02s user 0.01s system 88% cpu 0.035 total

$ time perl -e ""
perl -e ""  0.00s user 0.00s system 70% cpu 0.006 total

$ time ./factor -e=""
./factor -e=""  0.06s user 0.08s system 95% cpu 0.149 total
```

nomennescio:
> This isn't quite a direct comparison, but I want to be able to be used for scripting use-cases, and things like CGI scripts, etc.
> 
> ```
> $ time python3 -c ""    
> python3 -c ""  0.02s user 0.01s system 88% cpu 0.035 total
> 
> $ time perl -e ""
> perl -e ""  0.00s user 0.00s system 70% cpu 0.006 total
> 
> $ time ./factor -e=""
> ./factor -e=""  0.06s user 0.08s system 95% cpu 0.149 total
> ```

If you want to go that way, why not use Factor with an embedded image?

mrjbq7:
The embedded image doesn't help much for startup time unless it's a deployed image stripping out stuff that's unused, it's basically just appended to the binary.

You are right about a deployed CGI script being faster, but "scripting" is a useful use-case... it might not be our sweet spot however I'd like to not make it worse. That's all.  I do like the stuff you're looking into however!

mrjbq7:
You could profile ``load_image`` and see which part of it is slow and think about improvements?

nomennescio:
I still need to rewrite existing code to support decompressing, even for non-compressed images. I can currently only compare on a Virtual Machine right now, so I'm not sure how representative any performance figures are.

mrjbq7:
You could start by checking the file extension and if its ``.image.z`` then uncompressing as part of the load, so it doesn't require any header changes...

nomennescio:
What I mean is that I have to rewrite the whole load_image to not use `FILE*` at all.

nomennescio:
> The embedded image doesn't help much for startup time unless it's a deployed image stripping out stuff that's unused, it's basically just appended to the binary.

Depends on how user timing is calculated; usually a program is loaded into memory, then started, and that's often the moment performance is measured, whereas loading 100+MB during your program is counted towards runtime.

> You are right about a deployed CGI script being faster, but "scripting" is a useful use-case... it might not be our sweet spot however I'd like to not make it worse. That's all. I do like the stuff you're looking into however!

If you want to optimise for scripting, you should take into account that many scripting languages require loading of libraries after starting. Furthermore, the default loaded Factor library is already very big. I don't think that gives a fair comparison.

I think a better comparison would be to compare how long different common scripting tasks take to execute between different languages.

nomennescio:
As for compression, if it's optional, it would at least give the benefit to people to whom disk space occupied by images does matter, at the cost of some additional code in the VM (or a binary library distributed with Factor).

I noticed that Linux uses zstd at many places, including booting, to e.g. decompress boot images

nomennescio:
clang generates warning for `fallthrough` attribute, whereas `zstd.c` explicitly checks if the compiler supports it:
```
#if defined(__cplusplus) && defined(__has_cpp_attribute)
# define ZSTD_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
#else
# define ZSTD_HAS_CPP_ATTRIBUTE(x) 0
#endif
```
```
#ifndef ZSTD_FALLTHROUGH
# if ZSTD_HAS_C_ATTRIBUTE(fallthrough)
#  define ZSTD_FALLTHROUGH [[fallthrough]]
# elif ZSTD_HAS_CPP_ATTRIBUTE(fallthrough)
#  define ZSTD_FALLTHROUGH [[fallthrough]]
```

Seems like a problem with clang to me.

nomennescio:
@mrjbq7 

Here are preliminary results, using zstd internal allocation of a temporary buffer.

```
/media/sf_project/software/factor.fork$ time for i in {1..100};do ./factor -i=factor.image -e="";done

real    0m24.924s
user    0m8.939s
sys     0m2.520s
/media/sf_project/software/factor.fork$ time for i in {1..100};do ./factor -i=factor.image.compressed -e="";done

real    0m27.851s
user    0m22.229s
sys     0m2.106s
```

That's on a VirtualBox, whose performance is hard to predict in terms of filesystem and memory performance, for a standard Factor image, on Debian. Notice the 6.8 compression factor!

```
-rwxrwx--- 1 root vboxsf 127529144 Feb 15 18:32 factor.image
-rwxrwx--- 1 root vboxsf  18801774 Feb 15 18:32 factor.image.compressed
```

I'm trying to get in-place decompression to work, but the API is currently not behaving as I expect it to.

nomennescio:
Basis + core
compression factor : 6.6
```
-rwxr-xr-x+ 1 192485800 Feb 15 18:58 factor.basis+core.image
-rwxr-xr-x+ 1  29304829 Feb 15 18:59 factor.basis+core.image.compressed
```

```
/media/sf_project/software/factor.fork$ time for i in {1..100};do ./factor -i=factor.basis+core.image -e="";done

real    0m33.658s
user    0m12.066s
sys     0m3.233s
/media/sf_project/software/factor.fork$ time for i in {1..100};do ./factor -i=factor.basis+core.image.compressed -e="";done

real    0m40.807s
user    0m33.082s
sys     0m2.777s
```

nomennescio:
Compression factor 7.4
```
-rwxrwx--- 1 root vboxsf 459713896 Feb 15 19:21 factor.basis+core+extra.image
-rwxrwx--- 1 root vboxsf  61979751 Feb 15 19:39 factor.basis+core+extra.image.compressed
```

```
/media/sf_project/software/factor.fork$ time for i in {1..100};do ./factor -i=factor.basis+core+extra.image -e="";done

real    1m15.754s
user    0m26.309s
sys     0m6.948s
/media/sf_project/software/factor.fork$ time for i in {1..100};do ./factor -i=factor.basis+core+extra.image.compressed -e="";done

real    1m38.515s
user    1m21.445s
sys     0m6.796s
```

75 seconds versus 98 seconds.

nomennescio:
And of course, I would like to see these benchmarks run on real systems, not just a VirtualBox...

mrjbq7:
Ideally the Factor VM wouldn't have a larger memory footprint, mmap should make it so we only "use" the parts of the image that we need. The new way would have a bigger footprint because the uncompressed thing effectively uses max(image_size)?

nomennescio:
Getting ZSTD to allocate as little memory as possible, and to avoid memcpy-s turns out to be quite experimental, and very non-trivial. But this is exactly where the expected performance gain is to be had; as we know EXACTLY the sizes of the decompressed data, which we can allocate upfront, reusing the current Factor data and code heaps.

nomennescio:
Running on Windows laptop with Cygwin:
```
$ time ./factor.com -i=factor.image -e='"Hello World!" print'
```
versus
```
$ time ./factor.com -i=factor.image.compressed -e='"Hello World!" print'
Hello World!
```
Run each command 10 times, record real time

Uncompressed image: 0.881 s
Compressed image: 0.942 s

So it takes **_61 ms_** to decompress a full Factor image (and that's with copying of a full buffer, which can be improved upon with in-place decompression)

Now look at the sizes of the images:
```
129002248 Aug 28 14:49 factor.image
 19084631 Aug 28 15:27 factor.image.compressed
```
I.e. a compression factor of _**6.76**_

nomennescio:
Compression also works on boot images:
```
4563240 Aug 28 14:44 boot.windows-x86.64.image
 546489 Aug 28 16:15 boot.windows-x86.64.image.compressed
```

Compression factor of  _**8.4**_

nomennescio:
To be clear; the compression is encoded in the header of the file, not in its filename, so the `.compressed` postfix can be removed

PR #2590, opened 2022-02-10
Refactor VM (untyped) defines into (typed) enum

nomennescio:
Refactor code to replace defines acting as an enum into a proper (typed) enum.
Refactor code to replace switch-case idiom with proper datastructure to convert to `const char*`

static inline with static local will share static local among (inline) static instances per translation unit


nomennescio:
That's fast!

Issue #2580, opened 2022-02-07
"Bug" in vm/image.cpp:factor_vm::save_image


nomennescio:
In vm/image.cpp:factor_vm::save_image there's a subtle bug hidden:
```
bool factor_vm::save_image(const vm_char* saving_filename,

  image_header h;

  h.magic = image_magic;
  h.version = image_version;
  h.data_relocation_base = data->tenured->start;
  h.data_size = data->tenured->occupied_space();
  h.code_relocation_base = code->allocator->start;
  h.code_size = code->allocator->occupied_space();
```

The `image_header` is then directly written to disk. However, the `image_header h;` does NOT initialize all fields to 0. Therefore the currently unused fields `reserved_1 reserved_2 reserved_3 reserved_4` get random values. This breaks potential backwards compatibility opportunities, might newer image files decide to start using these fields while retaining the image_version to ensure backwards compatibility.

Simple solution:
```
image_header h = {};
```

nomennescio:
It's common practice to initialize reserved fields with 0 to guarantee forwards compatibility.

mrjbq7:
Nice find!

nomennescio:
@mrjbq7 I checked actual "factor.image" files headers. All had different and random values in the reserved_* fields. Now that the code is changed, that should not happen anymore. I cannot think of all scenarios that might be impacted, but I can imagine some old image file being loaded by a newer version of factor. It cannot assume these fields are 0. Therefore existing factor images will NOT be forward compatible if these fields are going to get used in the future without changing the image version. That of course will make such image incompatible with existing Factor executables.

I'm not sure how big of an issue it is to bump the image version (the last time that happened was in 2006), but I can imagine it's to be avoided. It is possible to extend the current image format while keeping the image version at 4; the area after the code section can store unrestricted data, and additional headers and data can be stored there, without breaking current code, and such can be detected by a new version of load_image. Before going there, it probably is best to discuss any current plans/ideas/wishes for what goes into a new image format, e.g. saving space (#2577)

nomennescio:
The issue was introduced with 1a72f731e03739fe8cf3cec902cfad7448dfaebf already in 2015, therefore all existing version 4 images after that have random values in these fields.

PR #2591, opened 2022-02-10
VM prepare image reserved fields usage

nomennescio:
The version 4 image format did not write 0 into reserved_* fields (#2580).
Therefore current version 4 images are not forward compatible with image versions that use these fields.
By introducing an escape code 0 into the data_size field,
three reserved_* fields now become available for redefinition, while keeping image version at 4.

The image.cpp changes can transparently read regular v4 image files, as well as ones with h.data_size=0 and h.escaped_data_size=data_size.


nomennescio:
I can imagine your viewpoint. Version 5 version bump might be most appropriate, this is more like a premature contract while keeping all options open and staying at version 4, while only explicitly declaring how fields are to be interpreted which is compatible with current ways. I was hoping for a kind of baseline to build upon, but we can also postpone such a decision.

nomennescio:
Made the code change less intrusive, with the same result.

mrjbq7:
Would you like this merged while you are also considering what other changes are needed?

nomennescio:
> Would you like this merged while you are also considering what other changes are needed?

Yes please. It doesn't break existing code, and we can still change it later, as we nowhere explicitly save into this format. As a next step for my compression proposal on my other experimental branch, I intend to use reserved_2 and reserved_3 for `compressed_data_size` and `compressed_code_size`, which requires some non-intrusive image header change like this.

I can also just put everything in an experimental branch, but I like to keep feature/experimental branches small. 

You can decide.

nomennescio:
Note that in a valid version 4 image file, actual `data_size` can never be 0 due to an assert in `raw_fread`. We will use this invariant as an escape value.

PR #2593, opened 2022-02-13
system-info: implement system-info.freebsd via sysctl

twopir:
This commit also moves the sysctl interface into unix.sysctl to share
it between OS X and FreeBSD.


mrjbq7:
Wooooo! Thanks, merged!

PR #2595, opened 2022-02-20
add color>hex

twopir:
found basically the same thing in the color-table extra, but this one can
round-trip an alpha channel.


erg:
Thank you!

PR #2596, opened 2022-02-23
put show-error-list in the touchbar

twopir:
avoids having to pop open Keyboard Viewer to hit the virtual F3 key under Touchbar settings that show Control Strip and App Controls only.


erg:
Merged. Thanks!

PR #2601, opened 2022-03-27
mediawiki.api: Add MediaWiki API

gifti258:


extra/mediawiki/api/api-docs.factor
```diff
@@ -0,0 +1,196 @@
...
+"} api-call"
+"\"query\" \"userinfo\" \"name\" [ of ] tri@" }
+{ $code " { { \"meta\" \"userinfo\" } } query \"name\" of" }
+$nl
+"The following two code snippets are also equivalent:"
+{ $code
+"{"
+"    { \"action\" \"query\" }"
+"    { \"list\" \"watchlistraw\" }"
+"} api-query"
```
mrjbq7, line 111: do you mean ``api-call`` here?

extra/mediawiki/api/api.factor
```diff
@@ -0,0 +1,216 @@
...
+IN: mediawiki.api
+
+TUPLE: oauth-login consumer-token consumer-secret access-token
+access-secret ;
+TUPLE: password-login username password ;
+
+C: <oauth-login> oauth-login
+C: <password-login> password-login
+
+SYMBOLS: basetimestamp endpoint botflag contact cookies
```
mrjbq7, line 16: It's no big deal, but perhaps these could be more readable with something like ``base-timestamp`` or ``bot?``?

Another idea might be to prefix them as we do in a few places, so ``mediawiki-bot?`` and ``mediawiki-contact``, etc?

Just an idea...

extra/mediawiki/api/api.factor
```diff
@@ -0,0 +1,216 @@
...
+: retry-after? ( response -- ? )
+    header>> "retry-after" of dup [ dup seconds sleep ] when ;
+
+: nonce-already-used? ( assoc -- ? )
+    "error" of
+    [ "code" of "mwoauth-invalid-authorization" = ]
+    [ "info" of "Nonce already used" swap subseq-start ] bi
+    and ;
+
+: readonly? ( assoc -- ? )
```
mrjbq7, line 104: Maybe the "is this readonly?" code could be separate from "sleep-if-readonly".

extra/mediawiki/api/api.factor
```diff
@@ -0,0 +1,216 @@
...
+        "warnings" "errors" [ over at [ ... ] when* ] bi@
+    ] do while 2nip ;
+
+<PRIVATE
+
+:: (query) ( params -- obj assoc )
+    { { "action" "query" } } params assoc-union api-call dup
+    dup "query" of [ nip ] when*
+    "siprop" params key? [
+        params "prop" "list" "meta" [ of ] tri-curry@ tri or or
```
mrjbq7, line 138: ``params { "prop" "list" "meta" } [ of ] map-find drop``

Maybe?

extra/mediawiki/api/authors.txt
```diff
@@ -0,0 +1 @@
+Giftpflanze
```
mrjbq7, line 1: Would you prefer to be a named author or a nicknamed author? Either is fine.

If a nickname, maybe use ``@giftpflanze``?

extra/mediawiki/api/api-docs.factor
gifti258: yes, thanks for catching it.

extra/mediawiki/api/api.factor
gifti258: I think those are good ideas.

mrjbq7:
I'm ready to merge it when you are!

extra/mediawiki/api/authors.txt
gifti258: I didn't want to use my real name because this would mean deanonymizing my wikipedia username. I think I will still go by Giftpflanze.

mrjbq7:
Great!

Do you want it to be “mediawiki” or did you have plans for that base vocab vs “mediawiki.api”?

gifti258:
> Do you want it to be “mediawiki” or did you have plans for that base vocab vs “mediawiki.api”?
If have a package named mediawiki.parser, a wikitext parser/replacer, but it don't deem it ready for release yet.

extra/mediawiki/api/api.factor
gifti258: Shouldn't we then also separate `code-200?` and `retry-after?`?

mrjbq7: If you want?  I was thinking that the sleep is part of a user control flow separate from request/response logic, but it's just thinking out loud.

gifti258: OK, I don't see how this could be meaningfully separated then.

extra/mediawiki/api/api.factor
gifti258:
```
params "prop" "list" "meta" [ of ] tri-curry@ tri or or
params { "prop" "list" "meta" } [ of ] with map-find drop
params { "prop" "list" "meta" } values-of sift first
params { "prop" "list" "meta" } intersect-keys values first
```
I think I will use the third one

mrjbq7:
Thanks!

PR #2602, opened 2022-03-27
db.mysql: Add low-level MySQL support

gifti258:


mrjbq7:
Does this inherit from some of the unmaintained code?

https://github.com/factor/factor-unmaintained/tree/main/db/mysql

Or the "db4" and "db5" branches that refactored the DB libraries?

https://github.com/factor/factor/tree/db4/basis/mysql

https://github.com/factor/factor/tree/db5/basis/mysql

It doesn't have to -- I'm just curious.  The DB4 and DB5 vocabularies aren't quite ready so don't worry about matching their interface -- better to have it working on the ``master`` branch.

gifti258:
I started with the unmaintained code. I didn't even know about the db4 and db5 branches.

mrjbq7:
I'm not sure if the db4 and db5 mysql codes work at all, so that's fine!

Do you want to keep adding to this PR or merge and then continue improving over time?

gifti258:
> Do you want to keep adding to this PR or merge and then continue improving over time?

What needs to be added?

mrjbq7:
Some tests similar to:

https://github.com/factor/factor/blob/master/basis/db/postgresql/postgresql-tests.factor

Maybe some error parsing to convert strings to factor errors:

https://github.com/factor/factor/blob/master/basis/db/postgresql/errors/errors.factor

Maybe some column type conversions

https://github.com/factor/factor/blob/master/basis/db/postgresql/lib/lib.factor#L154

mrjbq7:
I'm happy to merge it at any stage, that's just what I'd think to work on to shift it to the same usefulness as our SQLite and PostgresQL bindings

gifti258:
> I'm happy to merge it at any stage, that's just what I'd think to work on to shift it to the same usefulness as our SQLite and PostgresQL bindings

My thinking was that no one would want to use MySQL as a backend when you can have SQLite. I just wanted to have an interface to existing MySQL databases at all. Therefore I didn't attempt to achieve higher-level db.tuples integration (which would also be too much work and too difficult for my level). I agree that tests would be nice, and I also think documentation would be nice. I could do both at a later time, or someone else. For now, I think it should be merged as is.

mrjbq7:
Great, thanks!

PR #2614, opened 2022-06-18
fix map-index always producing array

Kacarott:


mrjbq7:
Well, this makes it match the documentation which says "Collects the outputs of the quotation and outputs them in a sequence of the same type as the input sequence.".

Hopefully it doesn't cause anyone problems who, for example, ``map-index`` on a string and expect it to produce an ``array``.

mrjbq7:
Thanks!

PR #2615, opened 2022-06-18
Allow `fuel-start-remote-listener` to work with a specified port

mariari:
The `integer?` check in `fuel-start-remote-listener` turns the integer into a boolean, causing `<tty-server> start-server` to error

```
Bad store to specialized slot
value t
class integer
``` 

removing the `integer?` check allows the function to work as intended


mrjbq7:
Interesting, it's been that way since 2009 (f34f7298eec95d54ab53e61bfb101802dcc97761).

mrjbq7:
Thanks!

PR #2637, opened 2022-07-23
Refactor core july21

erg:
To see the diff


core/sequences/sequences.factor
```diff
@@ -447,51 +458,43 @@ PRIVATE>
 : (3each) ( seq1 seq2 seq3 quot -- n quot' )
     [ setup-3each ] dip compose ; inline
 
-: finish-find ( i seq -- i elt )
-    over [ dupd nth-unsafe ] [ drop f ] if ; inline
-
-: (find) ( seq quot quot' -- i elt )
-    pick [ [ (1each) ] dip call ] dip finish-find ; inline
+: element/index ( i/f seq -- elt/f i/f )
+    [ maybe-nth ] [ drop ] 2bi ; inline
```
mrjbq7, line 462: instead, how about we get rid of ``maybe-nth`` and fix stack effect for ``index/element``:

```factor
: element/index ( i/f seq -- elt/f i/f )
    '[ [ _ nth ] [ f ] if* ] keep ;

: index/element ( i/f seq -- i/f elt/f )
    dupd '[ _ nth ] [ f ] if* ;
```

core/assocs/assocs.factor
```diff
@@ -30,15 +30,15 @@ M: assoc assoc-like drop ; inline
 
 <PRIVATE
 
-: (assoc-each) ( assoc quot -- seq quot' )
+: assoc-operator ( assoc quot -- alist quot' )
     [ >alist ] dip [ first2 ] prepose ; inline
 
-: (assoc-stack) ( key i seq -- value )
+: search-assoc-stack ( key i seq -- value/f )
```
mrjbq7, line 36: rename to ``assoc-stack-from``

Issue #2643, opened 2022-08-08
"1 2 3" infer should error


erg:
```factor
"1 2 3" infer .
( -- x x x x x )
```

twopir:
```factor
IN: scratchpad "abc" infer .
( -- x x x )
IN: scratchpad { 1 2 3 } infer .
( -- x x x )
```

looks like sequences, in general, are inferred as though their individual elements all become stack items. should it just be prevented on strings?

mrjbq7:
I don’t think it makes sense to infer anything that’s not a word or callable. We could make it throw I guess.

erg:
Yeah. I was trying to infer a quotation and noticed it didn't fail for a string like I expected.

twopir:
Making it throw an error breaks bootstrap, alas; it seems like something in the class/throwing infrastructure isn't set up when trying to throw.

Trying this instead:

```factor
: infer ( quot -- effect )
    dup callable? not
    [ drop ( -- x ) ]
    [ [ infer-quot-here ] with-infer drop ] if ;
```

This infers all things that aren't callable to be self-evaluating, instead of inferring a sequence as something which stacks all its elements.

mrjbq7:
Have to check words too for e.g., ``\ + infer``

twopir:
oh, good catch.

twopir:
turns out really only callables should be inferred; words are handled by `M: word stack-effect`.

The failing windows test in PR #2649 is bootstrap failure due to missing an image; not caused by this.

PR #2649, opened 2022-08-10
don't infer things that aren't callable

twopir:
Addresses #2643

### DIFF
--- a/basis/stack-checker/stack-checker.factor
+++ b/basis/stack-checker/stack-checker.factor
@@ -9,7 +9,9 @@ stack-checker.visitor.dummy ;
 IN: stack-checker
 
 : infer ( quot -- effect )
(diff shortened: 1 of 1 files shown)


mrjbq7:
I might be inclined to do something like 

```
callable check-instance
```

And have it throw ``not-an-instance``

twopir:
```
$ ./factor
Factor 0.99 x86.64 (2176, heads/alexm/infer-this-4130bc707d, Aug 10 2022 19:58:07)
[Clang (GCC Apple LLVM 12.0.0 (clang-1200.0.32.2))] on macosx
IN: scratchpad "x" infer 
not-an-instance
obj   "x"
class callable

Type :help for debugging help.
IN: scratchpad [ "x" ] infer .
( -- x )
```

mrjbq7:
Thanks!

PR #2652, opened 2022-08-12
teach hooks how to delegate, by abstracting consultation-quotation construction

twopir:
This allows for delegation of hooks and their associated singletons,
working around their inability to be subclassed.


mrjbq7:
Looks pretty good!

I was hoping somehow it could auto-detect GENERIC: vs HOOK: and handle it transparently so you don't have to specify the ``hook-var`` in the ``HOOK-CONSULT:`` usage, but I haven't looked to see if it's reasonably doable.

twopir:
we couldn't see a reasonable way to go from "a singleton that's in a hook" back to the hook-var itself, at least by introspecting on the word-props.

twopir:
per Discord,

> the property is already there, it's the combination word property on a generic, it's a hook-combination if it's a hook and it contains the var being dispatched on 

so with that, we ought to be able to eliminate the need to specify the variable as well. Shall hack on this today.

theAlexes:
"today" being many months later, i guess...

mrjbq7:
Merged, thanks!!

PR #2660, opened 2022-08-20
sequences.repeating: Fix repeat behaviour

gifti258:
The current behavior is confusing:

```
{ 1 2 3 } 2 repeat
{ 1 1 2 2 3 3 }
```

I propose to change it to the following, consistent with the behavior of `cycle`:

```
{ 1 2 3 } 2 repeat
{ 1 2 3 1 2 3 }
{ 1 2 3 } 6 cycle
{ 1 2 3 1 2 3 }
```


mrjbq7:
If you do that, does it make sense to have one virtual sequence implementation and have repeat do something like “ over length * cycle”, or vice versa.

gifti258:
`repeat` should probably be kept and renamed to e.g. `repeat-elements` like in the second commit.

mrjbq7:
👍 for rebase and merge!

PR #2661, opened 2022-08-22
Vim highlighting update

olus2000:
I have had some issues using the vim highlighting tools that come with Factor. I tracked those issues down to two causes: wrongly defined commands and lack of a `iskeyword` option set in the syntax definition file. These have been fixed and tested on vim 9.0.135.


mrjbq7:
Thanks!!

PR #2666, opened 2022-08-24
Put brackets around ipv6 addresses in `inet6 present`

twopir:
This disambiguates host from port. Another option is to use `.` to separate host from port for ipv6, like some netstats do.


mrjbq7:
Thanks!

PR #2669, opened 2022-08-25
Refactor sequences

erg:
I tried to make sequences words better by factoring and renaming and rearranging the stack effects. This branch is starting to diverge from master so I'd rather try to get this code into master and fix any inconsistencies or critiques.


mrjbq7:
Not a huge fan of all the ``seq-`` and ``-of`` words for 0.99?

The other cleanups look great!

PR #2680, opened 2022-09-08
Block caret code

tgunr:
Introduces 3 kinds of caret shapes: line, box, filled box

(cherry picked from commit 749ecba7c3e80891109273ff714e993afbe003c0)

# Conflicts:
#	basis/ui/gadgets/editors/editors.factor

### DIFF
--- a/basis/ui/gadgets/editors/editors-docs.factor
+++ b/basis/ui/gadgets/editors/editors-docs.factor
@@ -40,16 +40,15 @@ HELP: caret-is-shape
 $nl
 { $code
 "  IN: ui.gadgets.editors"
-"  2 caret-is-shape set"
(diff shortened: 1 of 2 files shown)


tgunr:
Probably needs some tweaking John.

tgunr:
Agreed, was looking for something like ENUM but didn't realize you can use SYMBOL without values. Checking in code shortly.

mrjbq7:
Cool! Thanks!

mrjbq7:
I might make one tweak to the docs and I might rename ``caret-is-shape`` to ``caret-shape``, just a second.

tgunr:
I thought I was having a name conflict so went back to caret-is-shape as the tuple has caret-shape.

PR #2682, opened 2022-09-08
 Add breakpoint and hit counters

tgunr:
Found these useful on occasion.

### DIFF
--- a/basis/tools/continuations/continuations.factor
+++ b/basis/tools/continuations/continuations.factor
@@ -26,6 +26,31 @@ SYMBOL: break-hook
 
 \ break t "break?" set-word-prop
 
+
+SYMBOL: break-counter
+
+: break-counter-get ( -- n )
+    break-counter get ;
+
+: break-count-zero ( -- )
+    0 break-counter set ;
+
+: break-count ( -- )
+    break-counter get dup
+    [  1 + ]
+    [ drop 1 ] if
+    break-counter set ;
+
+: break-count= ( n -- )
+    break-count
+    break-counter get swap >=
+    [     continuation callstack >>call
+    break-hook get call( continuation -- continuation' )
+    after-break  ]
+    [ ] if ;
+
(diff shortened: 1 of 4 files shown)


basis/tools/continuations/continuations.factor
```diff
@@ -26,6 +26,47 @@ SYMBOL: break-hook
 
 \ break t "break?" set-word-prop
 
+SYMBOL: hit-counter
+
+: (hit-count-inc) ( -- )
+    hit-counter get dup
```
mrjbq7, line 32: Probably should just be ``hit-counter inc``

basis/tools/continuations/continuations.factor
```diff
@@ -26,6 +26,47 @@ SYMBOL: break-hook
...
+    [  1 + ]
+    [ drop 1 ] if
+    hit-counter set ;
+
+: hit-count-reset ( -- )
+    0 hit-counter set ;
```
mrjbq7, line 38: I'm not sure about how this scoping will work, this sets it in the current namespace, and then the other words "get/set" it into the new namespace, so if your word does things like ``with-variable`` that binds a new namespace then maybe this won't work like you expect?

basis/tools/continuations/continuations.factor
tgunr: Yes, better name. I wrote this a long time ago.

mrjbq7: Not just a name, you can use ``inc`` from namespaces

basis/tools/continuations/continuations.factor
tgunr: Yes, you're right, I just checked it doesn't work unless you remain in the current namespace. Suggestions?

tgunr: On second thought, it's probably not worth pursuing. Just close the request. If I need it again will address it then.

PR #2686, opened 2022-09-14
Add absolute-path to normalize any new root path. E.G. permit ~user

tgunr:
User should be able to use tilde in path.


mrjbq7:
Merging, but we'll need to upload new boot images to fix the builders.

PR #2693, opened 2022-09-29
promote unicode.control-pictures to basis

twopir:


erg:
Looks good!

erg:
@twopir There are more control-pictures to support!

Screen Shot 2022-09-29 at 17 17 22

PR #2699, opened 2022-10-06
Fix `when-empty`

alex-ilin:
According to `if-empty`, the second quotation should have effect `( ..a seq -- ..b )`, but `when-empty` supplies `[ ]`.
It should be `[ drop ]` instead.


alex-ilin:
This will not work, because the code using the word is not updated here.

mrjbq7:
Doesn’t the default quotation cause the seq to stay on the stack?

alex-ilin:
According to the declared stack effect (`when-empty ( seq quot -- )`), calling
`seq [ ] when-empty` should be equivalent to
`seq empty? [ ] when`, but it's actually equivalent to
`dup seq empty? [ ] when`.

alex-ilin:
Edit: I meant `seq dup`, not `dup seq` in the last line above.

PR #2700, opened 2022-10-06
sequences: fix stack effect declarations for when-empty and unless-empty

alex-ilin:
Consider the updated stack effects. Are they more accurate?

### DIFF
--- a/core/sequences/sequences.factor
+++ b/core/sequences/sequences.factor
@@ -31,9 +31,8 @@ M: sequence shorten 2dup length < [ set-length ] [ 2drop ] if ; inline
 : if-empty ( ..a seq quot1: ( ..a -- ..b ) quot2: ( ..a seq -- ..b ) -- ..b )
     [ dup empty? ] [ [ drop ] prepose ] [ ] tri* if ; inline
 
-: when-empty ( seq quot -- ) [ ] if-empty ; inline
-
-: unless-empty ( seq quot -- ) [ ] swap if-empty ; inline
+: when-empty ( ... seq quot: ( ... -- obj ) -- ... seq/obj ) [ ] if-empty ; inline
+: unless-empty ( ... seq quot: ( ... seq -- ... ) -- ... ) [ ] swap if-empty ; inline
 
 : delete-all ( seq -- ) 0 swap set-length ;
 


mrjbq7:
What is “v”?And we may want it to be ( … seq quot: ( … — … seq ) — … seq ) since it can dip under the stack I think “ABC” f [ dup ] when-emptyOn Oct 6, 2022, at 1:31 PM, Alexander Ilin ***@***.***> wrote:﻿Consider the updated stack effects. Are they more accurate?



  https://github.com/factor/factor/pull/2700

Commit Summary

  f376f28  sequences: fix stack effect declarations for when-empty and unless-empty


File Changes  (1 file)

  
    M
    core/sequences/sequences.factor
    (5)
  


Patch Links:

  https://github.com/factor/factor/pull/2700.patch
  https://github.com/factor/factor/pull/2700.diff

alex-ilin:
How about now?

mrjbq7:
***@***.***?On Oct 6, 2022, at 1:44 PM, Alexander Ilin ***@***.***> wrote:﻿
How about now?

alex-ilin:
Good point! : )

alex-ilin:
I have updated the PR.

mrjbq7:
I’m not near a computer, but do the -docs.factor have to be updated also?On Oct 6, 2022, at 1:48 PM, Alexander Ilin ***@***.***> wrote:﻿
I have updated the PR.

alex-ilin:
The docs are fine as they are. They only talk about these words being variants of `if-empty`, so updating the stack effects are the last bit of clarification that's needed.

alex-ilin:
This PR is ripe for the taking!

mrjbq7:
Should merge it soon. I’m traveling without a laptop. Maybe @erg can. On Oct 7, 2022, at 1:50 PM, Alexander Ilin ***@***.***> wrote:﻿
This PR is ripe for the taking!

core/sequences/sequences.factor
```diff
@@ -31,9 +31,8 @@ M: sequence shorten 2dup length < [ set-length ] [ 2drop ] if ; inline
 : if-empty ( ..a seq quot1: ( ..a -- ..b ) quot2: ( ..a seq -- ..b ) -- ..b )
     [ dup empty? ] [ [ drop ] prepose ] [ ] tri* if ; inline
 
-: when-empty ( seq quot -- ) [ ] if-empty ; inline
-
-: unless-empty ( seq quot -- ) [ ] swap if-empty ; inline
+: when-empty ( ... seq quot: ( ... -- seq ) -- ... seq ) [ ] if-empty ; inline
```
erg, line 34: The stack effect for `when-empty` implies that it's the same `seq` that gets output but that's not the case since it's dropped.

What about:
```
: when-empty ( ... seq quot: ( ... -- obj ) -- ... obj ) [ ] if-empty ; inline
```

erg: or `seq'` if we assume these output a seq

alex-ilin: This implies that the output is always the result of the quotation, but in some cases it'll be the original `seq`.

alex-ilin: The following was suggested in Discord
`: when-empty ( ... seq quot: ( ... -- v ) -- ... seq/v ) [ ] if-empty ; inline`

alex-ilin: How about this as a compromise?
`: when-empty ( ... seq quot: ( ... -- seq' ) -- ... seq/' ) [ ] if-empty ; inline`

alex-ilin: Or maybe `: when-empty ( ... seq quot: ( ... -- obj ) -- ... seq/obj ) [ ] if-empty ; inline`?

erg: `seq/obj` is the best so far?

alex-ilin: Pushed.

erg:
Merged and fixed help lint and some weird spacing. Thanks!

PR #2702, opened 2022-10-15
fix(semver): parsing semver is too permissive

rgrinberg:
The previous function would parse semver such as "1.2.3.4". The new implementation uses EBNF and is more conservative.

### DIFF
--- a/extra/semver/semver.factor
+++ b/extra/semver/semver.factor
@@ -26,11 +26,14 @@ TUPLE: semver
     { prerelease initial: "" }
     { build initial: "" } ;
 
-: parse-semver ( str -- semver )
-    "+" split1
-    [ "-" split1 ] dip
-    [ "." split [ string>number ] map first3 ] 2dip
-    semver boa ;
+EBNF: parse-semver [=[
+    nr      = [0-9]+ => [[ string>number ]]
+    part    = nr | [-0-9A-Za-z]+ => [[ >string ]]
+    pre     = '-'~ part
+    bld     = '+'~ part
+    version = nr '.'~ nr '.'~ nr pre? bld?
+              => [[ [ first4 ] [ 4 swap nth ] bi semver boa ]]
+]=]
 
 : <semver> ( str -- semver ) parse-semver ; inline
 


erg:
I added some more test cases and fixed what you found.

I'm not big on EBNF when you can just use split. Maybe if EBNF is faster?

```
EBNF: ebnf-parse-semver [=[
    nr      = [0-9]+ => [[ string>number ]]
    part    = nr | [-0-9A-Za-z]+ => [[ >string ]]
    pre     = '-'~ part
    bld     = '+'~ part
    version = nr '.'~ nr '.'~ nr pre? bld?
              => [[ [ first4 ] [ 4 swap nth ] bi semver boa ]]
]=]

! failing EBNF test case
"1.2.3-erg.0" ebnf-parse-semver
unable-to-fully-parse-ebnf
remaining ".0"


! after my patch:
"1.2.3.4" parse-semver
malformed-semver
parts { 1 2 3 4 }

"1.2.3.4-erg.0" parse-semver
parts { 1 2 3 4 }

```

rgrinberg:
It can be fixed without ebnf but I find it concerning that since range parsing still uses ebnf, there will always be these slight inconsistencies.

erg:
I'm sure it's not perfect yet, but does EBNF in both parts fix anything? There must be some errors in the parse-range EBNF with all those lines. And why are half my unit tests commented out? XD (Oh, looking again, I got them from parsing nodejs package.json files I think and did not handle malformed stuff like `v1.2.3`.

mrjbq7:
Why do we have two semantic versioning vocabs!

rgrinberg:
> Why do we have two semantic versioning vocabs!

What's the second one?

Okay, fixing the old parser is fine by me.

mrjbq7:
extra/semvar and extra/semantic-versioningOn Oct 15, 2022, at 6:42 PM, Rudi Grinberg ***@***.***> wrote:﻿

Why do we have two semantic versioning vocabs!

What's the second one?
Okay, fixing the old parser is fine by me.

Issue #2641, opened 2022-08-01
Proposal: aliases for terse bitwise operators


nomennescio:
Maybe it's just me, but I find myself resorting to bit operators quite often, and am longing for a more terse representation, such as in C. However, some symbols are already in use, and it's better to avoid nameclashes.

I therefore propose the following terse bit operators, each start with `.` (pronounce: "bit"), followed by the familiar C operator

```
ALIAS: .& bitand
ALIAS: .| bitor
ALIAS: .^ bitxor
ALIAS: .~ bitnot
```

And the following abbrev for left shift is strangely missing too:
```
: 2* ( n -- 2n ) 1 shift ; inline
```

Unfortunately, shift operators `<<` and `>>` are in use by `syntax`, which is always present, therefore
```
ALIAS: .<< shift
: .>> ( x n -- x/2^n ) neg shift ; inline
```

If you like this, I can prepare a pull request.

twopir:
we'd appreciate all of these, especially since we can never remember the directionality of `shift`

razetime:
since `.` is already under use by many formatting words, qould it make sense to have the words have a `b` postfix? (akin to `/i` and similarly named words)

giving us `&b`, `|b` and so on.

nomennescio:
> since `.` is already under use by many formatting words, qould it make sense to have the words have a `b` postfix? (akin to `/i` and similarly named words)
> 
> giving us `&b`, `|b` and so on.

There is no ambiguity; `.` as postfix means "print", `.` as prefix is only used by `cuda.ptx`, and no ambiguity would result to have it as bit operators, hence "bit"-or, "bit"-and etc.

erg:
In general we don't just add `ALIAS:` because `ALIAS:` is a half-baked feature and having multiple names for things increases complexity and decreases readability.

As for the name, `.` means print in Factor, so `.|` looks like some kind of print/pipe shell thing to my eyes. I like the suggestion on discord for `b|` and `b<<` etc over `.`

Really this is the kind of thing you should put in `.factor-boot-rc` or `.factor-rc` or some kind of `USE: golf` vocab.

nomennescio:
> In general we don't just add `ALIAS:` because `ALIAS:` is a half-baked feature and having multiple names for things increases complexity and decreases readability.

Just curious, why is `ALIAS:` a half-baked feature in your opinion?

[...]

PR #2706, opened 2022-11-04
introduce syntax.terse

twopir:
Addresses #2641; realized that `b>>` and `b<<` conflict with accessors, so changed to using backtick instead of b.

### DIFF
--- a/extra/syntax/terse/authors.txt
+++ b/extra/syntax/terse/authors.txt
@@ -1,0 +1,1 @@
+Alex Maestas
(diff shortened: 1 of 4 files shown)


erg:
This is better imo. Thanks!

mrjbq7:
USE: golf has fewer letters! 😂On Nov 8, 2022, at 9:57 AM, Doug Coleman ***@***.***> wrote:﻿
This is better imo. Thanks!

PR #2707, opened 2022-12-02
Add sequence>sequence, sequence>array combinators

gifti258:


jonenst:
Is there a better name? I think sequence>sequence is to confusing


Given the '<' and '>' in
input<sequence
output>sequence


sequence><sequence ?
sequence<>sequence ?
sequence>call>sequence?
input<sequence>output ?
input<sequence<output ?
input<sequence|output>sequence ?
input<output>sequence ?
input,output<>sequence ?

Something else ?

mrjbq7:
What does it do?Can you provide some tests or examples?On Dec 2, 2022, at 10:44 AM, Jon Harper ***@***.***> wrote:﻿
Is there a better name? I think sequence>sequence is to confusing
Given the '<' and '>' in
inputsequence
sequence><sequence ?
sequence<>sequence ?
sequence>call>sequence?
inputoutput ?
input<sequence<output ?
input<sequence|output>sequence ?
inputsequence ?
input,output<>sequence ?
Something else ?

PR #2714, opened 2022-12-08
match: Fix match-cond documentation

gifti258:
Match vars are self-evaluating, any mention of get can be confusing.


erg:
Thanks!

PR #2720, opened 2022-12-14
New parser for lint.vocabs

Capital-EX:
I've updated `lint.vocabs` to use a less fragile system for parsing. It also makes the parsing process more readable than the long regexp I had. Finally, it should prevent false positives with regards to the `multiline` vocab. Some words have been removed in the refactor process. Additionally, `find-unused.` will show a new import header for the user:

Screenshot_20221213_223840

### DIFF
--- a/extra/lint/vocabs/vocabs-docs.factor
+++ b/extra/lint/vocabs/vocabs-docs.factor
@@ -57,49 +57,6 @@ HELP: find-unused.
     }
 } ;
 
-HELP: get-imported-words
-{ $values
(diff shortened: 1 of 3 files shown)


mrjbq7:
I'm fine if you want to ``USE: vocabs.prettyprint.private`` instead of duplicating it

Capital-EX:
I removed the duplication and added `vocabs.prettyprint.private` to the imports

mrjbq7:
Thanks!!

PR #2722, opened 2022-12-16
continuations: Clarify ignore-error, ignore-error/f, ignore-errors stack effects

gifti258:
If these words are called with any other `quot` stack effects, they don't compile.


mrjbq7:
Thanks!

PR #2725, opened 2022-12-17
ascii, unicode: add vowels and vowel?

alex-ilin:
Could someone please verify that this change doesn't break anything?
Like bootstraps, lints, etc.

### DIFF
--- a/basis/english/english.factor
+++ b/basis/english/english.factor
@@ -106,6 +106,11 @@ CONSTANT: plural-to-singular $[ singular-to-plural [ swap ] assoc-map ]
 
 PRIVATE>
 
+CONSTANT: vowels "aeiou"
+
+: vowel? ( ch -- ? )
+    vowels member? ; inline
+
 : singularize ( word -- singular )
     dup >lower {
         { [ dup empty? ] [ ] }
@@ -114,9 +119,7 @@ PRIVATE>
         { [ dup "s" tail? not ] [ ] }
         {
             [
-                dup "ies" ?tail [
-                    last "aeiou" member? not
-                ] [ drop f ] if
+                dup "ies" ?tail [ last vowel? not ] [ drop f ] if
             ] [ 3 head* "y" append ]
         }
         { [ dup "es" tail? ] [ 2 head* ] }
@@ -130,9 +133,7 @@ PRIVATE>
         { [ singular-to-plural ?at ] [ ] }
         {
             [
-                dup "y" ?tail [
-                    last "aeiou" member? not
-                ] [ drop f ] if
+                dup "y" ?tail [ last vowel? not ] [ drop f ] if
             ] [ but-last "ies" append ]
         }
         {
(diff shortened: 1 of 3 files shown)


error256:
Vowels are sounds, assigning this quality to letters is already somewhat weird. Even if you do, it is locale-dependent and has nothing to do with `ascii` or `unicode` properties. It probably should be in `english`?

alex-ilin:
I'm only formalizing the already existing convention, which is somewhat English-specific.
I could not find the corresponding attribute in the Unicode DB.
It could be the first step into formalizing the concept and beginning its support in Factor.

catb0t:
Well, my 2 cents is also that the `"aeiou"`-based definition specifically should be in `english`, because of "äęïôú" and things, complicated by both ascii and unicode. Other than that, i think this won't break anything.

alex-ilin:
https://www.merriam-webster.com/dictionary/vowel
> a letter or other symbol representing a vowel

Capital-EX:
I strongly believe that this should be under an `english` vocabulary if it gets added.

[Vowels](https://en.wikipedia.org/wiki/Vowel) are a phonological concept and should not be in `unicode` nor `ascii`. Unicode and ASCII represent orthography not phonology. Phonology can vary massive by region and culture.

Generalizing across all languages would be an arduous task. Even just targeting English, you will have complex edge cases. The `u` in `qu` [labializes](https://en.wikipedia.org/wiki/Labialization) the `q`  and is not a vowel in this context. The y in joy and spy is a vowel. Even in languages that can fit into ASCII, you have things like `y` and `w` being used as vowels in a language like [Welsh](https://en.wikipedia.org/wiki/Welsh_orthography#Letter_names_and_sound_values).

Adding this feature to `unicode` or `ascii` would be inherently unsound and/or incomplete. Such a feature should be added to a vocabulary that clearly indicates it's limitations and it domain of usage.

If this feature cannot handle Japanese's Katakana and Hiragana, Arabic's vowel diacritics, Hindi's abugida, etc, etc, then it cannot go into `unicode` (as it *doesn't* handle unicode). If it cannot handle all languages that use Latin Basic and/or ASCII, then it should not go into `ascii`.

Capital-EX:
Also, this vocabulary will miss words like `cycle`, `cyclone`, `krypton`, `rhythm`, `rye`, and more.

94tx:
I agree with @Capital-EX in that, at least in its current state, does not fit into the `unicode` category. As they said, Unicode handles orthography, not phonology. Maybe if we made an IPA vocabulary... (joke)

For example, in Spanish, we have cases like the sequence `gue` not having the `u` pronounced (therefore the u doesn't phonetically count as a vowel), but `gua` having the `u` count as a vowel, `y` fully counting as a vowel in all instances (even though in some placements it's *kinda* pronounced as a consonant it has a consistent pronunciation that makes it a vowel), etc.

mrjbq7:
I don't have a strong opinion about orthography and phonology, but I would probably put it in ``english`` vocab for now. Especially as the ascii/unicode implementations are the same and feel duplicative. We can always move it later if it feels right to promote the concept above a specific language.

alex-ilin:
I've updated the PR.

PR #2726, opened 2022-12-18
shuffle: Fix 2reach

gifti258:
Currently `2reach` is the same as `2pick`. In the sequence `2dup`, `2over`, `2pick`, `2reach`, this should dip one level deeper.


mrjbq7:
Thanks!

PR #2727, opened 2022-12-21
inverse: Fix swapped + and *

gifti258:
`2 2 *` and `2 2 swap *` evaluate to `4`. `4 [ 2 * ] undo` evaluates to `2` while `4 [ 2 swap * ] undo` evaluates to `1/2`. Similarly with `+`, where `4 [ 2 swap + ]` evaluates to `-2` instead of `2`.


erg:
Bad bug. Thanks!

PR #2734, opened 2023-01-09
Add Generators

Kacarott:
Draft PR, as a place for discussions on what is good and what is bad. Happy to hear all feedback!


mrjbq7:
That's pretty cool! Do you want to have this merged or keep working through comments?

Btw, for anyone devs trying to merge, we should cherry-pick that dc19fece26be48514946d1eff3f91c272e18fe07  commit so the PR merge doesn't bring the merge commit with it, not sure why that's there.

Kacarott:
I'm happy for it to be merged already if you are happy with it (Keep in mind that this will be my first contribution to the Factor codebase, so a good review of the code might be warranted).

mrjbq7:
Have you thought about how you'd implement a ``yield from`` feature to yield from another generator?

```python
def generator2():
    for i in range(10):
        yield i

def generator3():
    for j in range(10, 20):
        yield j

def generator():
    yield from generator2()
    yield from generator3()
```

Kacarott:
Good point on `take-all`, I've changed it.

I also added a word to check if a generator is exhausted, however this doesn't check whether the generator will actually produce any more value, only if it has previously raised a `stop-generator` exception.

And finally, I added a `yield-from`, which was something I didn't even think about before but is definitely useful.

I've managed to somehow include yet another "merge" commit somehow, I am really not sure how this happens... But this can all be squashed to a single commit, right?

mrjbq7:
Merged, thanks!!

PR #2751, opened 2023-01-25
Print script auto-use message to stderr

Bubbler-4:


mrjbq7:
Want to use ``with-output>error`` instead?

Bubbler-4:
Something's gone wrong on my side, will try again.

PR #2755, opened 2023-01-28
Editors idea

erg:
Decouple loading editors and setting editors.

```factor
! in .factor-boot-rc or .factor-rc

    USE: editors
    SET-EDITOR: visual-studio-code
```


erg:
Fixed in 26ac0736e6bae9edc45479a46b59f17261daf2c6

Issue #2197, opened 2019-10-21
Git archive inconsistency: dangling tags 0.96 0.95 and 0.94


nomennescio:
When preparing a pull request for the (partially) restored source code history of Factor before the current first git commit. I encountered an inconsistency in the current git archive:
Tags 0.96, 0.95 and 0.94 point into dangling commits! That means those were never merged back into the main branch.  If you point at tag 0.93, you'll see in Gitk it shows:
Follows: 0.92
Precedes: 0.97

This is a serious issue, as it is quite nasty to try to reconstruct a tree with a corrected mergepoint, it will need rebase or filter-branch (the latter is cleaner).

On top of that, I have to redo a lot of my work on reconstructing Git paleo-history, and will not be able to create a pull-request until this is properly solved.

I'm willing to help to fix this.

erg:
I noticed for .92 and .93 that the commit ids were present in factorcode.org but not in github. Something is pretty messed up already. I also had github do a git-gc or something a few years ago because we basically had duplicate commits everywhere. At this point, I don't really have a good understanding of the repo.

One thing going on is that factorcode.org does a force push to github every 5 minutes. I guess I don't see the point of doing it this way anymore, and changing to github as the truth source would let us use the github ui to merge patches.

```bash
cat force-push-github
#!/bin/sh
export PATH=$PATH:/usr/local/bin/
cd /git/factor.git
git push -f git@github.com:factor/factor.git 'refs/tags/*' 'refs/heads/*'
```

nomennescio:
Duplicate commits is a sure sign of a partial rewrite of history, for instance caused by people not updating their local archive after a rewrite, and then pushing their old branch, which will pull in all old commits again.

All is not lost. Somewhere after tag 0.93 the branch need to be 'cut open', and the '0.96' tag needs to be reinserted into the main branch. The bad news is, is that it requires a rewrite of history, or maybe it can be avoid using git replace objects. This can also be used to solve the zeropaddedfilemode issue, for which the currently prescribed 'fix' is to do a fast-export - fast-import combo.

This would need to be coordinated, where everyone finishes their commits on branches, push it remotely, and after a rewrite of the archive, make a fresh clone again.

If you're considering this route, I would be interested in rewriting the first git commit to point to the older archive history, which I have already prepared.

This can be repaired, but some pain will be felt.

nomennescio:
[...]

PR #2759, opened 2023-02-04
Rename sorting words

gifti258:
This PR would rename words as follows:
* `natural-sort` → `sort` (shorter and makes more sense)
* `sort` → `sort-with`
* `sort-with` → `sort-by` (fits in better with other words that take a quotation that turns elements into keys)
* `sort-by` → `sort-with-spec`
* `sort-by-with` → `sort-with-spec-by`
* `sort-keys-by` → `sort-keys-with-spec`
* `sort-values-by` → `sort-values-with-spec`
* `inv-sort-with` → `inv-sort-by`

It would also add (*`inv-natural-sort`* →) `inv-sort`, (*`inv-sort`* →) `inv-sort-with`, `inv-sort-keys` and `inv-sort-values`.
The vocabulary sorting.slots would be renamed to 'sorting.specification'.

This PR is made to properly run tests and to get feedback.


mrjbq7:
Alternative idea: merge sorting.slots words into sorting?On Feb 3, 2023, at 7:36 PM, Giftpflanze ***@***.***> wrote:﻿This PR would rename words as follows:

natural-sort → sort (shorter and makes more sense)
sort → sort-with
sort-with → sort-by (fits in better with other words that take a quotation that turns elements into keys)
sort-by → sort-with-spec
sort-by-with → sort-with-spec-by
inv-sort-with → inv-sort-by

It would also add (inv-natural-sort →) inv-sort and (inv-sort →) inv-sort-with.
The vocabulary sorting.slots would be renamed to 'sorting.specification'.
This PR is made to properly run tests and to get feedback.



  https://github.com/factor/factor/pull/2759

Commit Summary

  674e8f6  sorting: Rename sort, sort-with, natural-sort


File Changes  (131 files)

  
    M
    basis/compiler/cfg/dominance/dominance.factor
    (2)
  
  
    M
    basis/compiler/cfg/linearization/linearization.factor
    (2)
  
  
    M
    basis/compiler/cfg/ssa/destruction/coalescing/coalescing.factor
    (2)
  
  
    M
    basis/compiler/cfg/ssa/interference/interference.factor
    (4)
  
  
    M
    basis/compiler/tests/redefine5.factor
    (2)
  
  
    M
    basis/compiler/tests/redefine8.factor
    (2)
  
  
    M
    basis/compiler/tests/redefine9.factor
    (2)
  
  
    M
    basis/compiler/tree/debugger/debugger.factor
    (2)
  
  
    M
    basis/compression/huffman/huffman.factor
    (2)
  
  
    M
    basis/compression/inflate/inflate.factor
    (2)
  
  
    M
    basis/cpu/arm/assembler/32/32.factor
    (1)
  
  
    M
    basis/cpu/arm/assembler/64/64.factor
    (1)
  
  
    M
    basis/cpu/arm/assembler/assembler.factor
    (1)
  
  
    M
    basis/cpu/arm/assembler/opcodes/opcodes.factor
    (1)
  
  
    M
    basis/editors/vim/generate-syntax/generate-syntax.factor
    (2)
  
  
    M
    basis/escape-strings/escape-strings.factor
    (2)
  
  
    M
    basis/fixups/fixups.factor
    (5)
  
  
    M
    basis/help/apropos/apropos.factor
    (2)
  
  
    M
    basis/help/search/search.factor
    (2)
  
  
    M
    basis/help/vocabs/vocabs.factor
    (4)
  
  
    M
    basis/http/server/static/static.factor
    (22)
  
  
    M
    basis/interval-maps/interval-maps.factor
    (2)
  
  
    M
    basis/logging/analysis/analysis.factor
    (2)
  
  
    M
    basis/math/primes/factors/factors.factor
    (2)
  
  
    M
    basis/math/primes/pollard-rho-brent/pollard-rho-brent.factor
    (2)
  
  
    M
    basis/math/statistics/statistics.factor
    (6)
  
  
    M
    basis/models/sort/sort.factor
    (2)
  
  
    M
    basis/regexp/minimize/minimize.factor
    (4)
  
  
    M
    basis/see/see.factor
    (6)
  
  
    M
    basis/sequences/parser/parser.factor
    (7)
  
  
    M
    basis/sorting/human/human-tests.factor
    (12)
  
  
    M
    basis/sorting/human/human.factor
    (4)
  
  
    D
    basis/sorting/slots/slots-docs.factor
    (48)
  
  
    D
    basis/sorting/slots/slots.factor
    (30)
  
  
    R
    basis/sorting/specification/authors.txt
    (0)
  
  
    A
    basis/sorting/specification/specification-docs.factor
    (48)
  
  
    R
    basis/sorting/specification/specification-tests.factor
    (30)
  
  
    A
    basis/sorting/specification/specification.factor
    (30)
  
  
    R
    basis/sorting/specification/summary.txt
    (0)
  
  
    M
    basis/sorting/title/title-tests.factor
    (2)
  
  
    M
    basis/splitting/monotonic/monotonic.factor
    (2)
  
  
    M
    basis/suffix-arrays/suffix-arrays.factor
    (2)
  
  
    M
    basis/tools/annotations/annotations.factor
    (2)
  
  
    M
    basis/tools/completion/completion.factor
    (2)
  
  
    M
    basis/tools/crossref/crossref.factor
    (2)
  
  
    M
    basis/tools/destructors/destructors.factor
    (4)
  
  
    M
    basis/tools/files/files.factor
    (6)
  
  
    M
    basis/tools/memory/memory.factor
    (2)
  
  
    M
    basis/tools/profiler/sampling/sampling.factor
    (2)
  
  
    M
    basis/tools/scaffold/scaffold.factor
    (4)
  
  
    M
    basis/ui/gadgets/menus/menus.factor
    (2)
  
  
    M
    basis/ui/tools/inspector/inspector.factor
    (2)
  
  
    M
    basis/unicode/unicode.factor
    (2)
  
  
    M
    basis/unix/linux/proc/proc.factor
    (6)
  
  
    M
    basis/vocabs/hierarchy/hierarchy.factor
    (4)
  
  
    M
    basis/vocabs/prettyprint/prettyprint.factor
    (3)
  
  
    M
    basis/xmode/catalog/catalog.factor
    (2)
  
  
    M
    basis/xmode/keyword-map/keyword-map.factor
    (2)
  
  
    M
    core/classes/algebra/algebra.factor
    (5)
  
  
    M
    core/sorting/sorting-docs.factor
    (25)
  
  
    M
    core/sorting/sorting-tests.factor
    (10)
  
  
    M
    core/sorting/sorting.factor
    (49)
  
  
    M
    core/source-files/errors/errors.factor
    (2)
  
  
    M
    core/vocabs/parser/parser.factor
    (2)
  
  
    M
    core/vocabs/vocabs.factor
    (2)
  
  
    M
    extra/alien/fortran/fortran.factor
    (2)
  
  
    M
    extra/anagrams/anagrams.factor
    (4)
  
  
    M
    extra/annotations/annotations-docs.factor
    (2)
  
  
    M
    extra/benchmark/interval-sets/interval-sets.factor
    (2)
  
  
    M
    extra/benchmark/sort/sort.factor
    (2)
  
  
    M
    extra/benchmark/splay/splay.factor
    (2)
  
  
    M
    extra/build-from-source/windows/windows.factor
    (2)
  
  
    M
    extra/c/lexer/lexer.factor
    (7)
  
  
    M
    extra/codebook/codebook.factor
    (4)
  
  
    M
    extra/color-table/color-table.factor
    (8)
  
  
    M
    extra/command-loop/command-loop.factor
    (2)
  
  
    M
    extra/compiler/cfg/gvn/testing/testing.factor
    (6)
  
  
    M
    extra/contributors/contributors.factor
    (2)
  
  
    M
    extra/did-you-mean/did-you-mean.factor
    (2)
  
  
    M
    extra/fuel/xref/xref.factor
    (6)
  
  
    M
    extra/gemini/server/server.factor
    (2)
  
  
    M
    extra/gml/core/core.factor
    (2)
  
  
    M
    extra/gopher/server/server.factor
    (2)
  
  
    M
    extra/gpu/render/render.factor
    (2)
  
  
    M
    extra/help/lint/coverage/coverage.factor
    (4)
  
  
    M
    extra/ifaddrs/ifaddrs.factor
    (2)
  
  
    M
    extra/images/atlas/atlas.factor
    (2)
  
  
    M
    extra/koszul/koszul.factor
    (8)
  
  
    M
    extra/lint/vocabs/vocabs.factor
    (4)
  
  
    M
    extra/machine-learning/decision-trees/decision-trees.factor
    (2)
  
  
    M
    extra/machine-learning/label-binarizer/label-binarizer.factor
    (2)
  
  
    M
    extra/machine-learning/label-encoder/label-encoder.factor
    (2)
  
  
    M
    extra/managed-server/chat/chat.factor
    (2)
  
  
    M
    extra/mason/test/test.factor
    (2)
  
  
    M
    extra/math/binpack/binpack.factor
    (2)
  
  
    M
    extra/math/extras/extras.factor
    (2)
  
  
    M
    extra/math/transforms/bwt/bwt.factor
    (2)
  
  
    M
    extra/multi-methods/multi-methods.factor
    (2)
  
  
    M
    extra/pair-methods/pair-methods.factor
    (2)
  
  
    M
    extra/papier/render/render.factor
    (2)
  
  
    M
    extra/project-euler/004/004.factor
    (2)
  
  
    M
    extra/project-euler/009/009.factor
    (2)
  
  
    M
    extra/project-euler/022/022.factor
    (2)
  
  
    M
    extra/project-euler/043/043.factor
    (2)
  
  
    M
    extra/project-euler/052/052.factor
    (2)
  
  
    M
    extra/project-euler/062/062.factor
    (2)
  
  
    M
    extra/project-euler/112/112.factor
    (2)
  
  
    M
    extra/project-euler/common/common.factor
    (2)
  
  
    M
    extra/punycode/punycode.factor
    (2)
  
  
    M
    extra/reports/noise/noise.factor
    (4)
  
  
    M
    extra/rosetta-code/anagrams-deranged/anagrams-deranged.factor
    (4)
  
  
    M
    extra/rosetta-code/count-the-coins/count-the-coins.factor
    (2)
  
  
    M
    extra/rosetta-code/knapsack/knapsack.factor
    (2)
  
  
    M
    extra/rosetta-code/top-rank/top-rank.factor
    (2)
  
  
    M
    extra/s3/s3.factor
    (6)
  
  
    M
    extra/snake-game/game/game.factor
    (2)
  
  
    M
    extra/sorting/bubble/bubble.factor
    (9)
  
  
    M
    extra/sorting/extras/extras.factor
    (4)
  
  
    M
    extra/sorting/quick/quick-tests.factor
    (8)
  
  
    M
    extra/sorting/quick/quick.factor
    (30)
  
  
    M
    extra/spelling/spelling.factor
    (2)
  
  
    M
    extra/tools/tree/tree.factor
    (2)
  
  
    M
    extra/units/units.factor
    (2)
  
  
    M
    extra/webapps/blogs/blogs.factor
    (2)
  
  
    M
    extra/webapps/mason/dashboard/dashboard.factor
    (2)
  
  
    M
    extra/webapps/pastebin/pastebin.factor
    (2)
  
  
    M
    extra/webapps/planet/planet.factor
    (6)
  
  
    M
    extra/webapps/wiki/wiki.factor
    (6)
  
  
    M
    extra/wordlet/wordlet.factor
    (2)
  
  
    M
    extra/wordtimer/wordtimer.factor
    (2)
  
  
    M
    extra/zoneinfo/zoneinfo.factor
    (2)
  


Patch Links:

  https://github.com/factor/factor/pull/2759.patch
  https://github.com/factor/factor/pull/2759.diff

razetime:
i think we should do one refactor at a time. The patch looks good to me.

nomennescio:
> This PR would rename words as follows:
> 
> * `natural-sort` → `sort` (shorter and makes more sense)
> * `sort` → `sort-with`
> * `sort-with` → `sort-by` (fits in better with other words that take a quotation that turns elements into keys)
> * `sort-by` → `sort-with-spec`
> * `sort-by-with` → `sort-with-spec-by`
> * `sort-keys-by` → `sort-keys-with-spec`
> * `sort-values-by` → `sort-values-with-spec`
> * `inv-sort-with` → `inv-sort-by`
> 
> It would also add (_`inv-natural-sort`_ →) `inv-sort`, (_`inv-sort`_ →) `inv-sort-with`, `inv-sort-keys` and `inv-sort-values`. The vocabulary sorting.slots would be renamed to 'sorting.specification'.
> 
> This PR is made to properly run tests and to get feedback.

Some suggestions are worth considering, but I don't like longer names for current short names. I actually like `natural-sort` as picking a "natural" sorting criterion, and `sort` being the more generic one.

mrjbq7:
I admit being really used to seeing ``natural-sort`` but i like this PR. 

Should we merge it? It will need new bootstrap images uploaded for the builders to pass, which I can take care of when you merge it.

@erg ?

gifti258:
There's still 5 test failures when doing `test-all` that I couldn't fix yet.

mrjbq7:
Probably the compiler tests, we could merge and fix the tests... or fix the tests and merge it.  Either is fine!

gifti258:
```
IN: scratchpad :test-failures

==== resource:basis/db/postgresql/errors/errors-tests.factor

resource:basis/db/postgresql/errors/errors-tests.factor

Generic word database<< does not define a method for the POSTPONE: f class.
Dispatching on object: f
(U) [ c-to-factor => ]
    c-to-factor
(U) [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(O) command-line-startup
(O) listener
(O) listener-loop
(O) listener-step
(U) [
        [ ~quotation~ dip swap ~quotation~ dip ] dip swap
        [ call get-datastack ] dip => swap [ set-datastack ] dip
    ]
(U) [ call => get-datastack ]
(O) test-vocabs
(O) test-vocab
(O) run-test-file
(O) test-postgresql
(O) M\ object database<<
(O) no-method
(O) M\ object throw
(U) [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]

==== resource:basis/db/postgresql/postgresql-tests.factor

resource:basis/db/postgresql/postgresql-tests.factor

Generic word database<< does not define a method for the POSTPONE: f class.
Dispatching on object: f
(U) [ c-to-factor => ]
    c-to-factor
(U) [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(O) command-line-startup
(O) listener
(O) listener-loop
(O) listener-step
(U) [
        [ ~quotation~ dip swap ~quotation~ dip ] dip swap
        [ call get-datastack ] dip => swap [ set-datastack ] dip
    ]
(U) [ call => get-datastack ]
(O) test-vocabs
(O) test-vocab
(O) run-test-file
(O) test-postgresql
(O) M\ object database<<
(O) no-method
(O) M\ object throw
(U) [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]

==== resource:basis/db/tuples/tuples-tests.factor

resource:basis/db/tuples/tuples-tests.factor

Generic word database<< does not define a method for the POSTPONE: f class.
Dispatching on object: f
(U) [ c-to-factor => ]
    c-to-factor
(U) [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(O) command-line-startup
(O) listener
(O) listener-loop
(O) listener-step
(U) [
        [ ~quotation~ dip swap ~quotation~ dip ] dip swap
        [ call get-datastack ] dip => swap [ set-datastack ] dip
    ]
(U) [ call => get-datastack ]
(O) test-vocabs
(O) test-vocab
(O) run-test-file
(U) [
        paste "PASTE" {
            ~array~
            ~array~
            ~array~
            ~array~
            ~array~
            ~array~
            ~array~
            ~array~
        } define-persistent [ test-foreign-key ] test-sqlite
        [ test-foreign-key-not-null ] test-sqlite
        [ test-cascade ] test-sqlite [ test-restrict ] test-sqlite
        [ test-foreign-key ] test-postgresql =>
        [ test-foreign-key-not-null ] test-postgresql
        [ test-cascade ] test-postgresql
        [ test-restrict ] test-postgresql
        [ db-assigned-person-schema test-tuples ] test-sqlite
        [ user-assigned-person-schema test-tuples ] test-sqlite
        [ user-assigned-person-schema test-repeated-insert ]
        test-sqlite [ test-bignum ] test-sqlite
        [ test-serialize ] test-sqlite
        [ test-intervals ] test-sqlite
        [ test-random-id ] test-sqlite
        [ db-assigned-person-schema test-tuples ] test-postgresql
        [ user-assigned-person-schema test-tuples ] test-postgresql
        [ user-assigned-person-schema test-repeated-insert ]
        test-postgresql [ test-bignum ] test-postgresql
        [ test-serialize ] test-postgresql
        [ test-intervals ] test-postgresql
        [ test-random-id ] test-postgresql [
            ~quotation~ ~quotation~ \ (must-fail-with) 517
            experiment
        ] test-sqlite [
            ~quotation~ ~quotation~ \ (must-fail-with) 522
            experiment
        ] test-postgresql suparclass f
        { ~array~ ~array~ } define-persistent subbclass "SUBCLASS"
        { ~array~ } define-persistent fubbclass "FUBCLASS"
        { } define-persistent [ test-db-inheritance ] test-sqlite
        [ test-db-inheritance ] test-postgresql
        string-encoding-test "STRING_ENCODING_TEST"
        { ~array~ ~array~ } define-persistent
        [ test-string-encoding ] test-sqlite
        [ test-string-encoding ] test-postgresql
        compound-foo "COMPOUND_FOO"
        { ~array~ ~array~ ~array~ } define-persistent
        [ test-compound-primary-key ] test-sqlite
        [ test-compound-primary-key ] test-postgresql timez "TIMEZ"
        { ~array~ ~array~ } define-persistent
        [ test-time-types ] test-sqlite
        [ test-time-types ] test-postgresql example "EXAMPLE"
        { ~array~ ~array~ } define-persistent
        [ test-blob-select ] test-sqlite ~10 more~
    ]
(O) test-postgresql
(O) M\ object database<<
(O) no-method
(O) M\ object throw
(U) [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]

==== resource:extra/memcached/memcached-tests.factor

resource:extra/memcached/memcached-tests.factor

Unix system call failed:

Connection refused (111)
(U) [ c-to-factor => ]
    c-to-factor
(U) [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(O) command-line-startup
(O) listener
(O) listener-loop
(O) listener-step
(U) [
        [ ~quotation~ dip swap ~quotation~ dip ] dip swap
        [ call get-datastack ] dip => swap [ set-datastack ] dip
    ]
(U) [ call => get-datastack ]
(O) test-vocabs
(O) test-vocab
(O) run-test-file
(U) [
        { t } [ ~quotation~ with-memcached length 0 > ]
        \ (unit-test) 28 experiment [ m/flush ] with-memcached =>
        [ "valuex" x m/set ] with-memcached
        { "valuex" } [ ~quotation~ with-memcached ] \ (unit-test) 33
        experiment [ m/flush ] with-memcached
        [ "valuex" x m/set "valuey" y m/set ] with-memcached
        { "valuex" } [ ~quotation~ with-memcached ] \ (unit-test) 38
        experiment
        { "valuey" } [ ~quotation~ with-memcached ] \ (unit-test) 39
        experiment [ m/flush ] with-memcached
        [ ~quotation~ with-memcached ] not-found?
        [ ~quotation~ with-memcached ] not-found?
        [ m/noop ] with-memcached [ m/flush ] with-memcached
        [ "valuex" x m/set ] with-memcached
        { "valuex" } [ ~quotation~ with-memcached ] \ (unit-test) 50
        experiment [ x m/delete ] with-memcached
        [ ~quotation~ with-memcached ] not-found?
        [ m/flush ] with-memcached
        [ ~quotation~ with-memcached ] not-found?
        [ ~quotation~ with-memcached ] not-found?
        [ "ex" x m/add ] with-memcached
        { "ex" } [ ~quotation~ with-memcached ] \ (unit-test) 59
        experiment [ "ex2" x m/replace ] with-memcached
        { "ex2" } [ ~quotation~ with-memcached ] \ (unit-test) 61
        experiment [ m/flush ] with-memcached
        { 0 } [ ~quotation~ with-memcached ] \ (unit-test) 65
        experiment
        { 1 } [ ~quotation~ with-memcached ] \ (unit-test) 66
        experiment
        { 212 } [ ~quotation~ with-memcached ] \ (unit-test) 67
        experiment
        { 8589934804 } [ ~quotation~ with-memcached ] \ (unit-test)
        68 experiment [ m/flush ] with-memcached
        [ "5" x m/set ] with-memcached
        { 4 } [ ~quotation~ with-memcached ] \ (unit-test) ~68 more~
    ]
(U) [ swap >n call => ndrop ]
    with-variables
(U) [
        [ over remote-address set <client> local-address set ] dip
        => with-stream
    ]
(U) [ over remote-address set <client> => local-address set ]
(O) <client>
(O) M\ array (client)
(O) M\ object (client)
(O) M\ object establish-connection
(O) (throw-errno)
(O) libc-error
(O) M\ object throw
(U) [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]

resource:extra/memcached/memcached-tests.factor: 28

Unit Test: { { t } [ [ m/version ] with-memcached length 0 > ] }

Unix system call failed:

Connection refused (111)
(U) [ c-to-factor => ]
    c-to-factor
(U) [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(O) command-line-startup
(O) listener
(O) listener-loop
(O) listener-step
(U) [
        [ ~quotation~ dip swap ~quotation~ dip ] dip swap
        [ call get-datastack ] dip => swap [ set-datastack ] dip
    ]
(U) [ call => get-datastack ]
(O) test-vocabs
(O) test-vocab
(O) run-test-file
(U) [
        { t } [ ~quotation~ with-memcached length 0 > ]
        \ (unit-test) 28 experiment => [ m/flush ] with-memcached
        [ "valuex" x m/set ] with-memcached
        { "valuex" } [ ~quotation~ with-memcached ] \ (unit-test) 33
        experiment [ m/flush ] with-memcached
        [ "valuex" x m/set "valuey" y m/set ] with-memcached
        { "valuex" } [ ~quotation~ with-memcached ] \ (unit-test) 38
        experiment
        { "valuey" } [ ~quotation~ with-memcached ] \ (unit-test) 39
        experiment [ m/flush ] with-memcached
        [ ~quotation~ with-memcached ] not-found?
        [ ~quotation~ with-memcached ] not-found?
        [ m/noop ] with-memcached [ m/flush ] with-memcached
        [ "valuex" x m/set ] with-memcached
        { "valuex" } [ ~quotation~ with-memcached ] \ (unit-test) 50
        experiment [ x m/delete ] with-memcached
        [ ~quotation~ with-memcached ] not-found?
        [ m/flush ] with-memcached
        [ ~quotation~ with-memcached ] not-found?
        [ ~quotation~ with-memcached ] not-found?
        [ "ex" x m/add ] with-memcached
        { "ex" } [ ~quotation~ with-memcached ] \ (unit-test) 59
        experiment [ "ex2" x m/replace ] with-memcached
        { "ex2" } [ ~quotation~ with-memcached ] \ (unit-test) 61
        experiment [ m/flush ] with-memcached
        { 0 } [ ~quotation~ with-memcached ] \ (unit-test) 65
        experiment
        { 1 } [ ~quotation~ with-memcached ] \ (unit-test) 66
        experiment
        { 212 } [ ~quotation~ with-memcached ] \ (unit-test) 67
        experiment
        { 8589934804 } [ ~quotation~ with-memcached ] \ (unit-test)
        68 experiment [ m/flush ] with-memcached
        [ "5" x m/set ] with-memcached
        { 4 } [ ~quotation~ with-memcached ] \ (unit-test) ~68 more~
    ]
(U) [
        2 load-locals -1 get-local <experiment> load-local
        silent-tests? get 0 get-local
        [ load-local 0 get-local experiment. 1 drop-locals ] curry
        unless -2 get-local execute => 0 get-local -1 get-local [
            2 load-locals -1 get-local 0 get-local ~quotation~ curry
            curry ~quotation~ if 2 drop-locals
        ] curry curry [ 2drop "Warning: test skipped!" print ] if
        3 drop-locals
    ]
    experiment
(O) (unit-test)
(U) [
        [ ~quotation~ dip swap ~quotation~ dip ] dip swap
        [ call get-datastack ] dip => swap [ set-datastack ] dip
    ]
(U) [ call => get-datastack ]
(U) [ [ m/version ] with-memcached => length 0 > ]
(U) [ swap >n call => ndrop ]
    with-variables
(U) [
        [ over remote-address set <client> local-address set ] dip
        => with-stream
    ]
(U) [ over remote-address set <client> => local-address set ]
(O) <client>
(O) M\ array (client)
(O) M\ object (client)
(O) M\ object establish-connection
(O) (throw-errno)
(O) libc-error
(O) M\ object throw
(U) [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]
```

mrjbq7:
Those aren't broken tests -- they just require configuring a postgresql database which our nightly builders do.

gifti258:
How can the tests pass on the master branch on the same machine though?

mrjbq7:
Install and start a postgresql database on a computer you can access, localhost or something nearby.

Create a user for the postgres database, something like:

```
sudo -u postgres psql -c "CREATE USER username WITH SUPERUSER PASSWORD 'password';"
```

Configure the postgres test database in your .factor-rc or something:

```factor
USING: db.postgresql db.tester ;
<postgresql-db>
    "hostname" >>host
    postgresql-test-db-name >>database
    "username" >>username
    "password" >>password
\ postgresql-db set-global
```

Running postgresq, redis, memcached, and other services to pass all the integration tests is a little bit of sysadmin and I'm perfectly happy if you want to make commits and then see if it passes the build machines that are already configured.

erg:
I like this PR! If `natural-sort` is so natural then why isn't it just called `sort`?

The rest of the names seem like an improvement as well. +1!!

nomennescio:
Hmm, ok, if that's the consensus. (Don't forget about the `sorting` child vocabularies either!) Can we discuss the `-with-spec-` variants? These seem a bit too long and non-descriptive for my taste. Can I make a suggestion here?

```
sort-by → sort-with-spec  → tuple-sort
sort-by-with → sort-with-spec-by  → tuple-sort-by
sort-keys-by → sort-keys-with-spec  → key-sort / sort-by-keys
sort-values-by → sort-values-with-spec  → value-sort / sort-by-values
```

mrjbq7:
USE: sorting.tuplesI kinda like that What’s an “unnatural sort”!On Feb 8, 2023, at 6:59 AM, nomennescio ***@***.***> wrote:﻿
Hmm, ok, if that's the consensus. (Don't forget about the sorting child vocabularies either!) Can we discuss the -with-spec- variants? These seem a bit too long and non-descriptive for my taste. Can I make a suggestion here?
sort-by → sort-with-spec  → tuple-sort
sort-by-with → sort-with-spec-by  → tuple-sort-by
sort-keys-by → sort-keys-with-spec  → key-sort / sort-by-keys
sort-values-by → sort-values-with-spec  → value-sort / sort-by-values

gifti258:
I named them `-with-spec` because they don't just sort tuples. Every ( x -- x ) word can be an accessor. We should extend the vocab to also support quotations.

nomennescio:
Let me take a little step back why I was not fond of all names; I think the general rule should be to be as terse and descriptive as possible, favoring shorter code for the most often use idiom. That's why I prefer `natural-sort` and `sort` over these proposed replacements `sort` and `sort-with`. I will clarify with some typical code:
```
: doit1 ( seq -- sorted ) natural-sort ;
: doit2 ( seq -- sorted ) [ <=> ] sort ;
```
Notice how for many common cases having a simple sort quotation is (almost) as short as using a default comparator.
With the proposed change this becomes
```
: doit1 ( seq -- sorted ) sort ;
: doit2 ( seq -- sorted ) [ <=> ] sort-with ;
```
which skews overall average code length.

Furthermore, another common naming scheme is to reserve general names for "generics", and specialized versions with longer names. This change is deviating from that. And the general `-with` prefix is *not* a common idiom for words that take a quotation, so it's not a required pattern.

As for the `-with-spec` variants; not only do they make most sorting words much longer, they are very non-descriptive -- what is a `spec`? -- and the general `with` does not add semantic value, therefore should be replaces/removed.

As a general remark; changing names of well-established interfaces with a legacy of ~15 years in the language is not something to be done lightly. If you would do that for people with existing codebases you're effectively breaking all of their code. Furthermore, don't underestimate the number of errors you're going to introduce because people have just learned these old interfaces. And you're now forcing the whole Factor community to learn new names. I think this PR was therefore published in too much haste. If it was my call, for that reason alone, I would revert it, until the discussion would have been settled.

As for your remark on that these don't just sort tuples: I think that's fair feedback, I will reply in another comment.

nomennescio:
As for the `-with-spec` I just noticed a complete rewrite of both the documentation (including *adding* a lot of descriptions, thanks for that!) and the vocabulary itself; it's now no longer `sorting.slots` but `sorting.specification`, where indeed it is now stated that this is about "sorting with multiple keys", instead of the former "sort slots" where only tuples where specified. I think such a more generic approach is indeed sensible, but is also a big overhaul. I still noticed that the `execute-accessor` "specifies" it's only operating on tuples, is that a mistake?
```
: execute-accessor ( obj1 obj2 word -- obj1' obj2' )
    [ ( tuple -- value ) execute-effect ] curry bi@ ;
```

Looking a bit more closely at your proposed naming scheme, you reserve the `-with` as idiom to specify a quotation, and the `-by` postfix to indicate transforming sequences elements into sortable "keys". In that sense `-by-spec` would be more fitting than `-with-spec`, and it's shorter too. 

However, now that the vocabulary has its intended effect to be "sort by multiple keys" (by applying a sequence of words to each sortable object to make subsequent "sortable keys" out of them), that does indeed shift away from 'tuple' to something more generic. I'm not too enthralled by name of the concept now called `sorting specifier` as `-spec` (because a `spec` or specifier is a much more generic concept than this "sorting specifier", which is therefore kind of a naming mismatch).

I think what best describes it conceptually is indeed "sorting by multiple keys in order, each key based on the same sortable object", which should be reflected in the name, which ideally should be both descriptive and short. As such I withdraw my previous proposal as describe here above, but I also think `sort-...-with-spec...` is also not the ideal candidate.

At this point I'm starting to think that a better way would be not to resort getting into a "jungle of names" for each detailed difference, because such a pattern typically reveals a concept that is struggling to be made explicit. I'm talking here about a real generic sort word that takes a real type and specializes on it. In this case such a type would describe different sorts of "sorting specifiers". Another "smell" here, is that the implementation of these sorting words is using several dynamic calls to `execute-effect`, which is impacting efficiency of such calls, as well as potential for generating exceptions. Using generics can make that "static".

( You can look at how e.g. Java handles this by building custom comparators, and operating on them:
```
Comparator<Duck> cmp = Comparator
    .comparing((Duck duck)-> duck.weight)
    .thenComparing((Duck duck)-> duck.age)
    .thenComparing(duck-> duck.name);
```
notice how a type is used here, where a "sorting spec" is just a specially structured sequence -- which in general is another "smell"! -- )

I think this would warrant another restructuring, with a focus on "sorting by custom comparator objects", which is really better than "sorting by multiple keys". For now I will stick to the renaming only, but I'd love to hear your feedback on this.

Bringing it back to the essentials (sorting on multiple keys), why not create a name out of ``sort-...-with-spec...` and base variants on that.

Some suggestions:
```
multiple-keys-sort
multiple-sort
keys-sort
comparator-sort
comp-sort
msort
ksort
csort
```

Using e.g. `msort` would give:
```
natural-sort → sort
sort → sort-with
sort-with → sort-by
sort-by → msort
sort-by-with → msort-by
sort-keys-by → msort-keys
sort-values-by → msort-values
``
(and all the `inv-` variants and child vocabs similarly)

nomennescio:
One more note: notice how the original (!) `sort` is already this generic by taking *any* comparator quotation.
What is lacking to facilitate doing sorting on multiple keys is actually a convenient way to build custom comparators for structures that require it. (Notice how `sorting.slots`/`sorting.specifiers` can easily (statically!) transform "sorting specifiers" into (composed) comparators!)

mrjbq7:
I agree on the sort-specs, I don't like how it does stack inference, and I don't like the duplication of the underlying sorting api.

As an aside, I find USE: sorting.specification to be too verbose, I'd prefer a shorter ``compare-with`` word that's put into the sorting vocab and used with the ``sort-by`` or ``sort-with`` words.

gifti258:
> Let me take a little step back why I was not fond of all names; I think the general rule should be to be as terse and descriptive as possible, favoring shorter code for the most often used idiom.

I agree. It seems though that `natural-sort` is a lot more common than `sort`, and when I grep the Factor code base there seem to be several times more occurences of `natural-sort` than `sort`. Also, for such a common sorting as `natural-sort`, the name is comically long. It is a wart that I want to see gone. `sort` fits much better.

> Furthermore, another common naming scheme is to reserve general names for "generics", and specialized versions with longer names. This change is deviating from that. And the general `-with` prefix is _not_ a common idiom for words that take a quotation, so it's not a required pattern.

No, but it is better than a more descriptive name like `sort-with-comparator`. This is a compromise, as `sort` is still an important word as the base of all other sorting words. I don't think `sort` should be reserved for a word that is not that commonly used. We could also name it `(sort)` but it maybe would be a little weird to expose a parenthesized word as a public interface.

> As a general remark; changing names of well-established interfaces with a legacy of ~15 years in the language is not something to be done lightly.

At the same time, we shouldn't let that held us back to make changes.

> By your (forced!) push, for people with existing codebases, you're effectively breaking all of their code. This is happening for real for e.g. the Codewars community; a new 0.99 version will effectively invalidate all existing solutions.

Is there no mechanism in place to still use older versions? How do they handle Python 2 vs. 3?

Between 0.98 and 0.99 `math.matrices` changed a lot, too. This alone will invalidate several of my solutions if it must be that any new version invalidates all older ones.

> Furthermore, don't underestimate the number of errors you're going to introduce because people have just learned these old interfaces.

There have been other breaking changes, and I don't like to have to change my own code outside of Factor. But we also have improved some word names.

If more people come forward and think the new names are garbage, I am happy to revert this PR and fork Factor to make it more to my liking (maybe some of it will be merged back).

> And you're now forcing the whole Factor community to learn new names. I think this PR was therefore published in too much haste. If it was my call, for that reason alone, I would revert it, until the discussion would have been settled.

I think that would be a bit harsh. The changes had support.

> I still noticed that the `execute-accessor` "specifies" it's only operating on tuples, is that a mistake?
> 
> ```
> : execute-accessor ( obj1 obj2 word -- obj1' obj2' )
>     [ ( tuple -- value ) execute-effect ] curry bi@ ;
> ```

I didn't see it as necessary to change it but it should probably be `obj` instead of `tuple`.
 
> Looking a bit more closely at your proposed naming scheme, you reserve the `-with` as idiom to specify a quotation, and the `-by` postfix to indicate transforming sequences elements into sortable "keys". In that sense `-by-spec` would be more fitting than `-with-spec`, and it's shorter too.

My reasoning is `-by` for keyer quotations and `-with` for everything else (comparator quotations, sorting specifications).

> However, now that the vocabulary has its intended effect to be "sort by multiple keys" (by applying a sequence of words to each sortable object to make subsequent "sortable keys" out of them), that does indeed shift away from 'tuple' to something more generic. I'm not too enthralled by name of the concept now called `sorting specifier` as `-spec` (because a `spec` or specifier is a much more generic concept than this "sorting specifier", which is therefore kind of a naming mismatch).

It was called sorting specification before and is called sorting specification after. The vocab name could also be `sorting.specs` if y'all like that better.

> I think what best describes it conceptually is indeed "sorting by multiple keys in order, each key based on the same sortable object", which should be reflected in the name, which ideally should be both descriptive and short. As such I withdraw my previous proposal as describe here above, but I also think `sort-...-with-spec...` is also not the ideal candidate.

Regarding sorting by multiple keys, the concept I think of is sorting tables by multiple columns. `table-sort` would be an appropriate name, what do you think of that?

> At this point I'm starting to think that a better way would be not to resort getting into a "jungle of names" for each detailed difference, because such a pattern typically reveals a concept that is struggling to be made explicit. I'm talking here about a real generic sort word that takes a real type and specializes on it. In this case such a type would describe different sorts of "sorting specifiers". Another "smell" here, is that the implementation of these sorting words is using several dynamic calls to `execute-effect`, which is impacting efficiency of such calls, as well as potential for generating exceptions. Using generics can make that "static".
> 
> ( You can look at how e.g. Java handles this by building custom comparators, and operating on them:
> 
> ```
> Comparator<Duck> cmp = Comparator
>     .comparing((Duck duck)-> duck.weight)
>     .thenComparing((Duck duck)-> duck.age)
>     .thenComparing(duck-> duck.name);
> ```
> 
> notice how a `comparator` type is used here to handle genericity, whereas Factor's "sorting spec" is just a specially structured sequence. "Specially structured sequences" are typically also a sign of a kludge to handle a more generic concepts. And concepts warrant their own classes. )
> 
> I think this would warrant another restructuring, with a focus on "sorting by custom comparator objects", which is really better than "sorting by multiple keys". For now I will stick to the renaming only, but I'd love to hear your feedback on this.

I would not like OO to be the only way to do this. I like the idea of combining several comparators to one. I would then think of a macro and it would look a lot like the existing `sorting.slots` implementation.

> Bringing it back to the essentials (sorting on multiple keys), why not create a name out of ``sort-...-with-spec...` and base variants on that.
> 
> Some suggestions:
> ```
> msort
> ksort
> csort
> ```

`msort`, `ksort` and `csort` are bad names.

nomennescio:
Thanks for your reply @gifti258.

As for your merge; I just think it went too quick (and I am against force pushing, see https://github.com/factor/factor/issues/2197) In general, as the Factor community gets more contributors, we're currently missing a more structured approach on how to agree on changes. I therefore asked @mrjbq7 if we can make a positive change towards such an approach. That will bring clarity on how to discuss and agree on pros and cons of proposals and deciding how and when to proceed with approving and merging. I think that's especially important for "breaking changes". And yes, breaking changes might be sometimes necessary, but should be avoided if possible, and that requires careful planning. I'm also trying to contribute to discussing the usefulness of this change with my remarks (and I'm glad to see we can agree on many points).

I recognize that part of the proposal I made is directed towards changing how `sorting.specifications` work, not just about picking names, and hence not strictly part of this rename action. I still think the `sort-..-with-spec..` names can be improved upon, and have no problem labeling some of my suggestions as "bad names", because I know it's hard to come up with good names.

But given in the light of my proposal, sorting is not the key issue, but defining a custom order on objects, based on different criteria in a predefined order. Once you have build a custom "comparator", sorting with it is trivial, and wouldn't *need* any special names, not even a special sort word. Therefore I think neither `table-sort` nor `multiple-sort` should be needed in the first place, once such a design is in place. I still think reserving the `sort` word to be able to also handle such comparators would be best, as to mimic a "generic" word as much as possible. (If `sort` means `[ <=> ] sort-with`, than I would prefer a design where `<=>` can handle such comparators if possible.)

As for your remark on OO, I can't exactly follow what you're trying to say; the whole of Factor is build around OO, the use of generics is pervasive in every detail of the language. The example I gave from Java is just to show the *concept* of a comparator, which is build piecewise, to exactly arrive at a single comparator that defines an order on objects by different criteria that are tried in a certain order (*"compare for length, if equal compare on weight, if equal compare on color"*) similar to how currently Factor's sorting specifications are kind of defining an order, but it now requires a special sort word. (Note that is is very close to how `compose` works, which also is supported by the compiler, hence macros might not even be needed. But such is implementation detail pending a detailed design).

gifti258:
> As for your merge; I just think it went too quick (and I am against force pushing, see #2197).

I don't see where I force-pushed anything. I force-pushed to the PR branch twice to replace the commit. And then I rebased that commit onto master. Did I bring in any objects that don't belong or mess up the repository otherwise?

> As for your remark on OO, I can't exactly follow what you're trying to say.

I had the impression that you needed to use tuples to make it work. But I think composing comparators piecewise is a good idea.

nomennescio:
I can't be checking every commit to see if the archive got messed up, it's your responsibility to not mess up the repository. But Github labeled the commit a force push. And force pushing should in general *never* be done, because it can bring in lingering objects from your local repo which were removed from the master repo. This has actually happened in the past (see #2197), and now takes a lot of work to fix -- which I volunteered to do -- and prevent from reoccuring. It's not rocket science, as even the Git documentation warns against force pushing, and it never needed in 99.99% of cases. It's just bad practice.

Kacarott:
I'm a bit late, but I really like this PR. Having `sort` be exactly that: sort, feels so much more *natural* than it was previously.

PR #2770, opened 2023-03-09
help: switch IRC links to Discord

bpollack:
The IRC server, while it still exists, has extremely low traffic, and isn't reliably checked by the developers.  This switches those references to instead be our Discord, which is quite active.


razetime:
this looks good, i think it's also useful to linkt eh gitter/matrix as a less active alternative since some people are reluctant to use a service like discord.

bpollack:
Good catch, @mrjbq7. Fixed. You okay if I merge this now?

PR #2775, opened 2023-03-17
Add example for using mutable stack effect

tgunr:


erg:
@tgunr You gotta up your git game!

PR #2776, opened 2023-03-31
Added LIKE" for creating 'WHERE column LIKE' db queries

tgunr:
This change permits the use of 'LIKE' statements. The current select-tuples currently uses column = 'text' if test is text. I added a LIKE" which can be used as 

`db [ T{ table  { column LIKE" this" } } select-tuples ] with-db`

This also works with `count-tuples` and the other db.tuple words which use the same underlying code.

### DIFF
--- a/basis/db/queries/queries.factor
+++ b/basis/db/queries/queries.factor
@@ -2,12 +2,21 @@
 ! See https://factorcode.org/license.txt for BSD license.
 USING: accessors arrays byte-arrays classes classes.tuple
 combinators continuations db db.errors db.tuples
-db.tuples.private db.types destructors kernel make math
-math.bitwise math.intervals math.parser namespaces nmake random
-sequences strings ;
+db.tuples.private db.types destructors kernel lexer make math
+math.bitwise math.intervals math.parser namespaces nmake
+peg.search random sequences strings strings.parser ;
 IN: db.queries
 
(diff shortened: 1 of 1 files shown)


mrjbq7:
Why’d you close this?We should get it working, I noticed the old persistency vocab has a similar syntax that creates a pattern object. On Sep 9, 2023, at 4:18 AM, Dave Carlton ***@***.***> wrote:﻿
Closed #2776.

tgunr:
Didn't know I closed it! How to reopen?

mrjbq7:
Maybe auto-closed when you "@[tgunr](https://github.com/tgunr) tgunr deleted the db.tuples  branch 3 hours ago"

mrjbq7:
The ``persistency`` vocab has this:

```factor
TUPLE: pattern value ;
C: <pattern> pattern
SYNTAX: %" parse-string <pattern> suffix! ;
M: pattern where value>> over column-name>> 0% " LIKE " 0% bind# ;
```

Perhaps we move these lines to ``db.queries``?

tgunr:
In the persistency vocab, you would have to create a pattern for each column you needed LIKE?

mrjbq7:
As far as I can tell the persistency one is identical in behavior (except for ``lexer get skip-blank``) to the one in this PR?

PR #2778, opened 2023-04-01
Fix `read1` and `readln` in the ui.listener

Capital-EX:
These changes should fix `read1` and `readln` in the UI Listener. I've added specialized `interactor-read*-finish` and `interactor-read*` words to handle the actual behavior or `read1` and `readln`. This should leave any word that depends on existing implementation of `interactor-read` unchanged.

Marking this as draft as I have to make changes to `read` as well. Currently, `n read n read ...` exhibits the same bug. Should be able to apply the same fix, but I need to take a break for today. Feel free to offer feedback.

### DIFF
--- a/basis/ui/tools/listener/listener.factor
+++ b/basis/ui/tools/listener/listener.factor
@@ -4,17 +4,18 @@ USING: accessors arrays assocs calendar combinators
 combinators.short-circuit concurrency.flags
 concurrency.mailboxes continuations destructors documents
 documents.elements fonts hashtables help help.markup help.tips
-io io.styles kernel lexer listener literals math math.vectors
-models models.arrow models.delay namespaces parser prettyprint
-sequences source-files.errors splitting strings system threads
-ui ui.commands ui.gadgets ui.gadgets.editors ui.gadgets.glass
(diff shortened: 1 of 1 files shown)


Capital-EX:
I've now fixed `read` to allow multiple line inputs. Gonna work on `read-until` later.

Capital-EX:
Progress:


- [x] readln
- [x] read1
- [x] read
- [ ] read-until
- [ ] read-quot

Capital-EX:
Well, that required a lot more re-writing than I expected. But, I think it should all be done. The only part I'm not confident about is the change to `read-quot`. 

It's behavior in the CLI Listener is to read each line as one quote. However, it's behavior in the UI Listener is to read *all* lines then evaluate them. So, I had to implement a means to check which mode `read-quot` should be in.

olus2000:
This PR seems to completely break `read-contents` on the `interactor`, which in all fairness makes it somewhat consistent with the cli listener. It doesn't react to the enter key alone (newlines can still be inserted with shift+enter), and there doesn't seem to be a way to end the input.

As for `readln`: the input is removed from display after enter is pressed if more input is requested.

PR #2795, opened 2023-06-22
io.crlf: add use-crlf to support output streams with CRLF for EOL

alex-ilin:
How do you like the idea of this stream modifier?
Not sure if that's in the spirit of things, to call such a state changer function that updates the context.
I'm open to suggestions about naming of the things in this PR.

The main thing to look at here is the word `use-crlf`, which modifies the current `output-stream` by wrapping it in a class with a custom `stream-nl` method. The latter outputs CRLF.

### DIFF
--- a/basis/io/crlf/authors.txt
+++ b/basis/io/crlf/authors.txt
@@ -1,2 +1,3 @@
 Daniel Ehrenberg
 Slava Pestov
+Alexander Ilin
--- a/basis/io/crlf/crlf-docs.factor
+++ b/basis/io/crlf/crlf-docs.factor
@@ -1,6 +1,6 @@
-! Copyright (C) 2009 Daniel Ehrenberg
+! Copyright (C) 2009, 2023 Daniel Ehrenberg, Alexander Ilin
 ! See https://factorcode.org/license.txt for BSD license.
-USING: help.syntax help.markup sequences ;
+USING: help.syntax help.markup io sequences ;
 IN: io.crlf
 
 HELP: crlf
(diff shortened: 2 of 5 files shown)


mrjbq7:
It’s an interesting idea, not sure io.crlf can depend on delegate. I personally dislike CRLF and think we should use LF everywhere. On Jun 22, 2023, at 10:04 AM, Alexander Ilin ***@***.***> wrote:﻿How do you like the idea of this stream modifier?
Not sure if that's in the spirit of things, to call such a state changer function that updates the context.
I'm open to suggestions about naming of the things in this PR.
The main thing to look at here is the word use-crlf, which modifies the current output-stream by wrapping it in a class with a custom stream-nl method. The latter outputs CRLF.



  https://github.com/factor/factor/pull/2795

Commit Summary

  bedc66e  io.crlf: add use-crlf to support output streams with CRLF for EOL


File Changes  (5 files)

  
    M
    basis/io/crlf/authors.txt
    (1)
  
  
    M
    basis/io/crlf/crlf-docs.factor
    (7)
  
  
    M
    basis/io/crlf/crlf-tests.factor
    (4)
  
  
    M
    basis/io/crlf/crlf.factor
    (22)
  
  
    M
    basis/io/crlf/summary.txt
    (2)
  


Patch Links:

  https://github.com/factor/factor/pull/2795.patch
  https://github.com/factor/factor/pull/2795.diff

alex-ilin:
> It’s an interesting idea, not sure io.crlf can depend on delegate.

Both are in `basic`, so I would assume so.

>  I personally dislike CRLF and think we should use LF everywhere.

I hear you, but a) `io.crlf` was arleady there, and b) I have to produce text files on Windows sometimes.

mrjbq7:
If it Bootstraps the dependency is okay.

mrjbq7:
This is cool! Merged!

Tweaked the interface slightly in 8c3dde8923ac7bcf46257fb71e7fecf5fc77fcd2.

Thanks!!

PR #2797, opened 2023-07-03
Use https rather than git protocol in download instructions

khinsen:
Only project members can clone via the git protocol.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ a boot image stored on factorcode.org.
 
 To check out Factor:
 
-* `git clone git://github.com/factor/factor.git`
+* git clone https://github.com/factor/factor.git
 * `cd factor`
 
 To build the latest complete Factor system from git, either use the


erg:
I believe it's anyone with a github ssh key, not just project members.

khinsen:
I have a GitHub SSH key, but it works only for repositories I have write access on. I suppose the git protocol can't provide read-only access, but that's just my hypothesis.

Anyway, I discovered this issue by trying to follow the instructions and failing!

bpollack:
This issue is conflating two different things:
 1. Anyone can clone anything over SSH. I just quickly verified that by cloning github.com/opencv/opencv over SSH (and you can trivially verify I not only am not a core contributor, but have never even contributed a patch).
 2. That said, `git:` is not the SSH protocol; it's the old unencrypted cloning protocol. That got disabled a couple years ago (see https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git). It won't work on _any_ repository, so it was a good catch to fix the docs.
Thanks for the patch, @khinsen! It's the right fix for the `README.md`. But definitely feel free to work with an SSH clone if you prefer.

khinsen:
Thanks @bpollack, I learned something about git today!

PR #2808, opened 2023-07-30
Brain-flak vocabulary

olus2000:
During a recent competition I implemented a compiler for the esoteric language [brain-flak](https://esolangs.org/wiki/Brain-Flak) as a Factor vocabulary. Detailed documentation can be found in the vocabulary but here's a quick summary:

Normal word `compile-brain-flak` and syntax word `b-f"` are responsible for compiling brain-flak snippets to Factor. These snippets operate on a `brain-flak` state object which can be constructed from a sequence of numbers. They can also be treated as function on sequences by calling them with `with-brain-flak` wrapper word.

Here's example usage of the vocab:
```factor
: fibonacci-sequence ( n -- array )
    1array [ b-f"<>((()))<>{({}[()])<>({}<>)<>(({})<>({}<>))<>}
                 <>{}{}" ] with-brain-flak ;
10 fibonacci-sequence
{ 1 1 2 3 5 8 13 21 34 55 }
```

### DIFF
--- a/extra/brain-flak/authors.txt
+++ b/extra/brain-flak/authors.txt
@@ -1,0 +1,1 @@
+Aleksander Sabak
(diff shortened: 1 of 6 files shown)


mrjbq7:
Awesome!! Thanks!!

PR #2809, opened 2023-08-05
ui.gadgets.tables: adds ctrl-p/n for row selection

inivekin:
this applies the bindings of the arrow keys to ctrl-n/p as well.
I suggest this as I am used to using ctrl-n/p for choosing between completions for vim/kakoune and this allows my muscle memory to not change.


mrjbq7:
Seems cool, thanks!!

PR #2811, opened 2023-08-09
arrays: add ?pair

alex-ilin:
What do you think, should we add this?
Kinda belongs to the arrays vocab, but not that useful to be in the core, so I'm not sure about the placement.
This word is useful if we want to convert the result of `assoc-find` back to the pair form.

### DIFF
--- a/core/arrays/arrays-docs.factor
+++ b/core/arrays/arrays-docs.factor
@@ -37,6 +37,8 @@ $nl
 { $subsections resize-array }
 "The class of two-element arrays:"
 { $subsections pair }
+"Conditionally creating arrays:"
+{ $subsections ?pair }
 "Arrays can be accessed without bounds checks in a pointer unsafe way."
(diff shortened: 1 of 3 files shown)


mrjbq7:
What's a use case for this?

alex-ilin:
Something like this: `: find-pair ( assoc value -- pair/f ) '[ nip _ = ] assoc-find ?pair ;`

mrjbq7:
In most of the assoc interface we don't treat ``{ key value }`` as a pair, though. With the exception maybe of ``>alist``.

I'm so-so on this, but if we have any examples where it would be useful in core/basis/extra that would be interesting.

And maybe a definition that does no work for the ``f`` case?

```: ?pair ( key value pair? -- pair/f ) [ 2drop f ] [ 2array ] if ;```

alex-ilin:
Nah, forget it. It's not that useful at all.

PR #2812, opened 2023-08-11
Allow dynamic linker environment variable usage

gamecubate:
re. https://discord.com/channels/780615045771821076/786055699271909428/1139352905078558822


mrjbq7:
Let’s see if that works! Thanks!

PR #2833, opened 2023-08-25
words: add unintern-word definition

Capital-EX:
This adds a corresponding class for `<uninterned-word>` via a `PREDICATE:`


mrjbq7:
Merged, thanks!

Issue #2832, opened 2023-08-25
`tools.crossref` Does Not Reliably Cross Reference.


Capital-EX:
Currently the `quot-uses` word in `tools.crossref` does support all literals in Factor's code base that can also hold quotations. Additionally, it doesn't support uninterned words.

I propose adding support for the follow:

- `L{ ... }` (list)
- `H{ ... }` (hashtable)
- `V{ ... }` (vector)
- `NH{ ... }` (number-hashtable)
- `VL{ ... }` (vlist)
- `VA{ ... }` (valist)
- `SH{ ... }` (sequence-hashtable)
- `PH{ ... }` (persistent-hash)
- `PV{ ... }` (persistent-vector)
- `LH{ ... }` (linked-hash)
- `DL{ ... }` (dlist)
- `HS{ ... }` (hash-set)
- `IH{ ... }` (identity-hashtable)
- `IHS{ ... }` (identity-hash-set)
- `SHS{ ... }` (sequence-hash-set)
- `AVL{ ... }` (avl)
- `TREE{ ... }` (tree)
- `SPLAY{ ... }` (splay)

Additionally `words` should add the following definition:

```factor
PREDICATE: uninterned-word < word
    vocabulary>> not ;
```

Capital-EX:
Upon deeper searching on how to eliminate unused deps using static analysis. I have found that `tools.crossref` has much deeper problems. Since it relies on traversing the body of words, any word generated using parsing words or compilation units, will not be handled correctly.

The smallest example is what `uses` produces for a short piece of `EBNF:` code:

```factor
USING: ascii multiline peg.ebnf sequences ;
IN: foobar

EBNF: number? [=[ 
    Number = .* ?[ [ digit? ] all? ]?
    Ascii = .* ?[ [ ascii? ] all? ]?
    rule = Number | Ascii
]=]
```

```factor
IN: scratchpad "foobar" vocab-uses.
namespaces
peg
peg.private
```

None over these vocabs are direct dependencies of the code I have written. If I was to replace my current `USING:` with these vocabs, I would receive a compilation error. Additionally, expanding to include `uninterned-words` produces more indirect deps:

```factor
IN: scratchpad "foobar" vocab-uses.
ascii
combinators.short-circuit
kernel
namespaces
peg
peg.private
sequences
syntax
```

Issue #2723, opened 2022-12-16
Improve `lint.vocabs` handling of `EBNF:`


gifti258:
Currently with the rewrite of lint.vocabs, vocabs only needed within the `[=[ … ]=]` block of `EBNF:` forms, along with `multiline` for the `[=[ ]=]` itself, are overlooked. Cases needing special consideration within the EBNF syntax block include `[[ … ]]`, `?[ … ]?` and `<foreign …>` rules.

gifti258:
@Capital-EX

Capital-EX:
I'll be looking into it!

Capital-EX:
When #2836 gets introduced I'll re-do `lint.vocabs` to use `tools.crossref`. My changes there introduce support for [uninterned words](https://docs.factorcode.org/content/article-uninterned-words.html). Fulling parsing arbitrary Factor code with support for parsing words is far to hard without re-implementing the parser itself.

`tools.crossref` instead allow me to operate on the generated code. This will make the tool make far more robust compared to parsing.

PR #2836, opened 2023-08-27
tools.crossref: add support for general sequences, assocs, and list.

Capital-EX:
Addresses #2832 (and later #2723) by adding support for all `sequences`, `assocs`, and `list`. I limited `ingored-sequences` to definitions in `core` or `basis` to avoid introducing dependencies into `extras`.

### DIFF
--- a/basis/tools/crossref/crossref.factor
+++ b/basis/tools/crossref/crossref.factor
@@ -18,7 +18,7 @@ SYMBOL: visited
 
 GENERIC#: quot-uses 1 ( obj set -- )
 
-UNION: ingored-sequences
+UNION: ignored-sequences
     iota
     range
     bits
@@ -29,7 +29,7 @@ UNION: ingored-sequences
     specialized-array
     bit-array ;
 
-M: ingored-sequences quot-uses 2drop ;
+M: ignored-sequences quot-uses 2drop ;
 M: object quot-uses 2drop ;
 
 M: word quot-uses over crossref? [ adjoin ] [ 2drop ] if ;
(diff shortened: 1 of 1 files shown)


JobLeonard:
Is `ingored` a typo of `ignored` or intentional?

Capital-EX:
That's a typo

Capital-EX:
Thinking about closing this PR. The implementation in the original cannot find direct dependencies if manipulated by syntax words. It only counts a direct dependency if it can be found in the body of a word. This is not correct as the following will report the wrong deps:

```factor
USING: ascii multiline peg.ebnf sequences ;
IN: foobar

EBNF: number? [=[ 
    Number = .* ?[ [ digit? ] all? ]?
    Ascii = .* ?[ [ ascii? ] all? ]?
    rule = Number | Ascii
]=]
```

```factor
IN: scratchpad "foobar" vocab-uses.
namespaces
peg
peg.private
```

None of the vocabularies required to parse foobar are present. I however was working from the incorrect assumption this vocabulary handled dependencies correctly. Thus, my changes will find required vocabularies, but includes many false positive:

```factor
IN: scratchpad "foobar" vocab-uses.
ascii
combinators.short-circuit
kernel
namespaces
peg
peg.private
sequences
syntax
```

Capital-EX:
Just going to go head and scrap these changes. Searching for vocabs via word definitions is not viable in Factor. It's extensibility renders this approach to static analysis unreliable. And, I am just going to leave this vocab as is.

PR #2843, opened 2023-09-01
io.files: using if-file-exists in various places

mrjbq7:
Experimenting with the new ``if-file-exists``, ``when-file-exists``, and ``unless-file-exists`` words.

They technically have a race condition between checking file existence and running the quotations, but the code below should be identical to how it worked before...

What do you think?


mrjbq7:
Closing, merged a similar commit in 589f0e83c82107ef260e60a61d3ff2b5a9c7add3.

PR #2850, opened 2023-09-06
Add set methods to ranges

Kacarott:
Adds specific instance methods to various `sets` generics. This is mostly for better performance, but notable is `intersect` which now returns a new intersection range rather than being converted to vectors first. `intersect` was based off the [Julia implementation](https://github.com/JuliaLang/julia/blob/8e14322b5aa344639dd86bf9eabb84afe831fcba/base/range.jl#L1185) of the same.

Code review, improvements, comments, ideas, problems, all welcome :)

### DIFF
--- a/core/ranges/authors.txt
+++ b/core/ranges/authors.txt
@@ -1,2 +1,3 @@
 Slava Pestov
 Doug Coleman
+Keldan Chapman
--- a/core/ranges/ranges-tests.factor
+++ b/core/ranges/ranges-tests.factor
@@ -58,3 +58,38 @@ USING: arrays kernel math ranges sequences sets tools.test ;
 
 ! Empty range
 { 0 } [ 1 0 1 <range> sum ] unit-test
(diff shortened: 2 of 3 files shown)


mrjbq7:
Whoops, I merged it too soon.

Had two issues ``USE: nath.functions`` typo.

And can't use ``basis`` library from core.

Kacarott:
Ah how did I miss that. I'll remake the PR once fixed

mrjbq7:
I jumped the gun!

It's all good.

Can ``(intersect)`` be implemented without math.functions? or do we need to have a sets.ranges type vocab that adds this behavior in during bootstrap?

Kacarott:
I am pretty sure the only function it needs from math.functions is `lcm`, but I can just inline that using `gcd` I think.

mrjbq7:
We could also move ``lcm`` to ``core:math``

PR #2851, opened 2023-09-07
Add set methods to ranges

Kacarott:
Remake of [this PR](https://github.com/factor/factor/pull/2850).

Seems to work correctly this time, with the exception of `make-my-image` causing some strange error...?

PR #2853, opened 2023-09-07
Fix for local definitions

Kacarott:
Due to locals storing their original definition (which includes many otherwise undefined words) under a `lambda` prop, using any local words in core would cause `make-my-image` to crash.

This solves that by properly emitting all the `local?` words it finds. There might be a better place to emit these words, but this seemed the cleanest and most direct. It also seems to work, which is a bonus.

I should note however that I am not very familiar with how images work, and so I am not 100% sure of the implications of this change...

This change was motivated by [this PR](https://github.com/factor/factor/pull/2851) which utilizes a locals word.


mrjbq7:
And this is because in non-optimized compiled code, the local is an uninterned word? so it doesn't show up in ``emit-words`` which emits ``all-words`` (those named and in the dictionary)...

mrjbq7:
Awesome!

PR #2855, opened 2023-09-08
Breakpoint list

tgunr:
Added code to add and remove breakpoints to a global breakfast which can be viewed by `breaklist.`


mrjbq7:
Looks like 3 different things?

1) not sure if we want an Xcode project file checked in?

2) FUEL changes look cool, but I don't use FUEL so hard for me to evaluate.

3) what's the breaklist? a list of words that have been annotated with breakpoint? it's not inclusive of words that have ``B`` in them, though.

PR #2859, opened 2023-09-10
math.functions: fold math.trig into math.functions

Capital-EX:
Move `deg>rad` and `rad>deg` to `math.functions` removing the need for `math.trig`


mrjbq7:
Merged, thanks!

What do you think of the name ``deg>rad`` versus ``degrees>radians`` ?

Capital-EX:
I think it's fine. A lot of other tool kits and languages use the terms `deg/rad` or `deg2rad/rad2deg` for converting between the two measurements.

PR #2861, opened 2023-09-10
Add documentation for sift!

chunes3:
Hello! Long-time Factor user and first-time pull requester here. I am very interested in adding documentation to Factor but I figured I'd start small to make sure I'm doing this correctly. This is documentation for the `sift!` word. Please scrutinize extra carefully to make sure I'm not messing anything up.

Old docs: 
old-sift

New docs:
new-sift

### DIFF
--- a/extra/sequences/extras/extras-docs.factor
+++ b/extra/sequences/extras/extras-docs.factor
@@ -1263,6 +1263,7 @@ HELP: sift!
     { "seq'" { "a resizable mutable " { $link sequence } } }
 }
 { $description "Removes all instances of " { $link f } " from a sequence." }
+{ $notes "The sequence " { $snippet "seq" } " MUST be growable. See " { $link "growable" } "." }
 { $side-effects "seq" }
 { $examples
     { $example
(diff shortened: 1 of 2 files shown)


alex-ilin:
Looks pretty good!
After your change, what does the "See also" section of the word `filter!` look like?

alex-ilin:
Maybe it would be a good idea to copy the "Notes" section from the `filter!` word.

chunes3:
Good idea on adding the notes section. Added.

As for the rest, are you saying it would be a good idea to add `sift!` to the `See also` of `filter!` (and maybe `sift`)?

alex-ilin:
I'm saying that due to the way `$see-also` works, your version might have replaced the original one, which was "filter, filter-as, reject, reject-as, reject!, sift, harvest". That would mean that the list got shorter.

chunes3:
`filter!` still has its original `$see-also` after the changes.

alex-ilin:
I see you are correct, I mixed `$see-also` with `related-words`.

mrjbq7:
Awesome, thanks!

PR #2868, opened 2023-09-16
Move Raylib's Code Comments into Factor's Help System

Capital-EX:
Having to read C code with lines that could span my ultrawide monitor was driving me insane, so I did something a little crazy and moved all of them into the help system.

I'll have to come back and do RLGL some other time. I nearly give myself carpal tunnel doing just Raylib.


mrjbq7:
Wow, awesome! Thanks!!

PR #2877, opened 2023-09-22
Arm64 mac

tgunr:
Code changes I made while debugging ARM64. 
In main-unix.cpp code added to test and display every combination of protection bits. Passes on Linux, fails on Mac.
In os-unix.cpp the guard pages are disabled, failure will occur later. 
This can probably go into main as the os-unix.cpp would only be compiled for Mac. The test code in main-unix is commented out.


gifti258:
Did you forget to add `Xcode/factorOSX/factor.entitlements` in 0606a0c?

PR #2878, opened 2023-09-25
parse HINFO records

theAlexes:
Just a quick hack to pull `HINFO` records, which must be two pieces of no more than 40 bytes each, length-prefixed.


mrjbq7:
Cool!

PR #2882, opened 2023-10-10
Refactor virtual sequences

nomennescio:
Refactored all applicable instances of `virtual-sequence` with `wrapped-sequence` and `sequence-view`. Leads to identical code when viewed with `optimized.`. All applicable unit-tests pass (didn't run `test-all` locally because it takes forever).

Result is significantly less use of boilerplate code. After the refactor, no more obvious candidates for refactoring remain.

__*Needs new boot image*__ because of changes in `sequences`.

### DIFF
--- a/basis/circular/circular.factor
+++ b/basis/circular/circular.factor
@@ -3,7 +3,7 @@
 USING: accessors arrays kernel math sequences strings ;
 IN: circular
 
-TUPLE: circular { seq read-only } { start integer } ;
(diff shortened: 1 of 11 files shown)


nomennescio:
I needed to make a boot image using```
disable-optimizer
make-my-image
```
because of (?) circular dependencies on `sequences`

mrjbq7:
Merged, thanks!

PR #2884, opened 2023-10-13
Anonymous Predicates

Capital-EX:
A draft PR for adding an anonymous version of predicate classes to Factor. Currently dispatch is not working correctly. Currently it seems the predicate is ignored by dispatch. Additionally, anonymous predicates are not supported in the class algebra.

What I have currently is enough to avoid raising an exception when using `M: predicate{ class [ pred ] } method`.

### DIFF
--- a/basis/bootstrap/syntax.factor
+++ b/basis/bootstrap/syntax.factor
@@ -88,6 +88,7 @@ IN: bootstrap.syntax
         "maybe{"
         "union{"
         "intersection{"
+        "predicate{"
         "initial:"
         "read-only"
         "call("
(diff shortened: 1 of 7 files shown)


mrjbq7:
What do you think of ``predicate{ fixnum [ 0 > ] }`` vs ``predicate{ fixnum 0 > }``

Capital-EX:
I'm indifferent to having the `[ ]` or ditching them. They are mostly there because it made implementing `<anonymous-predicate>` feel more straight forward.

mrjbq7:
Why is ``rank-class`` set to 1.5?

Capital-EX:
> Why is `rank-class` set to 1.5?

Temporary thing because I didn't want to change all the class ranks until I figured out how to integrate into the class algebra. All anonymous classes rank above their respective onymous classes, so I just used 1.5 to place it between `tuple-class` and `predicate-class`. 

There wasn't anything say I *couldn't* use a float so I just threw it in there.

mrjbq7:
Merged, thanks!

We might want to  change ``1.5`` to something else later, but seems to work.

PR #2885, opened 2023-10-14
Feature maximum minimum

nomennescio:
Needs new boot image

### DIFF
--- a/basis/circular/circular-tests.factor
+++ b/basis/circular/circular-tests.factor
@@ -1,6 +1,6 @@
 ! Copyright (C) 2005, 2006 Alex Chapman, Daniel Ehrenberg
 ! See https://factorcode.org/license.txt for BSD license
-USING: arrays circular kernel math sequences sequences.private
+USING: arrays circular kernel literals math sequences sequences.private
 strings tools.test ;
 IN: circular.tests
 
@@ -18,6 +18,9 @@ IN: circular.tests
 { [ 2 3 1 ] } [ { 1 2 3 } <circular> 1 over change-circular-start [ ] like ] unit-test
 { [ 3 1 2 ] } [ { 1 2 3 } <circular> 1 over change-circular-start 1 over change-circular-start [ ] like ] unit-test
 { [ 3 1 2 ] } [ { 1 2 3 } <circular> -100 over change-circular-start [ ] like ] unit-test
+
+{ $[ { 1 2 3 } minimum ] } [ { 1 2 3 } <circular> minimum ] unit-test
+{ $[ { 1 2 3 } maximum ] } [ { 1 2 3 } <circular> maximum ] unit-test
 
 { "fob" } [ "foo" <circular> CHAR: b 2 pick set-nth >string ] unit-test
 { "boo" } [ "foo" <circular> CHAR: b 3 pick set-nth-unsafe >string ] unit-test
--- a/basis/circular/circular.factor
+++ b/basis/circular/circular.factor
@@ -18,6 +18,9 @@ PRIVATE>
 
 M: circular virtual@ circular-wrap seq>> ; inline
 
+M: circular minimum seq>> minimum ; inline
+M: circular maximum seq>> maximum ; inline
+
 : change-circular-start ( n circular -- )
     ! change start to (start + n) mod length
     circular-wrap start<< ; inline
--- a/core/sequences/sequences-tests.factor
+++ b/core/sequences/sequences-tests.factor
@@ -390,8 +390,6 @@ M: bogus-hashcode hashcode* 2drop 0 >bignum ;
 { 24 } [ { 1 2 } { 3 4 } [ + ] [ * ] 2map-reduce ] unit-test
 { 2 96 } [ 2 { 3 3 3 3 } { 4 4 4 4 } [ [ dup ] 2dip * * ] [ + ] 2map-reduce ] unit-test
 
-{ 4 } [ 5 <iota> (maximum) ] unit-test
-{ 0 } [ 5 <iota> (minimum) ] unit-test
 { 4 } [ 5 <iota> supremum ] unit-test
 { 0 } [ 5 <iota> infimum ] unit-test
 { 4 } [ 5 <iota> maximum ] unit-test
--- a/core/sequences/sequences-tests.factor
+++ b/core/sequences/sequences-tests.factor
@@ -390,6 +390,13 @@ M: bogus-hashcode hashcode* 2drop 0 >bignum ;
 { 24 } [ { 1 2 } { 3 4 } [ + ] [ * ] 2map-reduce ] unit-test
 { 2 96 } [ 2 { 3 3 3 3 } { 4 4 4 4 } [ [ dup ] 2dip * * ] [ + ] 2map-reduce ] unit-test
 
+{ 4 } [ 5 <iota> (maximum) ] unit-test
+{ 0 } [ 5 <iota> (minimum) ] unit-test
+{ 4 } [ 5 <iota> supremum ] unit-test
+{ 0 } [ 5 <iota> infimum ] unit-test
+{ 4 } [ 5 <iota> maximum ] unit-test
+{ 0 } [ 5 <iota> minimum ] unit-test
+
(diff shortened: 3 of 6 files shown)


mrjbq7:
You shouldn’t need (minimum) and (maximum) words. In the one place it’s used you can just use call-next-method On Oct 13, 2023, at 5:36 PM, nomennescio ***@***.***> wrote:﻿
@nomennescio requested your review on: #2885 Feature maximum minimum.

nomennescio:
> You shouldn’t need (minimum) and (maximum) words. In the one place it’s used you can just use call-next-method On Oct 13, 2023, at 5:36 PM, nomennescio ***@***.***> wrote:﻿ @nomennescio requested your review on: #2885 Feature maximum minimum. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: ***@***.***>

These are utility words also used at another place in the implementation, and available if more types need such "worker" words.

mrjbq7:
It should just be in M\ sequence minimum. There is no need for utility words.On Oct 13, 2023, at 5:44 PM, nomennescio ***@***.***> wrote:﻿

You shouldn’t need (minimum) and (maximum) words. In the one place it’s used you can just use call-next-method On Oct 13, 2023, at 5:36 PM, nomennescio @.> wrote:﻿ @nomennescio requested your review on: #2885 Feature maximum minimum. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: @.>

These are utility words also used at another place in the implementation, and available if more types need such "worker" words.

nomennescio:
> It should just be in M\ sequence minimum. There is no need for utility words.On Oct 13, 2023, at 5:44 PM, nomennescio ***@***.***> wrote:﻿ You shouldn’t need (minimum) and (maximum) words. In the one place it’s used you can just use call-next-method On Oct 13, 2023, at 5:36 PM, nomennescio @.> wrote:﻿ @nomennescio requested your review on: #2885 Feature maximum minimum. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: @.> These are utility words also used at another place in the implementation, and available if more types need such "worker" words. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: ***@***.***>

Hmm, did you check this usecase?
```
M: cycles minimum dup full-cycle? [ circular>> minimum ] [ (minimum) ] if ; inline
```

mrjbq7:
call-next-methodOn Oct 13, 2023, at 5:49 PM, nomennescio ***@***.***> wrote:﻿

It should just be in M\ sequence minimum. There is no need for utility words.On Oct 13, 2023, at 5:44 PM, nomennescio @.> wrote:﻿ You shouldn’t need (minimum) and (maximum) words. In the one place it’s used you can just use call-next-method On Oct 13, 2023, at 5:36 PM, nomennescio @.> wrote:﻿ @nomennescio requested your review on: #2885 Feature maximum minimum. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: @.> These are utility words also used at another place in the implementation, and available if more types need such "worker" words. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: @.>

Hmm, did you check this usecase?
M: cycles minimum dup full-cycle? [ circular>> minimum ] [ (minimum) ] if ; inline

mrjbq7:
I apologize if that was cryptic, I was running errands and trying to respond by email.

This is what I mean:

```factor
GENERIC: minimum ( seq -- elt )
M: sequence minimum [ ] [ min ] map-reduce ; inline

GENERIC: maximum ( seq -- elt )
M: sequence maximum [ ] [ max ] map-reduce ; inline

M: cycles minimum 
    dup full-cycle? [ circular>> minimum ] [ call-next-method ] if ; inline
M: cycles maximum 
    dup full-cycle? [ circular>> maximum ] [ call-next-method ] if ; inline
```

nomennescio:
I might be missing something, but if I try it like you propose building a boot image crashes, or when I just reload sequences Factor hangs and I need to kill it. I suspect an infinite compile loop because of the types? Didn't fully analyze yet, first need coffee.

nomennescio:
Allright, tried the commandline listener and tried to recompile, then was able to get a meaningful error message. Issue was a typo leading to a compiler issue that would hang the UI listener. Fixed the typo. Now all works. Will make a new commit out of this.

mrjbq7:
Looks pretty good, I'd change a few things:

* Instead of using literals in the test ``$[ { 1 2 } minimum ]``, I think it would be better to put the correct answer.
* I'm not wild about all the inlining, are there any examples of normal code that benefit from it? 
* It's probably better to make ``supremum`` and ``infimum`` aliases instead of duplicated generics?
* It's only a style thing, and if you feel strongly then reject this point, but the cycles generic methods are like 89 chars wide, and I know we have some occasional large lines in the stdlib but I like to try and make them go onto a second line when that happens at least.

nomennescio:
> Looks pretty good, I'd change a few things:
> 

Thanks, some comments

> * Instead of using literals in the test `$[ { 1 2 } minimum ]`, I think it would be better to put the correct answer.

My thinking was to make the intent clear that we're testing the specialization against the general `sequence` minimum.
I don't mind changing it to the correct answer though.

> * I'm not wild about all the inlining, are there any examples of normal code that benefit from it?

That I find a very difficult question to answer, I'm not sure what we should do to evaluate benefits versus drawbacks. To me it seems intuitive to use inline where the effect is to just add a small piece of code (like `seq>>`) instead of adding a dispatch. The folding and elimination of multiple dispatches in a row is just so nice. I agree if code is "too" large you would need a bit more effort to justify.

I think it would be nice if we could have standardized benchmarks on speed and code size to measure the effects in real code.

> * It's probably better to make `supremum` and `infimum` aliases instead of duplicated generics?

My thinking here is that we might introduce types that are not finite sequences like `sequence` but more like mathematical generated infinite sequences which don't have a maximum or minimum, but do have a  `supremum` and `infimum`, which can then be used to specialize. You could even use that to generate an error for infinite sequences (e.g. a Fibonacci infinite sequence) that *don't* have a supremum, without the need to get stuck trying to traverse an infinite sequence.

> * It's only a style thing, and if you feel strongly then reject this point, but the cycles generic methods are like 89 chars wide, and I know we have some occasional large lines in the stdlib but I like to try and make them go onto a second line when that happens at least.

No problem.

> * We should add a fixups for supremum/infimum as they are likely to be deleted at some point, maybe even before 0.100.

See my point why I think supremum/infimum are still valuable. I agree at many points in the current code we could replace these calls. Maybe a second pull request?

And what about the names? Do you want to finish discussing `maximum` versus `max-of` (7 vs 6 chars) before committing?

mrjbq7:
Haha ``(7 vs 6 chars)`` as much as I love all our code golf friends, I don't think our primary motivator should be density of naming :-)

Maybe before we merge it we can check with a few people to get their opinion @gifti258 @Kacarott @razetime @erg.

I lean slightly towards ``min-of`` if only because then it's closer to ``min`` as a name, but i also really prefer english names so ``minimum`` is good for that...

My thinking on the test cases was just that if there's a bug in the super-class minimum that passes through to the subclass they'll both be the same but both be wrong... I suppose that would be detected other ways, just thought it's cleaner testing a specific answer than a computed answer.

For something like someday we might need a supremum that works differently, when that someday comes we can add it back and it will probably look different...

razetime:
I prefer `maximum` and `minimum` as they are single english words. it also lines up better with the other max and min words in `math.statistics`.

mrjbq7:
Okay, well I'm fine with that -- @nomennescio want to take a look at those other supremum/infimum words for renames before we merge this?

nomennescio:
> Okay, well I'm fine with that -- @nomennescio want to take a look at those other supremum/infimum words for renames before we merge this?

Sorry, missed this. I don't get your point, which other words?

mrjbq7:
No worries, i'll work on this

mrjbq7:
Okay, i merged these and a few others, lets see how it settles out

PR #2890, opened 2023-10-18
editors: Added test for brew installation of VSCodium.

tgunr:
Was using VSCodium and it was not being found on Mac.


erg:
fixed a different way in beaa2d479348c38c257c8f4e191685e6a022da79

thanks!

PR #2893, opened 2023-10-20
Fix: correct gl-function loading

TriedAngle:
suffix appends the "modern" function at the end, which makes it so the old deprecated opengl1 function gets loaded which are not part of the opengl standard anymore and make usage of shader debugger like renderdoc impossible. This fixes it and should also preserve functionality on old hardware.


mrjbq7:
Awesome, I see - so this changes it to look in the order that it's stored and pick the first available function. It used to find old-before-new and now its new-before-old.

Thanks!!!

PR #2903, opened 2023-10-24
Update README.md

harshhere905:


erg:
No, this is grammatically incorrect.

tgunr:
What the heck is going on here with all these spam attempts. I just don't get it, I guess they didn't get enough attention as a child.

PR #2905, opened 2023-10-27
Update README.md

harshhere905:


gifti258:
Platform support and documentation are not features of the library but the language.

harshhere905:
Hey @gifti258 I came to know  that you are right and making pr for this issue was wrong... And I would .ake any further pr's lke this in the future.. Please can

PR #2910, opened 2023-11-01
calendar: do not call day-of-week twice in weekday?

blin:
Calling `today weekday?` fails with 

```
Generic word year>> does not define a method for the fixnum class.
Dispatching on object: 3
```

Sorry for not including the tests, didn't have the time to figure how to do that yet.


razetime:
if you do `"calendar" edit-tests` in the listener it should open the file for you. Should be easy to understand unit tests through the `tools.test` vocab docs after that.

I will merge this once the regression test is added.

blin:
Done, please take another look.

mrjbq7:
Awesome, thank you!!

PR #2913, opened 2023-11-05
farkup: revent literal from being modified

Capital-EX:
Found a bug while doing some post-processing on the generated XML from Farkup. All tags generated by Farkup share the same `attrs` object. This make modifying this attribute after the fact hazardous.

This PR fixes that.


mrjbq7:
Cool!! Thanks!!

PR #2914, opened 2023-11-06
assocs: update docs and give example with hashtable

Capital-EX:
Remove out of date note about `zip-as`. Additionally add example with hashtable.


mrjbq7:
Thanks!!

PR #2919, opened 2023-11-09
added vocab prettyprinter to classes in extra

bitli:
New vocab to display the class hierarchy.
Pls check if the doc is OK, 
The methods :: print-children ( h c i -- ) uses local variable, not sure if use stack operation is possible/desirable.

### DIFF
--- a/extra/classes/prettyprinter/prettyprinter-docs.factor
+++ b/extra/classes/prettyprinter/prettyprinter-docs.factor
@@ -1,0 +1,34 @@
+! Copyright (C) 2023 Jean-Marc Lugrin.
+! See https://factorcode.org/license.txt for BSD license.
+USING: help.markup help.syntax ;
+IN: classes.prettyprinter
+
+ARTICLE: "classes.prettyprinter" "Print the hierarchy of a class"
+{ $vocab-link "classes.prettyprinter" } " supports the printing of the class hierarchy to the listener or to any text stream." $nl
+"The class name and vocab name are clickable, a P indicates that the classs is PRIVATE." $nl 
+"See " { $link hierarchy. } "."
+;
+
+HELP: hierarchy.
+{ $values { "class" "a class, use " { $snippet "tuple" } " to print the whole hierarchy" }  }
+{ $description "Print the class hierarchy layout on the output stream, with the name of its vocabulary." }
+{ $errors "Throws an error if " { $snippet "class" } " is not a class." }
+{ $examples
+    { $unchecked-example "tuple hierarchy." }  
+    { $unchecked-example " \"GADGETS.TXT\" utf8 [ gadget hierarchy. ] with-file-writer" } 
+}
+;
+
+HELP: class-hierarchy
+{ $values { "hash" "An " { $snippet "hashtable" } " mapping classes to list of children" }  }
+{ $description "Extract the hierarchy of all classes in the form of an hashtable" 
+    "with the class as the key and a vector of all children class in alphabetic order as a child." }
+{ $notes { "This is made public for other tools that want to explore the class hierarch." } }
+;
+
+HELP: not-a-class-error
+{ $description "Throws a " { $link not-a-class-error } "." }
+{ $error-description "Thrown by " { $link hierarchy. } " if the parameter is not a class." } 
+;
+
+ABOUT: "classes.prettyprinter"
--- a/extra/classes/prettyprinter/prettyprinter.factor
+++ b/extra/classes/prettyprinter/prettyprinter.factor
@@ -1,0 +1,91 @@
+! Copyright (C) 2023 Jean-Marc Lugrin.
+! See https://factorcode.org/license.txt for BSD license.
+
+! Print a class heierarchy in the listener
+! or to a file: "TESTDUMP.TXT" utf8 [ gadget hierarchy. ] with-file-writer
+
+
+USING:  classes hashtables ui.gadgets assocs kernel sequences prettyprint  vectors 
+math io formatting strings sorting accessors io.styles vocabs ;
+
+IN: classes.prettyprinter
+
+ERROR: not-a-class-error ;
+
+<PRIVATE
+
+CONSTANT: in-col 40
+
+: add-child  ( c h  -- )
+    over ! ( c h -- c h c )
+    superclass-of  ! ( c h c -- c h s )
+    swap at* ! (c h s -- c s h -- c v ? ) vector for superclass but if f ignore
+        [ swap suffix! drop ]  ! ( c v -- )
+        [ 2drop ] 
+    if 
+;
+
+: print-class-name ( c -- )
+    dup name>> swap write-object
+;
+
+
+: print-in ( c -- )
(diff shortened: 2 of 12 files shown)


extra/classes/prettyprinter/prettyprinter-docs.factor
```diff
@@ -0,0 +1,34 @@
+! Copyright (C) 2023 Jean-Marc Lugrin.
+! See https://factorcode.org/license.txt for BSD license.
+USING: help.markup help.syntax ;
+IN: classes.prettyprinter
+
+ARTICLE: "classes.prettyprinter" "Print the hierarchy of a class"
+{ $vocab-link "classes.prettyprinter" } " supports the printing of the class hierarchy to the listener or to any text stream." $nl
+"The class name and vocab name are clickable, a P indicates that the classs is PRIVATE." $nl
```
mrjbq7, line 8: s/classs/class

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
+! Copyright (C) 2023 Jean-Marc Lugrin.
+! See https://factorcode.org/license.txt for BSD license.
+
+! Print a class heierarchy in the listener
```
mrjbq7, line 4: typo in heierarchy

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+
+: print-in ( c -- )
+    vocabulary>> lookup-vocab dup name>> 
+    dup ".private" tail? [ " P" ]  [ "  " ] if  write ! Mark if private
+    " IN: " write 
+    swap write-object 
+;
+
+: print-leader ( i -- )
+    ! 2 * CHAR: . <string>  write ! leader
```
mrjbq7, line 41: once you get it working, i suggest generally removing alternative versions of things -- unless they provide good explanatory context, or you use the lines to swap between two implementations or something

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+: print-class ( c i -- ) 
+    2dup print-leader print-class-name ! ( c i -- c i )
+    2 * over name>> length + ! Current column  c ci -- 
+    dup in-col < [ in-col swap - CHAR: \x20 <string>  write  ] [ drop  ] if
+    print-in 
+    nl 
+;
+
+: print-superclasses ( c -- )
+    superclass-of dup
```
mrjbq7, line 55: the form ``dup [ ... ] [ drop ] if`` can be simplified to ``[ ... ] when*``

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+;
+
+: print-root-class ( c -- ) 
+    [ print-class-name ]
+    [ print-superclasses ] bi
+    nl
+;
+
+ :: print-children ( h c i -- )
+    i 0 = [ c print-root-class ] [ c i print-class  ] if
```
mrjbq7, line 71: this could be ``c i [ print-root-class ] [ print-class ] if-zero``

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+    c h at
+    [ h swap i 1 + print-children ] each
+ ;
+
+PRIVATE>
+
+: class-hierarchy ( -- hash )
+    ! Hastable class -> empty mutable vector
```
mrjbq7, line 79: typo in hastable

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+    classes  [ drop V{ } clone ] zip-with >hashtable
+    ! for each child-class, add it to its parent vector
+    classes [  over add-child ] each
+    [ sort ] assoc-map
+;
+
+: hierarchy. ( class -- )
+    dup class? [ not-a-class-error ] unless
```
mrjbq7, line 87: we have a utility word the helps eliminate these kind of specific type checking errors: ``class check-instance``

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+    i 0 = [ c print-root-class ] [ c i print-class  ] if
+    c h at
+    [ h swap i 1 + print-children ] each
+ ;
+
+PRIVATE>
+
+: class-hierarchy ( -- hash )
+    ! Hastable class -> empty mutable vector
+    classes  [ drop V{ } clone ] zip-with >hashtable
```
mrjbq7, line 80: we have some ``-as`` words that allow you to do the type conversion as part of the action, so it's slightly faster and more typical to do ``H{ } zip-with-as`` rather than ``zip-with >hashtable``

mrjbq7:
Pretty cool!!

Gave some minor feedback above.

A couple of options here:

1) fix/comment on the code feedback then let me know when ready to merge
2) merge as-is and we can work on it sometime in the future

I'm happy to keep the current name ``classes.prettyprinter`` but it sorta conflicts with our ``foo.prettyprint`` vocabulary that is a convention for implementing ``pprint*`` on types. I don't have a real suggestion, except maybe ``classes.hierarchy``, but I'm also fine keeping it as its named.

extra/classes/prettyprinter/prettyprinter.factor
```diff
@@ -0,0 +1,91 @@
...
+IN: classes.prettyprinter
+
+ERROR: not-a-class-error ;
+
+<PRIVATE
+
+CONSTANT: in-col 40
+
+: add-child  ( c h  -- )
+    over ! ( c h -- c h c )
```
mrjbq7, line 20: it is often useful to have stack effect comments like this as you learn factor and/or build something complex, but then we often remove them before pushing unless the stack effect comment is particularly useful for some internal non-obvious quotation

extra/classes/prettyprinter/prettyprinter.factor
bitli: Thanks.  Just one thing: the exception becomees "not and instance" instead of "not a class".  A class is indeed an instance of `class`, but I wonder if this not confusing (may be not for expert factor programmers ?)

mrjbq7: Yeah, that's a fair point.  It mostly replaced not-a-string, not-an-integer, not-a-byte-array, etc... I'm fine with anything you want to do, just making suggestions.

bitli:
I have no strong opinion on the name, I selected prettyprinter because the prettyprinter has stuff like callstack. or object-table. that serve about the same instance. But it may be confusing, as the vocab name is not exactly the same.
The other utilities I prepared is (currently) in a vocabulary 'gadget.hierarchy'. If that one is ok for the gadget, then classes.hierarchy could be good for this one, so they are consistent. What do you think ?

mrjbq7:
Thanks!!

PR #2921, opened 2023-11-14
misc: Add new dict for multiline comments

TheWitheredStriker:
Hello. This PR adds a new entry to the Factor grammar, which highlights the block comments `/* Comment */` provided by the `multiline` vocabulary. Previously this was not the case, leading to results like this:

image

I can testify that the new dict works:

image
(I appended block comments to a random snippet from Rosetta Code in order to quickly test them)

Feel free to test the grammar yourself using [NovaLightshow](https://novalightshow.netlify.app/).

If there are any problems with this PR that I should address, please do not hestitate to inform me. This is my first time contributing to Factor (or to any programming language for that matter). I have read the contribution guidelines and believe that all of the guidelines are accounted for, but I might be mistaken; please let me know if I am.

To whoever looks over this PR, I hope you will have a good day!

### DIFF
--- a/misc/Factor.tmbundle/Syntaxes/Factor.tmLanguage
+++ b/misc/Factor.tmbundle/Syntaxes/Factor.tmLanguage
@@ -372,6 +372,14 @@
 			<string>comment.parens.factor</string>
 		</dict>
 		<dict>
+			<key>begin</key>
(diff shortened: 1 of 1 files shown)


mrjbq7:
Thanks!

PR #2922, opened 2023-12-06
 Range methods

Kacarott:
Added more efficient vector methods to ranges. I added them in `math.vectors` due to it being in basis. Review, edit, changes, are of course all welcome.

### DIFF
--- a/basis/math/vectors/vectors-tests.factor
+++ b/basis/math/vectors/vectors-tests.factor
@@ -1,5 +1,5 @@
 USING: math.vectors tools.test kernel specialized-arrays compiler
-kernel.private alien.c-types math.functions ;
+kernel.private alien.c-types math.functions ranges arrays ;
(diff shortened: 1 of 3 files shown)


mrjbq7:
Another option is making a ``math.vectors.ranges`` and then adding a require-when

```
{ "math.vectors" "ranges" } "math.vectors.ranges" require-when
```

So that you don't have conjoined implementations....

Kacarott:
Is using `ranges.private` better? I was under the impression that using `.private` vocabs was a great sin, but I can happily change that.

About your second idea, I have never even heard of `require-when`, that might be a nice approach. Does the `require-when` line live in `math.vectors` then?

mrjbq7:
I have no problem with it using ``ranges.private``, I also don't really have a problem with making ``>range<`` public, just that if nobody is using it, maybe its not necessary to make it public... don't have a strong opinion one way or the other

PR #2930, opened 2023-12-28
math.combinatorics: fix <k-permutations> for k=0

olus2000:
There is always one 0-permutation, the empty permutation. Previous code produced no permutations for k=0. I removed the check that broke that case.

`load-all \ <k-permutations> usage` tells me that this word isn't used in any definitions.


erg:
Can you please add a unit test with the expected output so we document the behavior we want and don't revert it someday without fixing a test?

erg:
Merged. Thanks!

PR #2934, opened 2024-01-15
sequences.parser: Rename and redefine `next`

olus2000:
Previous definition of `next` returned the current element, which made it inconsistent with `peek-next` which returns the following element. I changed the word to instead return the following element, and renamed the old behavior to `consume`, which doesn't seem to have relevant collisions in other vocabs.

Behavior I named `consume` is used in `compression.run-length`, `html.parser` and `lint.vocabs` vocabs. Behavior I named `next` I see useful as a user of the vocab.

A big change that I would like to request a comment on is moving the vocab to `extra`. It was mentioned in passing in the discussion on discord, but only later I noticed that the vocab is used by `compression.run-length` which is in `basis`.

I also added tests to affected words, and documentation to some words in the vocab including affected words.


mrjbq7:
Looks good! Because of the way our builders work, they make sure the core only uses core, basis only uses core/basis, and extra uses core/basis/extra, so we'll have to either move ``compression.run-length`` to extra, or move it back. I'm inclined to move that to extra also.

PR #2948, opened 2024-02-16
misc/vim/ftplugin: Add autopairs, colorcolumn instead of margin error

gifti258:
This adds an optional autopairs functionality.

It also disposes of highlighting overlong lines as errors. Instead I think it would be nicer to use the less glaring mechanism of the colorcolumn option, as the line length limit is often not adhered to.

### DIFF
--- a/misc/vim/README.md
+++ b/misc/vim/README.md
@@ -1,58 +1,73 @@
 Vim support for Factor
 ======================
 
-This directory contains various support files that make editing Factor code
-more pleasant in Vim.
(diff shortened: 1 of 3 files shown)


mrjbq7:
i find the error highlight glaring sometimes, but the color column is weird too.  i'll give it a try for a bit.

mrjbq7:
Thanks !

mrjbq7:
As an aside, i wonder if we should relax it to 78 characters or something

PR #2953, opened 2024-02-25
Improve scrolling on Windows

alex-ilin:
Support the horizontal scroll wheel on Windows.

Increase the default scrolling speed, which is too low on Windows (1.5 lines of text in the Listener instead of the standard 3). Add the option to scroll 10 times faster while pressing Shift.


mrjbq7:
Merged, thanks!!

PR #2954, opened 2024-02-27
Sqlite3 rename

erg:
- dont use c-string at all since we are autogenerating bindings
- rename all sqlite- to sqlite3-

dunno what we want here, seeing how it looks


erg:
I don't really like these changes.

PR #2957, opened 2024-03-07
math.parser: add arbitrary base conversion words

razetime:
Deals with sequences of numbers exclusively. An alphabet can be used later to create a string.

Looking for feedback.


razetime:
made some naming changes and added convenience words based on @gifti258 's feedback.

mrjbq7:
Should we merge this?

Any thoughts about naming consistency with ``: digit> ( ch -- n )`` and ``: >digit ( n -- ch )``?

mrjbq7:
@razetime want to finish merging this? any thoughts on the naming consistency with the other digit words?

mrjbq7:
I guess I'm wondering if we should swap the names of either ``>digit`` or ``>base-digits``.

```factor
IN: scratchpad 15 >digits .
{ 1 5 }
IN: scratchpad 1 >digit .
49
```

I would think if we are making "digits" they would have the same definition.

mrjbq7:
Hmm, I guess ``>digit`` says "Outputs a character representation of a digit.", so technically it's saying it takes a "digit" which and makes a ch.

mrjbq7:
Well think about that and if you are fine with it, lets merge!

PR #2961, opened 2024-04-03
discord: fix interaction create, add interaction edit

valpackett:
Some improvements that I needed to patch while working on https://github.com/booniepepper/dt-discord-bot for @booniepepper:

- `create-interaction-response` that isn't fixed to respond "pang" :D and is the right endpoint for calling when responding to a websocket notification, not a webhook
- `edit-interaction-response` as there's a very quick first response deadline and we needed to run code, which could take a second, so we edit that response after execution is done

Not included: I also needed to patch `gateway-identify-json` to use `"intents": 1` to make it work with a bot key that has very limited (interaction only) permissions; I'm not sure what would be the best way to handle this, but it would be nice if there was a convenient way to tell the library which permissions are needed.

### DIFF
--- a/extra/discord/discord.factor
+++ b/extra/discord/discord.factor
@@ -73,6 +73,8 @@ TUPLE: discord-bot
     discord-post-request json-request ;
 : discord-post-json ( payload route -- json )
     [ >json ] dip discord-post-request add-json-header json-request ;
+: discord-post-json-no-resp ( payload route -- )
+    [ >json ] dip discord-post-request add-json-header http-request 2drop ;
 : discord-patch-json ( payload route -- json )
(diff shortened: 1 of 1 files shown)


erg:
Hey! I merged your patch and added an `intents` slot to the `discord-bot-config` in f9c620851258488f563f9a38cfb8585553944b2c

Should look like:

```factor
USING: discord namespaces ;
T{ discord-bot-config
    { client-id "3229392" }
    { client-secret "fsdasdfasfda" }
    { token "asdfasdf" }
    { application-id "1289128" }
    { intents 1 }
}
discord-bot-config set-global
```

Maybe we could import the intent names instead of using these magic numbers.

PR #2968, opened 2024-05-02
Build infra

nomennescio:
Stores history of runs and benchmarks in DB.
Next step would be adding functionality to view formatted benchmark results in `webapp.mason`.

### DIFF
--- a/extra/mason/authors.txt
+++ b/extra/mason/authors.txt
@@ -1,2 +1,3 @@
 Eduardo Cavazos
 Slava Pestov
+nomennescio
--- a/extra/mason/build/build.factor
+++ b/extra/mason/build/build.factor
@@ -45,7 +45,7 @@ IN: mason.build
         build-child
         [ notify-report ] [
             status-clean eq?
-            [ notify-upload upload-docs release ] when
+            [ notify-benchmarks notify-upload upload-docs release ] when
         ] bi
         notify-finish
         finish-build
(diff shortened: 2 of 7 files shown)


mrjbq7:
This might potentially insert multiple runs per gitref, due to rebooting the masons, if that matters, and we should make sure we don't implicitly sort by run-id if we insert the historical benchmarks.

Also, it's de-normalized but the factor db tuple api doesn't quite make normalizing queries properly, so i assume you want to do something like select host/os/cpu on the /speed endpoint (and then optionally maybe a benchmark name in some dropdown), and then render charts.  Or there aren't that many benchmarks, could just render them all after choosing host/os/cpu.

I was thinking maybe the primary key should just be host/os/cpu/gitref instead of a run-id. But maybe that doesn't matter, would be silly to have a flat line in the charts because we kicked the masons a few times while doing whatever with them.

mrjbq7:
I sorta thought storing it as de-normalized rows of host/os/cpu/gitref/benchmark/time would be simpler... but up to you.

nomennescio:
> I sorta thought storing it as de-normalized rows of host/os/cpu/gitref/benchmark/time would be simpler... but up to you.

True, but I was concerned about the size of the DB increasing unnecessarily.

nomennescio:
> This might potentially insert multiple runs per gitref, due to rebooting the masons, if that matters, and we should make sure we don't implicitly sort by run-id if we insert the historical benchmarks.

Multiple runs per gitref are not a problem, just give more accuracy to benchmark data. Furthermore, the data can't discern aspects that impact runtime that aren't caught by a gitref, such as changes to build machines configurations. If that becomes important, additional characteristics can be added to make that distinction.

PR #2969, opened 2024-05-02
Build infra

nomennescio:
Add basic display and filtering of performance results. This is ready to be put in production, if you agree.
Any fancy markup can be done in the future.


mrjbq7:
I think this is merged, I had to make a small patch in 92edb2594ca233ddf92bf09e9563d5f133e28e22 as we weren't getting benchmarks uploaded properly.

nomennescio:
> I think this is merged, I had to make a small patch in [92edb25](https://github.com/factor/factor/commit/92edb2594ca233ddf92bf09e9563d5f133e28e22) as we weren't getting benchmarks uploaded properly.

I added showing benchmark results; but because this is on a branch, it still shows the original commits too

mrjbq7:
I kicked off the new mason's lets see if they upload benchmarks into the db

Then let's see if the endpoint works!

nomennescio:
> I kicked off the new mason's lets see if they upload benchmarks into the db
> 
> Then let's see if the endpoint works!

It works! https://builds.factorcode.org/benchmark-results
I do however see a problem with nanoseconds being interpreted as negative numbers,
I will check and make a fix.

PR #2977, opened 2024-05-08
Feature vm compression

nomennescio:
Ready to merge


vm/zstd.hpp
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2022-2024 nomennescio
+// See https://factorcode.org/license.txt for BSD license.
+
+namespace lib { namespace zstd { extern "C" {
```
mrjbq7, line 4: why does it need a C++ namespace instead of just including ``zstd.c`` in image.cpp?

mrjbq7:
Should we link against a compression library instead of including it directly?

mrjbq7:
Is it really 50ms of overhead?

That's a lot of overhead versus running no code:

```
$ time ./factor -e=""

real	0m0.098s
user	0m0.036s
sys	0m0.062s
```

vm/zstd.hpp
nomennescio: Because zstd.c -- which is generated externally and I don't want to edit -- by default put everything in the global namespace, which can lead to name clashes. This prevents any (future) name clash.

nomennescio:
> Should we link against a compression library instead of including it directly?

I thought of that, but that is unsafe; so early in the vm life cycle phase, we then need to rely on OS level loading, which can fail. What's worse, the OS could load the wrong version of the library.

> Is it really 50ms of overhead?
> 
> That's a lot of overhead versus running no code:
> 
> ```
> $ time ./factor -e=""
> 
> real	0m0.098s
> user	0m0.036s
> sys	0m0.062s
> ```

Full overhead for decompressing a 120MB full Factor image. I think it's fully worth it; if you care about 50 ms for starting Factor, you should just use an uncompressed image, which is the default anyway. Maybe that time can improve, but I think this is already quite optimal, as I avoided allocating additional buffers when possible.

nomennescio:
> Should we link against a compression library instead of including it directly?

> Should we link against a compression library instead of including it directly?

I thought of that, but that is unsafe; so early in the vm life cycle phase, we then need to rely on OS level loading, which can fail. What's worse, the OS could load the wrong version of the library.

nomennescio:
> > Should we link against a compression library instead of including it directly?

As a mitigation of the slight increase executable size due to static linking, I was already thinking of adding *native* zstd support to Factor, reusing that code.

nomennescio:
> Is it really 50ms of overhead?
> 
> That's a lot of overhead versus running no code:
> 
> ```
> $ time ./factor -e=""
> 
> real	0m0.098s
> user	0m0.036s
> sys	0m0.062s
> ```

Just to be sure, did you run that with a *compressed image*? What is the timing for loading with an *uncompressed image*?

nomennescio:
I'll also be doing a compression experiment with a `load-all` image; if we can practically start using that, then run-time loading of vocabs can be prevented at all.

nomennescio:
> I'll also be doing a compression experiment with a `load-all` image; if we can practically start using that, then run-time loading of vocabs can be prevented at all.

Got some issues with `load-all` on Windows, but nevertheless:
```
-rwxrwxr-x+ 1  129890744 May  8 23:57 factor.image
-rwxrwx---+ 1  571727240 May 10 15:18 factor-all.image
-rwxrwx---+ 1   78380047 May 10 15:28 factor-all.image.compressed
```

i.e. the compressed image for a load-all, is smaller than a normal uncompressed image. Its compression factor is 7.3.

mrjbq7:
What issues did you have with load-all?On May 10, 2024, at 6:34 AM, nomennescio ***@***.***> wrote:﻿

I'll also be doing a compression experiment with a load-all image; if we can practically start using that, then run-time loading of vocabs can be prevented at all.

Got some issues with load-all on Windows, but nevertheless:
-rwxrwxr-x+ 1  129890744 May  8 23:57 factor.image
-rwxrwx---+ 1  571727240 May 10 15:18 factor-all.image
-rwxrwx---+ 1   78380047 May 10 15:28 factor-all.image.compressed

i.e. the compressed image for a load-all, is smaller than a normal uncompressed image. Its compression factor is 7.3.

nomennescio:
> What issues did you have with load-all?

Just missing dlls, and one accidently bound (?) to Outlook which give a funny warning:
```
(process:17792): GLib-GIO-WARNING **: 14:21:52.240: Unexpectedly, UWP app `Micro
soft.OutlookForWindows_1.2023.1114.100_x64__8wekyb3d8bbwe' (AUMId `Microsoft.Out
lookForWindows_8wekyb3d8bbwe!Microsoft.OutlookforWindows') supports 1 extensions
 but has no verbs
```

vm/zstd.hpp
nomennescio: And to keep compilation snappy, I prefer to prevent unnecessary source dependencies; `zstd.o` won't change until a new version of the lib is generated. (In that sense it's a bit unfortunate Factor uses one `master.hpp` entangling all dependencies.)

mrjbq7:
> Just to be sure, did you run that with a _compressed image_? What is the timing for loading with an _uncompressed image_?

this is a current default 130 mb *uncompressed* image, on linux-64

nomennescio:
> time ./factor -e=""

Right, thanks for the feedback; I ran a single time on my Windows machine
```$ time ./factor.com -i=factor.image.compressed -e=""

real    0m0.982s
user    0m0.000s
sys     0m0.015s
/c/dev/project/build/factor [30104]
$ time ./factor.com -i=factor.image -e=""

real    0m0.934s
user    0m0.000s
sys     0m0.000s
```
so about 50 ms on top of 930 ms. Can't test the Linux times.

mrjbq7:
Those times don’t look right. I think your Factor is loading stuff on startup maybe your -rc file. Run factor, then ‘save’ then try the compare again. On May 10, 2024, at 2:34 PM, nomennescio ***@***.***> wrote:﻿

time ./factor -e=""

Right, thanks for the feedback; I ran a single time on my Windows machine

real    0m0.982s
user    0m0.000s
sys     0m0.015s
/c/dev/project/build/factor [30104]
$ time ./factor.com -i=factor.image -e=""

real    0m0.934s
user    0m0.000s
sys     0m0.000s

so about 50 ms on top of 930 ms. Can't test the Linux times.

nomennescio:
> Those times don’t look right. I think your Factor is loading stuff on startup maybe your -rc file. Run factor, then ‘save’ then try the compare again.

Yes, it does a `USE:` on an editor, to set that variable. I think that's quite typical use, but nothing else.

But I don't understand why you are so focusing on 50 ms for decompressing. I don't think it's relevant at all, as it's a totally reasonable amount of time to decompress into 130 MB, you cannot avoid it if you want decompression, and it's purely optional to begin with, you still have 0 ms added load time if you just use uncompressed images. What's your concern here?

But I add the timing anyway: Freshly build, started, created boot image, booted, `save`, ` USE: binary.image.factor.compressor`, then `compress-current-image`
```
$ time ./factor.com -i=factor.image -e=""

real    0m0.281s
user    0m0.015s
sys     0m0.015s
$ time ./factor.com -i=factor.image.compressed -e=""

real    0m0.330s
user    0m0.000s
sys     0m0.031s
```

mrjbq7:
Well for me, if it was 50 millis it’s a 50% performance reduction. But in
your example it’s like 15% which if it held true to my computer is only 20
millis.

It matters for some use cases like CGI or short lived command line
programs, and I’m already frustrated we take 100 millis to mmap the image
file and I want to make it faster, not slower.

nomennescio:
> Well for me, if it was 50 millis it’s a 50% performance reduction. But in your example it’s like 15% which if it held true to my computer is only 20 millis. It matters for some use cases like CGI or short lived command line programs, and I’m already frustrated we take 100 millis to mmap the image file and I want to make it faster, not slower.

I understand your concern, but it's not related to this patch; this is about adding *optional* compression support, which *only* has a very limited cost to startup *if you use a compressed image*.

As for using `mmap`, I hardly doubt that will work for any typical code, or will work across platforms; once code starts executing, pages will get read from disk, the additional time taken will just be delayed. For me that is however a different discussion from what this patch tries to achieve. If you want to work on reduction in load time, is it not better to open a dedicated issue to that? We can exchange ideas and work on it there.

nomennescio:
Some more statistics in #2577

nomennescio:
> > I'll also be doing a compression experiment with a `load-all` image; if we can practically start using that, then run-time loading of vocabs can be prevented at all.
> 
> Got some issues with `load-all` on Windows, but nevertheless:
> 
> ```
> -rwxrwxr-x+ 1  129890744 May  8 23:57 factor.image
> -rwxrwx---+ 1  571727240 May 10 15:18 factor-all.image
> -rwxrwx---+ 1   78380047 May 10 15:28 factor-all.image.compressed
> ```
> 
> i.e. the compressed image for a load-all, is smaller than a normal uncompressed image. Its compression factor is 7.3.

With maximum compression that goes further down to:
```
-rwxrwx---+ 1   71276368 May 12 08:15 factor-all.image.compressed.22
```
a compression factor of 8.0.

nomennescio:
@mrjbq7 do you have any objections if I put this already in the master branch? We can always adapt things later.

mrjbq7:
Well, a few things need to be addressed I guess if you want to eventually merge it:

1) Where does zstd.c come from? I don't see it in Facebook's release or source files.
2) I don't generally like merge commits, and not ones with conflicts because then we have to figure out if code was appropriately de-conflicted...
3) I don't really like the ``binary.image.factor.compressor`` vocab name, maybe ``tools.image-compressor`` or something.
4) I'm not sure I like this approach, and it adds a lot of overhead -- why aren't we just compressing the whole file, instead of sections of it. what benefit does that give us?
5) what use cases do we really care about file size anyway, and have you looked to see what the bulk of the image files size is taken up by anyway? maybe there are other approaches to reducing file size
6) does the VM look for the compressed file or does this only work with ``-i=factor.image.compressed``
7) does this work in a deployed image?
8) It's a lot slower: that uncompressed, we need to speed up factor startup not slow it down

```
$ time ./factor -i=factor.image -e=""  
./factor -i=factor.image -e=""  0.10s user 0.02s system 71% cpu 0.170 total

$ time ./factor -i=factor.image.z -e=""
./factor -i=factor.image.z -e=""  0.18s user 0.04s system 90% cpu 0.235 total
```

nomennescio:
I think it's important that we're on the same page here; do you understand that compression/decompression is transparently done in the VM loader, which is determined by inspecting the header of an image file (not its name) and you *don't* get a performance penalty if you keep on using uncompressed images? And that by default Factor keeps on using uncompressed images? Which are not affected at all by this feature?

mrjbq7:
Yes, I also understand it's 1) a lot slower that uncompressed 2) not used by anyone by default 3) not tested with deploy 4) not necessarily the best approach to solving startup speed and 5) the solution to #4 might change this a lot

nomennescio:
> Yes, I also understand it's 1) a lot slower that uncompressed 2) not used by anyone by default 3) not tested with deploy 4) not necessarily the best approach to solving startup speed and 5) the solution to #4 might change this a lot

Compression and startup speed are totally unrelated issues. I'm not working on startup speed with this issue. This is about compression. Please make a separate issue out of startup speed. Compression is never going to help you with that. And improving startup speed is never going to help with compression.

I don't get it why you keep lumping these together.

As for 2. of course it's not used by anyone until we have such a feature...
As for 3. it should not impact deployment as normal images are still 100% supported

mrjbq7:
What is the point of implementing this if nothing ever uses it by default, if ``save`` doesn't save a compressed image, if ``deploy`` doesn't deploy a compressed image, if this is just a side feature with no tests and only complexity it will never be used.

The reason I lump this with startup performance is 1) it makes it slower and 2) it's entirely possible that either whole-file compression or block-based compression is faster and the latter would allow us to lazy uncompress segments, which we don't currently do and would completely change what you have done.

As an aside, I don't like the malloc in load_data_heap, if the tenured space is not initialized large enough, we should error out.  The old code path made sure it was initialized large enough, why would it not be correct for the compressed path.

If you have RAM to hold a zlib uncompressed file, you certainly have hard drive space to hold an uncompressed image.

nomennescio:
> What is the point of implementing this if nothing ever uses it by default, if `save` doesn't save a compressed image, if `deploy` doesn't deploy a compressed image, if this is just a side feature with no tests and only complexity it will never be used.

I don't get you. What do you mean by "if nothing ever uses it by default"? If you want to compress an image, you compress it, and presto, it just works. The reason I introduce the current defaults is to not disturb existing usage. I prefer such an approach; only people who explicitly want it can use it, the others don't notice anything different. Especially since people like you don't want to add 50ms startup time to Factor, which is fine. From there we can work on the other scenarios you describe. Consider it an alpha test with features to be added.

However, this feature is 2 years in the making, and I don't want to postpone it another 2 years before doing all kind of intermediate steps and constantly merging in from master. It's a relatively small change which doesn't harm any future changes.

As for tests, it's not a problem to introduce tests.

> The reason I lump this with startup performance is 1) it makes it slower and 2) it's entirely possible that either whole-file compression or block-based compression is faster and the latter would allow us to lazy uncompress segments, which we don't currently do and would completely change what you have done.

It wouldn't change a thing for making things slower, as it's optional. If we come up with a better feature in the future, that's fine by me. But currently we have no compression at all, and this feature is ready to be integrated to give compression. If you think you can make faster and better compression and uncompression, go ahead, but I won't hold my breath for that to happen. This feature won't block that any such improvements at all. To not introduce a feature because in the future we might find a a better solution goes against much of the used philosophy to first get things working, then improve upon it.

> As an aside, I don't like the malloc in load_data_heap, if the tenured space is not initialized large enough, we should error out. The old code path made sure it was initialized large enough, why would it not be correct for the compressed path.

That's fine if you want to error out, I can implement it like that. The extra space is needed to load the compressed image and have room for an uncompressed image. That's the way decompression works. The malloc is only a fallback mechanism in case the image doesn't fit in the preallocated space, but typically isn't needed.

And b.t.w, the old code path doesn't make sure it's initialized large enough, it just picks fixed sizes in which Factor images currently just happen to fit. And the exact same old code path is still in place for uncompressed images, it's 100% backwards compatible.

> 
> If you have RAM to hold a zlib uncompressed file, you certainly have hard drive space to hold an uncompressed image.

That's certainly untrue, because it depends on personal use cases.

I have limited time to work on Factor, so I don't want this to drag for another 2 years, that's a waste of my time, and I find that very frustrating. It also doesn't exactly stimulate me to keep on contributing, and I'm afraid that has happened to other contributors in the past, who stopped contributing and left the community. I'm open to accommodate to improvements or changes you would like to see for this feature, but I don't think you've given arguments to not introduce this other than basically that you wouldn't use it. I don't think this feature is going to hurt Factor in any way imaginable, nor does it impact any future activity on "better compression/better startup times", but gives a nice feature to people who want to save disk space, of which I'm one.

I'm happy to also discuss things one-on-one to see if we maybe misunderstand each other.

mrjbq7:
Because if this is only so that you can load compressed images, it's not going to get merged.

It would get merged if:

1) compressed images are things people want to use
2) compressed images are supported by the tools
3) compressed images are saved/loaded appropriately by the tools
4) compressed images are natively supported on launch (e.g., looking for ``factor.image.z`` and if not found then ``factor.image``), or perhaps that doesn't matter since the file embeds compression and a flag, but then save needs to change to save it compressed optionally
5) not just tests, but literally if this is a one-user feature then its not worth it, but if you paint a path of usefulness sure.

I don't want to slow startup, so then I think I'm unlikely to make compressed images a default, and I also don't want to miss an ``xz`` moment someday because now we have some complicated compression thing when it's not being used/tested enough.

I suppose you're saying why not merge it, then we can delete it later if we dislike it, well in that context i might be interested, but you've previously been wanting to be very careful about what public guarantees we make about things like the VM header flags and backwards compatibility, that makes me a lot less interested in trying things and then getting stuck supporting them.

mrjbq7:
It makes no sense to save disk space when you have RAM, and things like ``squashfs`` exist...

nomennescio:
> Because if this is only so that you can load compressed images, it's not going to get merged.
> 
> It would get merged if:
> 
> 1. compressed images are things people want to use
You want to vote on that or did you have something else in mind? Can't remember we ever did that for another feature.

> 2. compressed images are supported by the tools
Which tools are you missing?

> 3. compressed images are saved/loaded appropriately by the tools
Which tools?

> 4. compressed images are natively supported on launch (e.g., looking for `factor.image.z` and if not found then `factor.image`), or perhaps that doesn't matter since the file embeds compression and a flag, but then save needs to change to save it compressed optionally
This is already as implemented. The only thing is that save currently does not call compress-image, but that's trivial to add. Just tell me how you like it.

> 5. not just tests, but literally if this is a one-user feature then its not worth it, but if you paint a path of usefulness sure.
Isn't this the same as 1.?

> I don't want to slow startup, so then I think I'm unlikely to make compressed images a default, and I also don't want to miss an `xz` moment someday because now we have some complicated compression thing when it's not being used/tested enough.
I'm happy to thoroughly comment and review the source code and explain every nook and cranny.

> I suppose you're saying why not merge it, then we can delete it later if we dislike it, well in that context i might be interested, but you've previously been wanting to be very careful about what public guarantees we make about things like the VM header flags and backwards compatibility, that makes me a lot less interested in trying things and then getting stuck supporting them.
I don't see a risk here; if at some point you want to drop VM support, you just remove it; all is needed is a utility that can read, write and uncompress existing compressed images and save it uncompressed, very similar to: `binary.image.factor.compressor`.

And as we already discussed previously, at some point in time we're going to introduce the next version of image header.
Again, I'm happy to thoroughly document and review the current on image header choices and its implications. I have been *very* careful about backward and forward compatibility.

nomennescio:
> > And b.t.w, the old code path doesn't make sure it's initialized large enough, it just picks fixed sizes in which Factor images currently just happen to fit. And the exact same old code path is still in place for uncompressed images, it's 100% backwards compatible.
> 
> isn't `h->data_size` correct for uncompressed size? why wouldn't that buffer be big enough to uncompress into?

```
  p->tenured_size = std::max((h->data_size * 3) / 2, p->tenured_size);

  data_heap *d = new data_heap(&nursery,
                               p->young_size, p->aging_size, p->tenured_size);
```
This indeed overallocates the data heap, but expect it to not grow beyond that (if I remember correctly)
However, not so for the code heap
```
  if (h->code_size > p->code_size)
    fatal_error("Code heap too small to fit image", h->code_size);
```

mrjbq7:
I merged it so you don't have to worry about tracking old branches.

I reserve the right to change it later, or move things around.

If you can get the ``malloc`` removal done that would be great.

mrjbq7:
> This indeed overallocates the data heap. However, not so for the code heap

I don't understand why the inflated size would ever be larger than ``h->code_size`` and ``h->data_size``.

nomennescio:
> I merged it so you don't have to worry about tracking old branches.
> 
> I reserve the right to change it later, or move things around.

Thank you. I'm happy to support or discuss changes you like to see.

Issue #2204, opened 2019-11-03
Change factor.exe to support stdio streams in a meaningful way


nomennescio:
In a recent issue (#2200) I discussed the need for factor.com if factor.exe could also handle stdio streams. Apparently there is a need to have factor.com to be running in the Console subsystem, so let's keep it that way. However, there is still a want to have a Factor program which does not run in the Console subsystem (which puts some unwanted restrictions on a program), but in the Win32 subsystem, but with support for stdio streams inherited from the parent process (as is typical for shells running in Mintty-like terminals).

I did some experiments and found out this is already supported out of the box, except for one tiny piece of code in `windows.factor`:
```
M: windows init-stdio
    console-app?
    [ init-c-stdio ]
    [ null-reader null-writer null-writer set-stdio ] if ;
```

The `console-app?` word checks if a Console is attached to this process (which is false for Win32 processes), in which case stdio is replaced by null-writers and a null-reader.

What if this was replaced by just the following code?
```
M: windows init-stdio
    init-c-stdio ;
```

In the experiment I did I got the following results:
* factor.com Console apps keep working as is
* factor.exe GUI apps keep working as is
* factor.exe apps that try to do stdio WITHOUT the parent process having supplied valid stdio handles (e.g. by running it from a cmd.exe Console window): keep working as is, i.e. no output on stdio AND no errors
* factor.exe apps that are executed by double-clicking in Windows File Explorer: keep working as is

So far so good, no change in observable behavior. Now what about using stdio on Win32 processes?

* factor.exe apps that try to do stdio WITH the parent process having supplied valid stdio handles (e.g. by running inside a Cygwin or MSys Mintty terminal) : do output/input over stdio handles instead of supplying null-writers

Exactly what I want! What about the last case, doing stdio from the Listener UI?

* factor.exe apps that force stdio from within the Listener UI WHEN stdio is supplied: works as expected sending data towards stdio 
* factor.exe apps that force stdio from within the Listener UI WHEN NO stdio is supplied: generate error message popup "I/O error `#9` Bad file descriptor"

ONLY the last case has different behavior. And I think you can argue that it is more than acceptable; you're forcefully trying to send data over non-existent stdio streams. Which is similar what would happen if you were to use stdio on files that were corrupted or similar things.

So my proposal is to change the code in windows.factor to the above code that allows stdio streams in Win32 processes.

But what would we gain? Well at least no annoying Console windows flying around if all we want to do is simple stdio. But there's something else.

[...]

PR #2980, opened 2024-05-14
#2204 : use std handles if associated with streams

nomennescio:
#2204

### DIFF
--- a/basis/io/files/windows/windows.factor
+++ b/basis/io/files/windows/windows.factor
@@ -226,10 +226,7 @@ M: windows wait-for-fd
 
 : console-app? ( -- ? ) GetConsoleWindow >boolean ;
 
-M: windows init-stdio
-    console-app?
-    [ init-c-stdio ]
-    [ null-reader null-writer null-writer set-stdio ] if ;
+M: windows init-stdio init-c-stdio ;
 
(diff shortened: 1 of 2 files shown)


mrjbq7:
Does -2 have a constant name somewhere defined?

mrjbq7:
Should we open ``nul`` only once using ``r+`` and share the handle or is that not necessary? since they can be closed independently by factor?

mrjbq7:
shouldn't we ``fopen("nul", "r+")`` or at least ``w`` for the stdout and stderr handles?

nomennescio:
> Should we open `nul` only once using `r+` and share the handle or is that not necessary? since they can be closed independently by factor?

I think it's indeed better to have 3 independent handles

nomennescio:
> Does -2 have a constant name somewhere defined?

I can introduce a constant, but I didn't see anything mentioned in the documentation

erg:
There's no named constant for this. i asked chatgpt so it's true.

we could do this if we wanted though:

```
#define INVALID_HANDLE_FD -2
#define VALID_HANDLE(handle, mode) (_fileno(handle) == INVALID_HANDLE_FD ? fopen("nul", mode) : handle)
```

PR #2988, opened 2024-05-21
test(namespaces): inc/dec/change/+@

rgrinberg:
test inc, dec, change, and +@


core/namespaces/namespaces-tests.factor
```diff
@@ -38,3 +38,29 @@ SYMBOL: toggle-test
...
 { t } [ [ test-initialize get-global ] { at* set-at } inlined? ] unit-test
 { t } [ [ test-initialize set-global ] { at* set-at } inlined? ] unit-test
+
+{ t } [
+    f toggle-test set
+    toggle-test [ not ] change
+    toggle-test get
+] unit-test
+
+SYMBOL: counter
```
mrjbq7, line 48: probably would change this so not to shadow the ``namespaces:counter`` word.

rgrinberg: renamed

PR #2989, opened 2024-05-21
test(tr): improve tests

rgrinberg:
* test translating the last character in tr-test
* use a translation string where the dictionary lengths do not match


basis/tr/tr-tests.factor
```diff
@@ -1,8 +1,12 @@
 IN: tr.tests
 USING: tr tools.test ascii ;
 
-TR: tr-test ch>upper "ABC" "XYZ" ;
+TR: tr-test ch>upper "ABCD" "XYZ" ;
```
mrjbq7, line 4: is this the same case as the mismatch one below? now that they have different number of chars?

rgrinberg: Indeed it is. Removed now

PR #3011, opened 2024-05-27
fix(promises): only raise error once

rgrinberg:


mrjbq7:
I’m not sure yet about this, have to look at it later when I have more time On May 27, 2024, at 4:52 PM, Rudi Grinberg ***@***.***> wrote:﻿



  https://github.com/factor/factor/pull/3011

Commit Summary

  2cc4d73  fix(promises): only raise error once


File Changes  (2 files)

  
    M
    basis/promises/promises-tests.factor
    (3)
  
  
    M
    basis/promises/promises.factor
    (22)
  


Patch Links:

  https://github.com/factor/factor/pull/3011.patch
  https://github.com/factor/factor/pull/3011.diff

mrjbq7:
the help lint error maybe related to change in definition:

```
==== FAILING LINT

==== resource:basis/lists/lazy/lazy-docs.factor

resource:basis/lists/lazy/lazy-docs.factor: 148

Asset: T{ link f lmerge }

Assertion failed
got    "Fall-through in case\nobject t"
expect "{ 1 4 2 5 3 6 }"
```

rgrinberg:
I think I fixed it. The fix is good on its own even without this PR I think.

mrjbq7:
I didn't pull this whole thing, but this breaks with the lists.lazy patch:

```
IN: scratchpad [ 1 ] [ 2 ] [ 3 ] 3lazy-list list>array .
{ 1 2 3 }
```

mrjbq7:
Maybe it could be

```factor
: lazy-cons ( car cdr -- promise )
    [ <promise> ] bi@ '[ _ _ <lazy-cons-state> ] <promise> ;
```

mrjbq7:
It also means something different, i guess in the pre-patch version it creates the value the promise would return already realized, in the new version it calls the quot to generate it on demand

rgrinberg:
Also pushed a `<value>` constructor from a realized value. This ends up saving us some pointless work creating/evaluating a quote.

mrjbq7:
This change would prevent any kind of *retry logic*, for example:

```
IN: scratchpad SYMBOL: foo
IN: scratchpad [ foo counter 4 > [ "not yet" throw ] unless 42 ] <promise>

IN: scratchpad force
not yet

IN: scratchpad force
not yet

IN: scratchpad force
not yet

IN: scratchpad force
not yet

IN: scratchpad force

--- Data stack:
42
```

What do you think?

rgrinberg:
It might be worth pointing out another instance where I found this confusing. The name of the vocab is `promises`, however, the implementation is very much how lazy values are implemented in other languages. Even facto itself uses this vocab to implement lazy lists. Since in the other programming languages I'm familiar with, a lazy value represents a value produced by running a computation. For example, in OCaml:

```
utop # let x = lazy (print_endline "foo"; failwith "foo");;
val x : 'a lazy_t = <lazy>
# Lazy.force x;;
foo
Exception: Failure "foo".
# Lazy.force x;;
Exception: Failure "foo".
```

So I wouldn't use a lazy computation to represent re-tries. If I needed re-tries, I would probably create a new lazy value for every try. Or, build something custom that supports re-tries.

However, given that this vocab's name is indeed `promises`, maybe the authors had something else in mind? My own opinion is towards the following changes:

1. Remember failures like this PR does
2. Rename the vocab to `lazy`
3. Redirect people that need actual promises to `concurrency.promises` as that already contains an implementation of what promises are traditionally and those indeed do not allow for saving failures.

Finally, I haven't yet verified this, but there's another issue with this vocab. Because the quotation generating the value is never erased after `force`, we end up always leaking whatever the quotation captures. My recommendation is to set the quotation to `f` after `force`. If we don't remember failures, we would need to keep the quotation until the first success. That also seems a bit suspicious to me.

mrjbq7:
Well, that's compelling. I have not liked how ``promise`` shadows both vocabularies.

Also, we currently only use the promises vocabulary in ``lists.lazy``, so perhaps this just moves to an implementation detail of that vocabulary...? Or do you think it's useful in general (currently not generally used).

rgrinberg:
I think it's useful on its own. Also, data structures other than lists can be made lazy.

PR #3017, opened 2024-05-30
test(msgpack): demonstrate eof handling

rgrinberg:
Reading from an input stream raises an error. This makes it difficult to read all msgpack objects from a stream until exhaustion.

Given that `f` is a valid msgpack value, it is annoying that we can't return false.

I suppose that one simple work around would be to take a quotation that will be called if the eof is reached. Another work around would be to return 2 values on the stack.


mrjbq7:
Fixed that issue in 7e40be79cd9e86844f439e18cc3e00a09eb39a61.

Added ``?read-msgpack`` and ``read-msgpacks``

PR #3021, opened 2024-05-31
feature(msgpack.rpc): introduce vocab for msgpack rpc

rgrinberg:
Implemented according to this spec https://github.com/msgpack-rpc/msgpack-rpc/blob/master/spec.md

I haven't yet written the docs because I'm still working on the API. It's at a point where it's usable though.

### DIFF
--- a/basis/msgpack/rpc/rpc.factor
+++ b/basis/msgpack/rpc/rpc.factor
@@ -1,0 +1,169 @@
+USING: msgpack kernel accessors assocs arrays sequences
+combinators make io threads destructors concurrency.locks
+io.streams.duplex math.parser ;
+
+IN: msgpack.rpc
+
+TUPLE: request msgid method params ;
+
+TUPLE: response msgid error result ;
+
+TUPLE: notification method params ;
+
+TUPLE: session active? incoming-requests outgoing-requests notification-callback request-callback ;
+
+M: request write-msgpack
+    0 swap [ msgid>> ] [ method>> ] [ params>> ]
+    tri 4array write-msgpack ;
+
+M: response write-msgpack
+    1 swap [ msgid>> ] [ error>> ] [ result>> ]
+    tri 4array write-msgpack ;
+
+M: notification write-msgpack
+    2 swap [ method>> ] [ params>> ]
+    bi 3array write-msgpack ;
+
+GENERIC: feed-packet ( session packet -- )
+GENERIC: stop ( session -- )
+
+! TODO close active requests
+M: session stop ( session -- )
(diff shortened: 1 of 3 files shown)


basis/msgpack/rpc/rpc.factor
```diff
@@ -0,0 +1,161 @@
...
+TUPLE: request msgid method params ;
+
+TUPLE: response msgid error result ;
+
+TUPLE: notification method params ;
+
+TUPLE: session active? incoming-requests outgoing-requests notification-callback request-callback ;
+
+M: request write-msgpack
+    [ msgid>> ] [ method>> ] [ params>> ]
```
mrjbq7, line 16: somewhat inefficient, this is faster:

```factor
M: request write-msgpack
     0 swap [ msgid>> ] [ method>> ] [ params>> ] tri 4array ;
```

basis/msgpack/rpc/rpc.factor
```diff
@@ -0,0 +1,161 @@
...
+GENERIC: stop ( session -- )
+
+! TODO close active requests
+M: session stop ( session -- )
+    f swap active?<< ;
+
+: <session> ( -- session )
+    t H{ } clone H{ } clone [ drop ] [ drop ] session boa ;
+
+: <request> ( msgid method params -- request )
```
mrjbq7, line 40: If you aren't going to do anything different, these are auto-generated using constructor syntax:

```
C: <request> request
```

basis/msgpack/rpc/rpc.factor
```diff
@@ -0,0 +1,161 @@
...
+
+: <response-error> ( id error -- response )
+    +msgpack-nil+ response boa ;
+
+M: notification feed-packet
+    swap notification-callback>> call( notification -- ) ;
+
+ERROR: invalid-packet packet ;
+
+: check-array ( obj -- obj )
```
mrjbq7, line 57: We also have ``array check-instance``

mrjbq7:
Pretty cool! Made a few small comments.

I would probably use ``<PRIVATE`` to keep some of the common words like ``start``, ``stop``, and ``recv-loop`` from shadowing other vocabs.

mrjbq7:
Either way, let me know!

basis/msgpack/rpc/rpc.factor
rgrinberg: That gives a worse error though

rgrinberg:
I made a bunch of stop private (but not start/stop). Should be ready from my end

PR #3025, opened 2024-06-01
completion: allow space to be part of search, and use TAB for completion

nomennescio:
Recent investigation into `CTRL-R` shows that SPACE will cancel the search, whereas it can be part of a command in command history. It's better to just use `TAB` for completion, and let `SPACE` be a search character.

### DIFF
--- a/basis/ui/tools/listener/completion/completion.factor
+++ b/basis/ui/tools/listener/completion/completion.factor
@@ -166,7 +166,6 @@ GENERIC#: accept-completion-hook 1 ( item popup -- )
 
 completion-popup H{
     { T{ key-down f f "TAB" } [ table>> row-action ] }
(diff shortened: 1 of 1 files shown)


mrjbq7:
we don't want this on every other completion

nomennescio:
> we don't want this on every other completion

So what's your suggestion to integrate this?

mrjbq7:
if SPACE is changed to be part of the input then it's gonna be annoying for all the other tab completions

nomennescio:
> if SPACE is changed to be part of the input then it's gonna be annoying for all the other tab completions

So what's your suggestion to be able to search history for embedded spaces. This code actually works.

mrjbq7:
Fixed in 13f123f96ddc4881171da6638bfeb0a66a7cfcad.

PR #3031, opened 2024-06-03
Allow non-singleton as first variant member

Kacarott:
Previously, variants without a singleton class as first member which had recursive slots, such as:
```
VARIANT: tree
    node { val }
    branch { { left tree } { right tree } }
    ;
```
would throw an error, due to `f` not being a valid initial value for the `tree` slot in `branch`. Now it will find a reasonable initial value (here it would be `T{ node f f }`) to use, allowing this kind of construct.

The tests use `eval` because I was not sure how best to write a test for a potential parse-time error. The added test fails for the previous code, and passes for the new code.

### DIFF
--- a/extra/variants/variants-tests.factor
+++ b/extra/variants/variants-tests.factor
@@ -1,6 +1,6 @@
 ! Copyright (C) 2009 Joe Groff.
 ! See https://factorcode.org/license.txt for BSD license.
-USING: kernel math tools.test variants slots ;
+USING: eval kernel math tools.test variants slots ;
 IN: variants.tests
 
 VARIANT: list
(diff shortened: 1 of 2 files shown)


mrjbq7:
I'm not sure, what about a direction where we think about adding this:

```
M: singleton-class initial-value* t ;
```

Kacarott:
I don't really see how that helps? The issue was that before the first variant entry was being checked whether it was a `word` (which to my understanding it *always* will be?) and therefore always returning the first entry and `t` as the initial value. But this doesnt work for non-singleton classes, it tries to set the initial value as the *class* itself. It only works for singleton words because they are both the class and the object. The code change just means that non-singleton classes properly get recursed upon to `initial-value*` again, which seems to me what the original idea was in the first place.

mrjbq7:
I guess what I'm wondering is maybe it should always call ``initial-value*``?

mrjbq7:
Is it weird to return the initial value of the first element of a union class?

Kacarott:
Ahhh, I get you now. Yeah that would work fine too, and probably more elegant.

Kacarott:
> Is it weird to return the initial value of the first element of a union class?

For variants specifically, it is [documented](https://docs.factorcode.org/content/word-variant-class%2Cvariants.html) to behave that way. Otherwise it kind of is weird? This still sort of relates to my questions about having classes with no initial value.

mrjbq7:
I still find it weird that the first class member provides the initial value

PR #3033, opened 2024-06-04
Improve `prettyprint.config` documentation; add docs to `pack`

ToxicFrog:
I'm not even using `pack` myself but I stumbled across it in `terminfo` and decided it could use a manual, especially since the really important part (the format table) is tucked away in pack.private.

### DIFF
--- a/basis/pack/pack-docs.factor
+++ b/basis/pack/pack-docs.factor
@@ -1,0 +1,81 @@
+USING: help.markup help.syntax math io ;
+IN: pack
+
+HELP: packed-length
+{ $values { "str" "a format string" } { "n" integer } }
+{ $description "Computes the packed size, in bytes, of the structure described by the given format string (without actually packing or unpacking any data)." } ;
+
+HELP: pack-native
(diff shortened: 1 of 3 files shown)


mrjbq7:
We haven't started converting documentation over to it yet, but we wrote an [easy help](https://re.factorcode.org/2023/06/easy-help.html) that makes this:

```
HELP: read-packed-le
{ $values { "str" "a format string" } { "seq" "a sequence of field values" } }
{ $description "Reads packed data from " { $link input-stream } " according to the field formats described by " { $snippet "str" } " and outputs a sequence containing the unpacked values. Packed data is assumed to be little-endian." } ;
```

Be able to be written more simply as:

```
HELP: read-packed-le
{ $values
    str: "a format string"
    seq: { $sequence "field values" }
}
{ $description
    Reads packed data from \ input-stream according to the field formats
    described by { $snippet "str" } and outputs a sequence containing the
    unpacked values. Packed data is assumed to be little-endian.
} ;
```

Note: it doesn't require all the string literals now

ToxicFrog:
The whole time I was working on this I was thinking something like that would be really nice, good to know I've been anticipated in that :)

PR #3050, opened 2024-06-29
build: Apply debug flags to C++ files

jryans:
I noticed while attempting some debugging that the debug compiler flags were only set for C files. This adds them to C++ files as well.


mrjbq7:
Ahh, lost in the Great Build Flag Reordering. Thanks!

PR #3052, opened 2024-06-30
Fix Clipboard on Windows

alex-ilin:
Hello!

I have noticed that the `paste` word does dummy `GlobalLock`/`GlobalUnlock` without actually using the locked handle. Then I also noticed there is a race condition between check and use.

So, here are the fixes.


mrjbq7:
Thanks!

PR #3053, opened 2024-07-01
vm: Update OS string for macOS

jryans:
As a final step for the Mac OS X to macOS rename, this updates the VM OS string to the new value of `macos`. This had to be done separately from the main change so that a boot image could first be prepared that supports both OS strings.

### DIFF
--- a/vm/os-macos.hpp
+++ b/vm/os-macos.hpp
@@ -1,7 +1,7 @@
 namespace factor {
 
 #define VM_C_API extern "C" __attribute__((visibility("default")))
-#define FACTOR_OS_STRING "macosx"
+#define FACTOR_OS_STRING "macos"
 
 void early_init();
 


mrjbq7:
Let’s not merge this quite yet, as new binaries won’t be compatible with old images. 

But let’s keep this PR for maybe a couple weeks from now when we have clean macOS builds again.  The build farm is having trouble uploading and I’m on a trip. Will fix in a week.

jryans:
@mrjbq7 Perhaps this can be merged now...?

mrjbq7:
The only people that would be affected by merging this would be anyone that has a "git pull" then "refresh-all" style of updating to latest code... anyone that did a bootstrap would get a consistent image with the binary... hmmmm, trying to decide if its overly conservative to delay merging.

PR #3055, opened 2024-07-06
help.handbook: Add Stacks article with all the stacks used internally

jryans:
As someone new to Factor, I was initially a bit confused by the various additional stacks used by language internals. As part of learning more about Factor, I've tried to understand the purpose of all of the stacks. I think that would be useful for other people too, so I've added it to a new implementation article in the handbook.

Please read carefully and double-check what I've written here, as I am still quite new to Factor, and I might have something mixed up. 😇


mrjbq7:
Cool, thanks!

PR #3059, opened 2024-07-13
zstd: Use dynamic library for decompression

jryans:
Recent work added optional support for compressed images, and as part of this, `zstd` sources were added directly into the Factor VM binary to enable decompression. While the feature of compressed images is surely useful for various deployment scenarios, adding the `zstd` code directly into the Factor VM increased the VM's binary size substantially (0.3 MB -> 1.0 MB), which is a high cost for an optional feature.

This changes to use the `zstd` dynamic library for decompression. (Compression was already using the dynamic library.) The `zstd` sources have also been removed, as they are no longer needed in tree. With this change the Factor VM binary returns to its smaller size as before.


mrjbq7:
:eyes:

@nomennescio

mrjbq7:
I wonder if we should re-use / merge these ``native_dlopen`` words with the existing ``ffi_dlopen`` words...

jryans:
> I wonder if we should re-use / merge these `native_dlopen` words with the existing `ffi_dlopen` words...

There's definitely some overlap... Actually, I already tried a round of calling the `native_*` functions from inside `ffi_*`, but it felt a bit awkward with the amount of casting inputs and outputs, so I just went for separate functions. If you'd like to see a version where `ffi_*` calls `native_*`, I can reconstruct that. Or if there's a different kind of reuse you have in mind, just let me know what you're thinking of, and I can give it a try.

mrjbq7:
Thanks! Let's give this a try for a bit!

PR #3061, opened 2024-07-18
Add an implementation of RFC2047 strings

theAlexes:
This implements RFC2047 Section 2: https://datatracker.ietf.org/doc/html/rfc2047#section-2


mrjbq7:
Thanks!!

PR #3062, opened 2024-07-27
Bitmap image scaling

inivekin:
Specifically enables multi-texture scaling and adds a gadget that makes the functionality easy to try via something like:

```
USE: images.viewer.scaling "/some/path/to/image.png" load-image <scaling-image-gadget> "scaling-image-test" open-window
```

### DIFF
--- a/basis/opengl/textures/textures-tests.factor
+++ b/basis/opengl/textures/textures-tests.factor
@@ -37,3 +37,33 @@ ${ GL_RGB9_E5 GL_RGB GL_UNSIGNED_INT_5_9_9_9_REV }
 
 ${ GL_R11F_G11F_B10F GL_RGB GL_UNSIGNED_INT_10F_11F_11F_REV }
 [ BGR float-11-11-10-components (image-format) ] unit-test
+
+! multi-textures are made of a grid of single-textures that, for this example
+! could have dims like the following:
+!  { { 512 512 } { 256 512 } }
+!  { { 512 256 } { 256 256 } }
+! with a total of 768x768 pixels, 
+! to scale each axis of the above image by half (384x384), each single-texture
+! will have a common factor to achieve this
+{
+ { 256 256 }
+}
+[ { 384 384 } { { { 1 1   } { 1/2 1   } }
+                { { 1 1/2 } { 1/2 1/2 } } } normalize-scaling-dims ] unit-test
+! the adapted scaling factor can be multiplied with the normalisation matrix
+{
+  { { { 256 256 } { 128 256 } }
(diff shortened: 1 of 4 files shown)


mrjbq7:
If this is working, how should it work?

Tried this:

```factor
USE: images.viewer.scaling
"/path/to/image.png" load-image <scaling-image-gadget> "scaling-image-test" open-window
```

When I resize the window the image doesn't resize? Is it supposed to?

mrjbq7:
Regarding the memoize, i would just not memoize it for now, make it a normal function unless there's a performance bottleneck immediately...

inivekin:
> If this is working, how should it work?
> 
> Tried this:
> 
> ```factor
> USE: images.viewer.scaling
> "/path/to/image.png" load-image <scaling-image-gadget> "scaling-image-test" open-window
> ```
> 
> When I resize the window the image doesn't resize? Is it supposed to?

Currently the pull only resizes with a middleclick drag on the image (dragging right will increase, left will decrease).

I temporarily removed the auto-resizing because it still had some dodgy behaviour to it under more complicated layouts but will try add it back in sometime in the next few days, before a merge.
I realise now that I think I need an arrow model to get updated within a draw-gadget* call for the dimensions to get passed to scrollable parents like viewports correctly.

Will remove the memo, what you say was what I was suspecting should be the case, thanks.

inivekin:
I've added the auto-scaling gadget with some quick associated docs, if curious.

I finally got it to handle propagating the changing gadget dimensions up, I based the autoscaling on assigning a `fill` percentage of the parent viewport/root (NOT the immediate parent gadget) so I do some things relying on the `root?` slot that I don't see often in the codebase - hopefully that isn't problematic.

A few simple things I still want to do before merge (I should get this done reasonably soon, ideally tomorrow):
- there may still a subtle issue with how certain `fill` behaviours work with multi-texture images that don't entirely line up with documented effects
- reducing code duplication, more factoring, fixing the USING:
- more and nicer to read docs/help.  

```
USE: images.viewer.scaling
"/path/to/image.png" load-image <autoscaling-image-gadget> 1 >>fill "autoscaling-image-test" open-window
```

mrjbq7:
Awesome! I'm happy to merge when you are ready, no worries!

inivekin:
Ok, added the fixes and better explanation of functionality. Happy for you to merge when you'd like.
Thanks for your input and if you think I should change anything else before merge, let me know.

mrjbq7:
Sounds great! I'll merge and play with it.

Another thing is maybe to make help easier to write, we have the new "easy help" now:

https://re.factorcode.org/2023/06/easy-help.html

PR #3075, opened 2024-10-05
raylib: fix log functions

greenfork:
The `TraceLog` function is actually variadic but it just concatenates the text to output, not a big deal. Let me know if it needs to be variadic, and some pointer to how write it, I don't know how to do it


erg:
The C FFI doesn't support variadic args yet, so let's go with what you have. Merged. Thanks!

Issue #3077, opened 2024-10-08
Deploy with resources fails on Linux


greenfork:
Reproduction:

```
cd factor-path/work
git clone https://github.com/JosephOziel/bitguessr.git
factor  # run listener
```

In listener:
```
"bitguessr" deploy
```

Outputs:

```
1: Note:
Added "tools.deploy" vocabulary to search path
Loading resource:work/bitguessr/bitguessr.factor
Loading resource:extra/raylib/raylib.factor
Loading resource:extra/raylib/util/util.factor
Loading resource:extra/raylib/raylib-docs.factor
{
    "bitguessr/bitguessr"
    "-pic=0"
    "-i=/home/grfork/.cache/factor/staging.math-threads-compiler-i..."
    "-vocab-manifest-out=/tmp/factor-temp/vocab-manifest-bitguessr"
    "-deploy-vocab=bitguessr"
    "-deploy-config=/tmp/factor-temp/deploy-config-bitguessr"
    "-output-image=/tmp/factor-temp/bitguessr.image"
    "-resource-path=/home/grfork/reps/factor/"
    "-run=tools.deploy.shaker"
}
Loading resource:basis/tools/deploy/shaker/shaker.factor
Loading resource:basis/tools/deploy/config/config.factor
Deploying bitguessr...
Loading /tmp/factor-temp/deploy-config-bitguessr
Loading resource:basis/debugger/debugger.factor
Loading resource:basis/fixups/fixups.factor
Loading resource:basis/io/styles/styles.factor
Loading resource:basis/colors/colors.factor
Loading resource:basis/ascii/ascii.factor
Loading resource:basis/delegate/protocols/protocols.factor
Loading resource:basis/strings/tables/tables.factor
Loading resource:basis/prettyprint/prettyprint.factor
Loading resource:basis/prettyprint/config/config.factor
Loading resource:basis/prettyprint/sections/sections.factor
Loading resource:basis/vocabs/prettyprint/prettyprint.factor
Loading resource:basis/prettyprint/backend/backend.factor
Loading resource:basis/lists/lists.factor
Loading resource:basis/prettyprint/stylesheet/stylesheet.factor
Loading resource:basis/hashtables/wrapped/prettyprint/prettyprint.factor
Loading resource:basis/hashtables/identity/prettyprint/prettyprint.factor
Loading resource:basis/locals/definitions/definitions.factor
Loading resource:basis/see/see.factor
Loading resource:basis/locals/prettyprint/prettyprint.factor
[...]

Issue #678, opened 2012-09-16
Deployed vocabularies can't have resources on linux


jonenst:
The executable name inside the deployed vocab is the vocabulary name. If the vocabulary or one of its child has a `resource.txt` file, then the name of the executable and the directory where the resource is copied are the same, which makes `make-directories` fail.

Don't know if it's best to just change the executable name to be different than the vocabulary name (perhaps suffix `.exe` or something like in window), or to allow to have a binary with the same name as the vocabulary, and move the resources away.

Note that a workaround is make the vocab with the resources not a child of the deployed vocab: ie instead of A and A.theme, use A and A-theme.

mrjbq7:
Fixed using ``.out`` extension and the strategy suggested by @jonenst.

Thanks!

PR #3080, opened 2024-10-12
tools.deploy.unix: add .out extension to deployed binaries

greenfork:
This is a workaround for #3077, maybe could be even considered as a solution.

As per suggestion from @mrjbq7 in [discord](https://discord.com/channels/780615045771821076/786055699271909428/1294306017286426666)


mrjbq7:
Thanks!!

samueltardieu:
Isn't #678 fixed by this as well?

greenfork:
To me the answer seems like yes

Issue #1099, opened 2014-07-08
Some issues with alien.libraries.finder


mrjbq7:
``` python
>>> from ctypes.util import find_library
>>> find_library("SDL")
"libSDL-1.2.so.0"
>>> find_library("SDL-1")
"libSDL-1.2.so.0"
>>> find_library("SDL-1.2")
"libSDL-1.2.so.0"
```

The first one ("SDL") doesn't work on Linux.

The first and second ones don't work on Mac OS X.

erg:
Does ``ldconfig`` give us the correct ordering for libraries? We are just filtering a list of matches based on name prefix and then taking the first library from that list. It seems like this is wrong because the ordering could be whatever order. For instance, on Linux, for ``"GL" find-library``, it finds:

erg:
Ok, here's the way they're ordered:

```
erg@erglinux:~$ ldconfig -v | grep "^/"
/sbin/ldconfig.real: Path `/lib/x86_64-linux-gnu' given more than once
/sbin/ldconfig.real: Path `/usr/lib/x86_64-linux-gnu' given more than once
/sbin/ldconfig.real: /lib/x86_64-linux-gnu/ld-2.19.so is the dynamic linker, ignoring

/usr/lib/x86_64-linux-gnu/libfakeroot:
/usr/local/lib:
/lib/x86_64-linux-gnu:
/usr/lib/x86_64-linux-gnu:
/sbin/ldconfig.real: /lib32/ld-2.19.so is the dynamic linker, ignoring

/usr/lib/x86_64-linux-gnu/mesa-egl:
/usr/lib/fglrx:
/usr/lib32/fglrx:
/usr/lib/x86_64-linux-gnu/mir/client8driver/mesa:
/lib32:
/usr/lib32:
/lib:
/usr/lib:
```

``/usr/lib/fglrx`` is below ``/usr/lib/x86_64-linux-gnu/``, so mesa gets loaded unless i run ``LD_LIBRARY_PATH=/usr/lib/fglrx ./factor``.

erg:
Filed a bug report with ubuntu.

https://bugs.launchpad.net/ubuntu/+source/fglrx-installer/+bug/1462515

erg:
Ok, it looks like ``ldconfig -p`` just reads the ``ld.so.cache`` file and spits out names in the order they would look up. However, this order is different from ``ldconfig -v``.

erg:
[...]

PR #3081, opened 2024-10-12
alient.libraries.finder.linux: don't prepend "lib" to libraries

greenfork:
In bindings libraries are specified with a full filename such as libraylib.so. Prepending "lib" to the beginning doesn't make sense in this case.

This is a breaking change.

[Discussion in Discord](https://discord.com/channels/780615045771821076/785087023022211082/1294684611049291874).

### DIFF
--- a/basis/alien/libraries/finder/linux/linux.factor
+++ b/basis/alien/libraries/finder/linux/linux.factor
@@ -41,11 +41,11 @@ CONSTANT: mach-map {
     { [ name-matches? ] [ arch-matches? ] } 2&& ;
 
 : find-ldconfig ( name -- path/f )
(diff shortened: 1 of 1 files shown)


greenfork:
TODO:
- Provide multiple name resolution strategies, similar to [MacOS](https://github.com/factor/factor/blob/348776e85a6cf291dab74302369f1e409e18dc77/basis/alien/libraries/finder/macos/macos.factor#L140)
- `libraylib.so` must find `/usr/lib/libraylib.so` if it exists, currently finds `/usr/lib/libraylib.so.450`. Related https://github.com/factor/factor/issues/1099

PR #3088, opened 2024-10-19
Try to improve continuations docs

jonenst:
Try to be consistent:
- resume a continuation
- continue the execution

Add explanation of the continuation lifetime

Add explanation around how continations can be used

ifcc, callcc0, callcc1, Add docs for what happens when the capture quotation returns, not only for when the continuation is resumed (it is actually the normal use case ! when using exception and no exception occurs !)

fix the stack effect of ifcc (inline word so no effect)
fix the docs of the stack effect of ifcc 

fix the docs of with-return and return, reify -> resume

### DIFF
--- a/core/continuations/continuations-docs.factor
+++ b/core/continuations/continuations-docs.factor
@@ -33,7 +33,7 @@ ARTICLE: "errors-anti-examples" "Common error handling pitfalls"
 $nl
 "In most other cases, " { $link cleanup } " should be used instead to handle an error and rethrow it automatically."
 { $heading "Anti-pattern #3: Dropping and rethrowing" }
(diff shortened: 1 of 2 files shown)


jonenst:
I noticed the stack effect were not variadic ( I think that's how we call the ...a ...b stuff in effects, i'm not sure it's been a long time...) for ifcc, callcc0, callcc1 but they could be. And for try/recover they have been written as variadic so maybe add that ?

jonenst:
Also the nondeterministic pattern "amb" that I mentionned in the examples is actually implemented in extra/backtrack:amb-lazy but I couldn't link to it from documentation in core from what I remember, is there a way to do it now ?

jonenst:
Let me know what you think, and don't hesitate to reword or add or remove anything needed

mrjbq7:
That’s great, thanks!

PR #3089, opened 2024-10-19
add row polymorphism declarations to continations, fix some bad ones

jonenst:
can someone check that the new stack effects are correct ?


jonenst:
don't merge there's a mistake in `cleanup` it doesn't get the error. Also it seems like row variables can't express the min height of mismatching branches (with exception or without exception) of cleanup/finally, so need to decide on what to do

jonenst:
never finished

PR #3093, opened 2024-11-14
command-line.startup: document -q flag

ToxicFrog:


mrjbq7:
We probably need a better way to add arguments that apply to the vocab
being run.

That “-q” isn’t an argument for factor *unless* it’s running the listener.
I almost made the patch you did also. Hmm

ToxicFrog:
Hmm. So it is.

That's not great; command line parsing is really ad-hoc, `command-line` just takes all the flags it sees and dumps them onto the namestack. There's no predeclaration of flags, no separation of flags from factor proper and the vocabulary being run, and no error checking -- `factor -wharg -blargh` is a-ok because nothing checks for those. And there's no exposure of flags understood by the `-run` vocab.

There's a much more comprehensive command line parsing library in `command-line.parser`, but that's in extra so the startup code can't use it. :/

mrjbq7:
Yes, but if we like command-line.parser we could promote it.

I wrote that to be like arg parse in Python, but there are other models
like getopt and opt parse and click and etc

mrjbq7:
Another idea is a simple per-vocab “add stuff to the command line help” protocol. The command-line.parser could respect it as well.

mrjbq7:
Let's not let the perfect be the enemy of the good -- merging for now. Thanks!

mrjbq7:
We haven't switched to ``main`` branch yet as primary development, it's still just synced from ``master``.  So, noticed this was merged into main and cherry-picked it into master.

PR #3094, opened 2024-11-14
terminfo and ANSI-stream improvements

ToxicFrog:
The `terminfo` vocabulary gains convenience words for querying support for specific capabilities: SGR, dim, and 8/256/rgb-colour support.

The `io.streams.escape-codes` vocabulary gets a few extra tests, documentation, and an overview article on formatted TTY output.

The `io.streams.ansi` vocabulary gets documentation, tests, and some significant refactoring, which brings with it support for the dim attribute to get an extra 7 colours on ttys that support it.


mrjbq7:
Awesome!

PR #3095, opened 2024-11-14
readline-listener: use history and autodetect colour support

ToxicFrog:
This is based on #3094 as it depends on features added to the terminfo vocabulary.

This adds support for persistent history using the same file as the GUI listener, including support for `$FACTOR_HISTORY`. It also uses terminfo to autodetect terminal colour support, and enables it if available. If the user has not overridden the colour scheme, it defaults to dark.


mrjbq7:
I merged the first few commits in the other PR, so cherry-picked the last two ``readline-listener`` commits!

Thanks!!

PR #3100, opened 2024-11-22
gtk4: draft implementation

knottio:
Initial draft of a `gtk4` vocab.

## Summary
- Added `cairo_rectangle_int_t` and `cairo_region_t` types to `cairo.ffi`, as they're required for parsing some GIR.
- Added `cairo.gobject.ffi` and `freetype2.gobject.ffi` vocab which processes their respective GIR files to load the `type-info`s required by some of the other GIR.
- Added new GIR files required by `gtk4`, and updated old ones.
- Removed some manually declared definitions from `glib.ffi` since they're now generated by the new GIR.
- Updated `gobject-introspection` to be able to process new GIR files.
- Added the `gtk4`, along with necessary dependents, `gsk`, `graphene`, and the aforementioned `.gobject.ffi`s.
- Added a simple GTK4 + OpenGL sample in `extras/gtk4-samples/opengl` to test the bare minimum.

## `.gobject.ffi`
Some GIR files aren't meant to be used for actually generating bindings for the libraries they describe, only for supporting GIR files that are meant for generating bindings. In these cases, I added a `.gobject.ffi` vocab to the root vocab for the library. These vocabs process the GIR files to generate the `type-info` required for processing other GIR files. This might need altered, since it currently still generates definitions. I'm unclear on the precise implications here.

## `gobject-introspection`
I added a `SKIP-DEFINITIONS:` parsing word (similar to `IMPLEMENT-STRUCTS:`) which takes a list of definitions to be skipped when parsing the GIR. This helps with workarounds for definitions which `gobject-introspection` fails to parse. A manual definition can then be written. I used it for `GObject.G_TYPE_FLAG_RESERVED_ID_BIT`, which is a constant of type `GType` and value `1`. `GType` does typedef to `gulong`, so it's probably feasible to get it to parse correctly, but I wasn't sure how to best approach that, so for now, it's using a workaround.

I moved `def-consts` to happen later in `def-namespace`, since otherwise some definitions haven't defined/deferred the necessary type yet.

`xml>array-type` was updated to handle nested arrays.

`fix-conts` is now named `fix-consts` and now uses `identifier-prefixes` instead of `symbol-prefixes`, resolving some ambiguity between the older and newer GIR files.

`load-functions` now filters out xml elements with attribute `moved-to`, since they'll otherwise result in re-definitions.

Added some more standard-types, some of which are needed, others weren't needed, but they're all assumed by `GNOME/gobject-introspection`.

## "GTK 2/3 symbols detected"
A standard bootstrap (on Linux) will use `ui.backend.gtk2`, which results in things I'm unclear on, but the end result is that attempting to run `gtk4-samples.opengl` will fail with an error message: `GTK 2/3 symbols detected. Using GTK 2/3 and GTK 4 in the same process is not supported`. I'm currently using `./factor -i=boot.unix-x86.64.image -exclude="tools help handbook ui ui.tools"` to build the image for testing `gtk4`. Haven't check exactly which components bring in `gtk2`, but just `-exclude="ui ui.tools` wasn't enough.

## To-Do

- [ ]  Add the appropriate dynamic libraries for `macos` and `windows`.
- [ ] Compare the exact definitions generated before and after the updated GIR files to check for anything important that might be missing.
- [ ] Decide how to best handle the GTK2 & GTK4 symbols conflict.


basis/cairo/gobject/ffi/ffi.factor
```diff
@@ -0,0 +1,10 @@
+! Copyright (C) 2024 knottio.
+! See https://factorcode.org/license.txt for BSD license.
+USING: alien.syntax gobject-introspection vocabs ;
+IN: cairo.gobject.ffi
+
+<< "cairo.ffi" require >>
+
+LIBRARY: cairo-gobject
+
+GIR: vocab:gir/cairo-1.0.gir
```
mrjbq7, line 10: it looks like we manually define the cairo.ffi, are there words here that we should manually define instead of using the GIR?  Alternatively, should we only use the GIR?

mrjbq7:
I'm going to look through here, but this is great! I should be able to start here for a new gtk4 backend and then make that the default, so we wouldn't have both symbols in the same namespace due to loading both old and new gtk dlls.

basis/cairo/gobject/ffi/ffi.factor
knottio: Oh, I forgot that when I wrote the `.gobject.ffi` vocabs, I intended on extending `gobject-introspection` to check for the `foreign="1"` attribute and handle them accordingly. But now that I looked at it again, that wouldn't work quite right, since some of the cairo structs needed by GDK4 are fully defined in the GIR rather than setting `foreign="1"`. So the current approach--as used by `gdk2`--of writing out the `FOREIGN-*-TYPE:` lines seems just fine. I thought there was a lot more to the cairo GIR than there is. Looks like only a few more types than what's defined in `gdk2.ffi`. I'll switch it over to that approach; that'll get rid of the double definitions I was seeing.

knottio:
Changed `cairo.gobject.ffi` back to the original approach used by `gdk2`. Still left it in the `.gobject.ffi` vocab, instead of moving it to `gdk4` the way `gdk2` is. Could potentially be used for both `gdk2` and `gdk4`. Let me know what's best.

Additionally I removed the `basis/freetype2` parts I added. It wasn't actually needed by any of the other GIR files.

knottio:
I changed `fix-consts` back to using `symbol-prefixes>>`, because using `identifier-prefixes>>` is wrong for some of the older GIR files. The newer GIR files all have `c:type` set to their C identifier, so now `fix-consts` uses that if it exists, and falls back to the prior implementation if it doesn't.

This also ends up changing some `glib.ffi` constants (`G_{MAJOR,MINOR,MICRO}_VERSION`, `G_SIZEOF_*`, `G_SYSDEF_*`) to start with `GLIB_` instead of `G_` which is how they're defined in GLib. The old GIR didn't have sufficient information to restore the proper prefix, but the new one does.

So far, I've only checked `glib.ffi` for missing words. Looks like aside from the aforementioned changes, it's just a few deprecated symbols that were removed from the new GIR.

I'll check the others for missing words soon.

knottio:
My `skip-definitions` implementation currently uses the `name` attribute to specify what should be skipped, which isn't the name typically used in Factor. This already tripped me up and caused double definitions of `GIOCondition` and `G_IO_*`, which I just fixed, but I still need to fix `SKIP-DEFINITIONS:` by making it take the expected name, not the name specified in the GIR `name` attribute.

mrjbq7:
Are you waiting for me to review this?

PR #3102, opened 2024-11-27
discord: handle network disconnects smoother

olus2000:
Currently when a discord bot using `discord` vocab loses internet access it will get a DNS error when attempting to reconnect. This change catches and ignores this error, and adds a wait of a configurable amount of time (2 seconds by default) to every reconnect attempt.

To prevent future deadlocks the reconnect thread now puts a message in the mailbox in the exit handler.

Also includes some minor cleanup like replacing `json-request-headers` with `http-request-json` from the new `http.json` vocab.

### DIFF
--- a/extra/discord/authors.txt
+++ b/extra/discord/authors.txt
@@ -1,1 +1,2 @@
 Doug Coleman
+Aleksander "olus2000" Sabak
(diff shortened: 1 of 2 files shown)


erg:
let's merge this! thanks!

PR #3112, opened 2025-02-22
sequences.extras: add deep-nth and deep-nth-of

### LINKED ISSUE #3111: Consider adding something like `deep-nth` for drilling into nested seqs
Pursuant to a conversation that came in the discord...

`deep-of` exists in `assocs.extras` to "drill in" to nested assocs.

```factor
! nested hashes
H{ 
  { 1 H{ { 1 "1-1" } 
         { 2 H{ { 3 "1-2-3" } } 
         } 
       } 
  }
}
! get the value at key 1, then the value at key 2 from that result, and the value of key 3 of that result
{ 1 2 3 } deep-of
=> "1-2-3"
```

The notion came up of `deep-nth` (and/or `deep-nth-of`) that would work have similar behavior for sequences indices instead of assoc keys.

```factor
{ "zero" { "one-zero" "one-one" } } { 1 1 } deep-nth-of
=> "one-one"
```

From a naming perspective, there's maybe a little ambiguity... `assocs.extras` has a couple of `deep-` words that drill down into nested assocs. `sequences.deep` has `deep-` words that are maybe more like 'walking' the seq, recursing into nested seqs, so if someone sees a `deep-` word implemented on sequences, it could be a little confusing that the 'deep' `deep-nth` seems to have a different meaning from the 'deep' in e.g. `deep-map`.

bobisageek:
Closes #3111 

The actual code is quite small. I tried to word the help description as clearly as I could manage, but I'm very open to polish/advice on it.

### DIFF
--- a/extra/sequences/extras/extras-docs.factor
+++ b/extra/sequences/extras/extras-docs.factor
@@ -1067,6 +1067,68 @@ HELP: nth-of
 { $contract "Outputs the " { $snippet "n" } "th element of the sequence. Elements are numbered from zero, so the last element has an index one less than the length of the sequence. All sequences support this operation." }
 { $errors "Throws a " { $link bounds-error } " if the index is negative, or greater than or equal to the length of the sequence." } ;
 
+HELP: deep-nth
+{ $values
+    { "index-seq" sequence } { "nested-seq" sequence }
+    { "elt" object }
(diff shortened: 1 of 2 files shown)


mrjbq7:
Thanks!

PR #3120, opened 2025-05-17
Create factor

gamestime102:


mrjbq7:
???

gamestime102:
We want to develop factor code and upload it to different platforms.

mrjbq7:
But what is this change supposed to be?

gamestime102:
to make the factor code more understandable.

PR #3126, opened 2025-06-29
math.runge-kutta: Add adaptive step size, allow butcher tableau spec

inivekin:
I was doing something with this and wanted it to be faster/more accurate/more flexible.

In changing it, I used cleave-array which seems to make the example charting functions (`lorenz.` `rabinovich-fabrikant.` and `cyclically-symmetric.` ) not want to compile - I assume because  factor can't resolve the expected stack effects for the cleave.
The function contents work fine when pasted into the listener. I ask because I suspect misunderstanding the use of `inline` and there might be a magically simple solution. Is there a word to specify the call stack effect for a cleave or something?

### DIFF
--- a/extra/math/runge-kutta/examples/examples.factor
+++ b/extra/math/runge-kutta/examples/examples.factor
@@ -3,51 +3,72 @@ math.runge-kutta sequences ui.gadgets ui.gadgets.charts
 ui.gadgets.charts.lines ui.gadgets.panes ui.theme ;
 IN: math.runge-kutta.examples
 
-: lorenz-dx/dt ( tx..n -- dx )
-    rest first2
-    swap - 10 * ;
+: lorenz-dx/dt ( x..nt -- dx )
+    first2 swap - 10 * ;
 
-: lorenz-dy/dt ( tx..n -- dy )
-    rest first3
-    28 swap - swapd * swap - ;
+: lorenz-dy/dt ( x..nt -- dy )
+    first3 28 swap - swapd * swap - ;
 
-: lorenz-dz/dt ( tx..n -- dz )
-    rest first3
-    [ * ] dip 8/3 * - ;
+: lorenz-dz/dt ( x..nt -- dz )
+    first3 [ * ] dip 2.666 * - ;
 
-: <lorenz> ( -- dx..n/dt delta tx..n t-limit )
-    { [ lorenz-dx/dt ] [ lorenz-dy/dt ] [ lorenz-dz/dt ] } 0.01 { 0 0 1 21/20 } 150 ;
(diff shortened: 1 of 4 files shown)


mrjbq7:
Cool!

Will look in a week when I get back from a trip.

mrjbq7:
What's the relationship between the new code and the old ``runge-kutta-4``?  Is the new code generic across nth-order approximations?

inivekin:
Made some speed improvements by typing certain functions (hopefully reasonable?) and accuracy improvements by adding in the relative tolerance comparison I was lazy in leaving out.

Also did some reading of solver libraries in julia and nim to compare. Apparently people use Tsit5 or Vern6 more nowadays (useful site [here](https://www.sfu.ca/~jverner/)) - even still the implementations in those langs were surprisingly, disappointingly similar or worse in speed/accuracy compared to this one.

As an aside, found [this blog post](https://manasataramgini.wordpress.com/2020/06/07/a-simple-second-order-differential-equation-ovals-and-chaos/) that gave a pretty equations with different params and recreated them:

image

```
: ovals-dx/dt ( a x..nt -- dx )
    second * ;
:: ovals-dy/dt ( b c x..nt -- dy )
    b neg x..nt first 3 ^ *
    x..nt last 2 * pi * c / cos
    + ;
:: <ovals> ( a b c -- delta dx..n/dt inital t-limit )
    0.001
    { [ a swap ovals-dx/dt ]
      [ [ b c ] dip ovals-dy/dt ] }
    { 0.0 0.0 0.0  } 100 ;
: ovals. ( -- gadget )
    { 6 1.5 3.5
    4.5 3.5 3.5
    6.5 3.5 3.5
    8.5 3.5 2
    7 3.5 2
    7.5 3.5 2
    11.5 3.5 2.2
    13.75 1.5 2.2
    3.75 1.5 5.2
    4.75 1.125 5.2
    4.75 3.125 5.2
    1.75 8.125 5.2
    1.75 8.125 15.75
    6.75 8.125 15.75
    6.75 4 15.75
    6.75 4 5.75 } 3 <groups> 4 <groups> [ [ first3 <ovals> <runge-kutta> chart new { { -1.5 1.5 } { -1 1 } } >>axes swap line new link-color >>color swap { 2 } cols-except >>data add-gadget ] map ] map <grid> <scroller> white-interior open-window ;
```
20250828_18h49m52s_grim
I use an autoscaling axes chart gadget (which I keep meaning to but not make a pull for) hence it looking slightly different.

For the thing I was going to use this for, I am rethinking how I am approaching it, so probably won't do anything with this - no worries if you'd rather discard this pull.

Issue #2046, opened 2018-08-13
chloe should generate HTML5


bpollack:
There's a couple ways to approach this, but my current thinking is this: XHTML is actually wonderful for a dev; it's just not really supported anymore. Besides which, XHTML5 isn't getting the new hotness.

So what I'm thinking is that we Factor an XHTMLish-to-HTML5 converter, and then use that with Chloe. We'll keep all the benefits of our existing XML infra, but generate stuff modern browsers want to see. This is similar to what e.g. JSX or Elm already do, so it seems pretty reasonable to me. (And I already started down this path in 4ef42357e13ebe6debb2a3072f096c496ba27a38, so it's mostly an issue of seeing it through.)

PR #3129, opened 2025-07-22
html.templates.chloe.compiler: HTML-compliant rendering of raw text elements.

olus2000:
HTML `style` and `script` elements are [raw text elements](https://html.spec.whatwg.org/multipage/syntax.html#elements-2), meaning their contents can't contain character references (like `&amp;` or `&gt;`), only raw text. Previously Chloe template compiler escaped characters `&`, `>` and `<` inside them, breaking inline scripts and stylesheets. The direction of Chloe development according to #2046 is to move away from generating XML representation and towards generating HTML representation, so I added special handling of the tags corresponding to raw text elements that prevents escaping their contents.

Notably this behavior was already correct in scripts and stylesheets rendered using the boilerplate Chloe tags `t:style`, `t:script`, `t:write-style` and `t:write-script`.

With this change `>` can be used literally in the raw text elements. To get `<` and `&` their escape codes `&lt;` and `&amp;` can be used, since a Chloe template is an XML file and these escapes will be correctly parsed in all contexts; my changes only affect HTML output.

### DIFF
--- a/basis/html/templates/chloe/chloe-docs.factor
+++ b/basis/html/templates/chloe/chloe-docs.factor
@@ -112,7 +112,7 @@ ARTICLE: "html.templates.chloe.tags.control" "Control-flow Chloe tags"
 ARTICLE: "html.templates.chloe.tags.form" "Chloe link and form tags"
 "The following tags are only available if the " { $vocab-link "furnace.chloe-tags" } " vocabulary is loaded."
 { $table
-    { { $snippet "t:a" } { "Renders a link; extends the standard XHTML " { $snippet "a" } " tag by providing some integration with other web framework features. The following attributes are supported:"
+    { { $snippet "t:a" } { "Renders a link; extends the standard HTML " { $snippet "a" } " tag by providing some integration with other web framework features. The following attributes are supported:"
         { $list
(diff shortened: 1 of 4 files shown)


olus2000 (commented):
Some of my personal formatting preferences are showing, they should be corrected before merging. I'll wait with committing them for more comments.

mrjbq7:
I started in a different direction in e3e6a1284a75630b37ce1d32f23d9db7d13b581c.

The main idea is that if you want "raw output" you can just embed ``<![CATA[ lots of raw <=> output]]>`` in the chloe template.

And in script tags, I think it best to require the user to embed the script contents in a ``CDATA`` tag because otherwise it wouldn't necessarily be a valid XML document.

For characters that *should* escape, here is how the XML parser handles it:

```
IN: scratchpad "<script>></script>" string>xml pprint-xml
<?xml version="1.0" encoding="UTF-8"?>
<script>
  &gt;
</script>
```

So that means as a user of ``<t:script>`` you'd need to change

```
<t:script>var x = 1 < 2;</t:script>
```

to

```
<t:script><![CDATA[var x = 1 < 2;]]></t:script>
```

mrjbq7:
If you think we should still support ``script`` and ``style`` differently, let me know!

PR #3132, opened 2025-08-17
mime.multipart: fix buffer refill behavior

olus2000:
The changed part of `dump-until-separator` is supposed to write the rest of the buffer leaving only as many bytes as the length of the separator (minus 1) to ensure proper detection of a separator that falls on the buffer-size boundary. Currently it improperly calculates the index to split the buffer at to `lenght(separator) - length(buffer) - 1` instead of `length(buffer) - length(separator) - 1`. This is a fix to that issue.


mrjbq7:
Thanks!!!

PR #3134, opened 2025-08-20
Enumerations

LeoMehraban:
Added enumerations. Enumerations are defined in the `classes.enumeration` vocabulary, and are fully documented there

Enumerations work similarly to their equivalents in C and other languages, allowing for the easy specification of related ordered constants. In addition, they have support for non-numeric values, as long as the user specifies what "incrementing" them means.

You can see the documentation for more information. It should also be noted that existing documentation in certain places has been updated. This includes adding a reference to enumerations in the documentation for classes. When the term "enumeration" is used to describe the `enumerated` class in the documentation, that has been changed to "enumerated sequences", which means basically the same thing, but is less ambiguous now. The note in the existing `ENUM:` word in `alien.syntax` that mentions that `ENUM:` should not be used for factor types now references `ENUMERATION:`, as that's probably what the user was searching for when they searched for "ENUM:"


LeoMehraban:
For the third question, alien `ENUM:` uses c-types, and I wasn't sure how to properly make a compatibility layer to merge the two.
For the second question, I'm not sure how to do that.
For the first question, C enums not being namespaced often causes people to put the name of the enum in the name of the element, and if you're doing that already, that's barely any different from namespaced enums

mrjbq7:
Thank you!

There were some conflicts merging this branch, I rebased it on master and merged it manually.

A couple things!

1) I moved it to ``extra/`` for now to let it bake, it's fully usable by anyone downloading Factor since we include extra/, but its just a reminder to me that I want to poke at it later.

2) The tests didn't run as-is, just due to the move of the ``ENUMERATION:`` syntax to the classes.enumeration vocab you had done.

3) I recommend making a feature branch for pull requests, that way your ``master`` branch can more easily track the upstream ``master`` branch.  Otherwise it will start getting full of merge commits and not be ref-for-ref identical which can make your dev life a little more interesting sometimes.

mrjbq7:
I also tagged it with your name, current year, and BSD license text which is our requirement for merging. Thanks!

LeoMehraban:
If it's in extra, then we might have to remove the references to it in the documentation for classes and alien.syntax, and move the support for the see protocol into the vocabulary itself, rather than in the `see` vocabulary with all the other class types like it currently is. I probably should have made it more self-contained from the start

LeoMehraban:
Nevermind, you already did that

mrjbq7:
Just temporary in extra/ !

PR #3148, opened 2025-10-18
sequences: created (bounds-check?) generic word

LeoMehraban:
The creation of `(bounds-check?)` fixes a minor problem with the implementation of certain virtual sequences (notably `circular`), and increases the flexibility of the sequence protocol.

Here's an example of where it could help:

`circular` has an infinite length, but it also still needs to provide an implementation for the `length` word. It does this by returning the length of the underlying sequence. 
This created a problem with the use of words like `subseq`, where users couldn't access indices outside of the underlying sequence. This happened because `check-slice` used `length` to check if an index was valid.
Now, `check-slice` uses `bounds-check?` (via the helper word `at-or-within-bounds?`), which itself calls `(bounds-check?)`
`circular` now overrides `(bounds-check?)` to always return true, so you can always create subseqs or slices of it with any length

(The reason `(bounds-check?)` and `bounds-check?` are separate is because the latter is implemented as a generic word that dispatches on the numeric input. It seems like the only implementation of this is `integer`, ensuring that the second input is always an integer. This should maybe be simplified, but theoretically, something other than `integer` could use this functionality, so it's still in for now)

### DIFF
--- a/basis/circular/circular.factor
+++ b/basis/circular/circular.factor
@@ -17,7 +17,7 @@ TUPLE: circular < wrapped-sequence { start integer } ;
 PRIVATE>
 
 M: circular virtual@ circular-wrap seq>> ; inline
-M: circular (bounds-check?) 2drop t ; inline
+M: circular bounds-check? 2drop t ; inline
 M: circular minimum seq>> minimum ; inline
 M: circular maximum seq>> maximum ; inline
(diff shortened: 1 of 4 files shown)


mrjbq7:
Because sequences aren’t infinite.

PR #3153, opened 2025-11-22
Adds "monadics" and "reverse" Vocabs to extra

coltsingleactionarmyocelot:
Adds an alternative vocabulary for using Haskell-style monads, applicatives, and functors.
This also requires the addition of a "reverse" vocabulary based on Benediktsson's [recent blogpost](https://re.factorcode.org/2025/09/std-flip.html) as a dependency, perhaps that should be rolled into the shuffle vocabulary instead.


mrjbq7:
Cool!

PR #3171, opened 2026-01-31
peg: repeat1-related parsers handle repeated ignores

olus2000:
Current behavior of `repeat0`, `repeat1` and related parsers is to pass any `ignore` values generated by the encapsulated parsers in the resulting AST.

This pull request changes this behavior to be similar to the `seq` family of parsers which will exclude `ignore`s from their output.

Note that the behavior I implemented considers a `repeat1` parser which successfully matched its enclosing parser but got only `ignore`s a successful match nevertheless, creating a case where `repeat1` may produce an empty sequence AST. In my opinion if the content of `repeat1` matched the whole `repeat1` should be a success, no matter what the resulting AST is.

### DIFF
--- a/basis/peg/parsers/parsers.factor
+++ b/basis/peg/parsers/parsers.factor
@@ -32,8 +32,8 @@ M: list-parser parser-quot
     '[
         input-slice V{ } clone <parse-result>
         _ parse-seq-element [
-            [ @ _ [ f ] if ] swap repeat-loop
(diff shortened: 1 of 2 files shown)


mrjbq7:
Interesting! Was going to merge it, wanted to think about it for a moment.

mrjbq7:
Thanks!

PR #3180, opened 2026-05-29
combinators.syntax: changed definitions of ||[ and &&[ to improve usefulness and consistancy

LeoMehraban:
I saw someone in the discord talking about how they can't figure out how the `||[` and `&&[` words worked, as they were undocumented. I also didn't know what they did, so I looked into the code and I found that they didn't really work like I expected them to

I expected them to be either syntactic sugar for `0||` and `0&&` or for `1 n||` and `1 n&&`. Instead, they were seemingly far less useful, just being syntactic sugar for `dup [ drop ... ] when` and `dup [ drop ... ] unless`. I'm pretty sure the first one could be simply represented `[ ... ] [ f ] if`, and the second one as `[ ... ] unless*`, both being so simple that they didn't need syntactic sugar.

Furthermore, the `0||` and `0&&` combinators could perhaps use some syntactic sugar, and would benefit from the `... | ... ]` style of the other combinator syntax words. On top of that, this would be more consistent, as the relationship between `&[` and `n&[` changes the amount of inputs, not the amount of quotations, so why should the relationship between `&&[` and `n&&[` be any different?

So that's what I did, alongside making some duplicate syntax words into aliases (perhaps we don't even need them), adding some comments to clarify slightly confusing code, and documenting the words with missing documentation

### DIFF
--- a/extra/combinators/syntax/syntax-docs.factor
+++ b/extra/combinators/syntax/syntax-docs.factor
@@ -1,4 +1,4 @@
-USING: combinators.short-circut help.markup help.syntax combinators kernel generalizations ;
+USING: combinators.short-circuit help.markup help.syntax combinators kernel generalizations ;
 IN: combinators.syntax
 
 HELP: &[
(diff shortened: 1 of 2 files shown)


mrjbq7:
Awesome, thanks!!

AndydeCleyre:
I love it, thank you!

Minor: spelling of "separated" in the documentation has the common typo "seperated. "

PR #3181, opened 2026-06-01
fix: fixed ui build on freebsd

ronnie-w:
Removed gtk libs from the vm/Config.freebsd, which resulted in its removal as a hard dependency in every built executable. Fixed ui builds on freebsd by tweaking necessary platforms.txt files.


mrjbq7:
Merged, thanks!

PR #3184, opened 2026-07-05
db.tester: skip PostgreSQL tests when no database is configured

zolk3ri:
This PR adds a condition to skip PostgreSQL tests when no database is configured.

### Problem Description

Without a configured `postgresql-db`, `test-postgresql` calls
`postgresql-template1-db`, which clones the global (now `f`) and runs
`>>database` on it, throwing `Generic word database<< does not define a method
for the POSTPONE: f class`.

### Implementation Details

Extend the existing 32-bit-Windows skip guard in `test-postgresql` to also skip
when `postgresql-db` is unset. One guard covers `db.tuples`, `db.postgresql`,
and `db.postgresql.errors` since each wraps its PostgreSQL cases in
`test-postgresql`.

### Testing

Without `postgresql-db` set, the PostgreSQL cases are skipped, so `"db" test`,
`"db.tuples" test`, `"db.postgresql" test`, and `"db.postgresql.errors" test`
all pass. With it set (`\ postgresql-db set-global`), the new guard clause is
false and they run unchanged.

### DIFF
--- a/basis/db/tester/tester.factor
+++ b/basis/db/tester/tester.factor
@@ -1,4 +1,5 @@
 ! Copyright (C) 2008 Slava Pestov, Doug Coleman.
+! Copyright (C) 2026 Zoltán Kéri <z@zolk3ri.name>
 ! See https://factorcode.org/license.txt for BSD license.
 USING: accessors assocs concurrency.combinators db db.pools
 db.postgresql db.queries db.sqlite db.tuples db.types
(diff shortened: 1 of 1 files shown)


mrjbq7:
I'm not sure about this -- only because we don't currently have a way to report the difference between a skipped test and a passing one.

zolk3ri:
> I'm not sure about this -- only because we don't currently have a way to report the difference between a skipped test and a passing one.

Hmm, you are right that skipped and passed are indistinguishable here, but this PR does not introduce that, it just extends the existing skip condition. `test-postgresql` already skips silently on `os windows? cpu x86.32?` and this PR just ORs in one more predicate when `postgresql-db` isn't configured.

zolk3ri:
I am closing this PR because I realized the problem is broader than just PostgreSQL. The same issue applies to other vocabs that depend on external dependencies, such as PostgreSQL, Redis, Memcached, or FFI libraries, and each currently fails in a different way.

So rather than adding a PostgreSQL-specific solution, I would prefer to work on a more general mechanism that can represent tests which are unable to run due to missing prerequisites (similar to a "skipped" state) instead of reporting them as ordinary failures.

PR #3187, opened 2026-08-16
bencode: Sort dictionary keys when encoding

rgrinberg:
Bencode dictionaries must encode keys in sorted raw-string order, but the association encoder currently follows the association's iteration order. This produces invalid output for order-preserving associations such as linked assocs when their insertion order differs from lexical order.

Sort keys before encoding entries and add a regression test whose insertion order differs from the required wire order.

Specification: [BEP 3 — Bencoding](https://www.bittorrent.org/beps/bep_0003.html#bencoding)


mrjbq7:
Thanks!!

PR #3188, opened 2026-08-16
ascii: Handle empty strings in capitalize

rgrinberg:
`capitalize` lowercases its input and then unconditionally changes the character at index zero, causing an out-of-bounds error for an empty string.

Skip the indexed update when the lowercased string is empty, and add a regression test covering that input.


mrjbq7:
Thanks!

PR #3189, opened 2026-08-16
splitting: Preserve empty slices in split1-last-slice

rgrinberg:
`split1-last-slice` used `if-empty` to detect failed matches, conflating a valid empty prefix slice with `f` when the selected match starts at index 0. Check for `f` directly so boundary matches return empty slices while missing matches still return `seq f`.

Adds regression coverage for matches at the start with and without a trailing suffix.


mrjbq7:
Thanks!

PR #3190, opened 2026-08-16
sequences: Reject minimum on empty iota

rgrinberg:
`minimum` promises to throw for an empty sequence, but the constant-time `iota` specialization returned `0` even for `0 <iota>`.

Delegate the empty case to the general sequence method so it retains the documented error behavior, while preserving the constant-time result for nonempty iotas. Includes a regression test for the empty case.

### DIFF
--- a/core/sequences/sequences.factor
+++ b/core/sequences/sequences.factor
@@ -1139,7 +1139,8 @@ M: repetition sum [ elt>> ] [ length>> ] bi * ; inline
 
 GENERIC: minimum ( seq -- elt )
 M: sequence minimum [ ] [ min ] map-reduce ; inline
-M: iota minimum drop 0 ; inline
+M: iota minimum
+    dup empty? [ call-next-method ] [ drop 0 ] if ; inline
 M: reversed minimum seq>> minimum ; inline
 M: repetition minimum elt>> ; inline
 
--- a/core/sequences/sequences-tests.factor
+++ b/core/sequences/sequences-tests.factor
@@ -408,6 +408,7 @@ M: bogus-hashcode hashcode* 2drop 0 >bignum ;
 
 { 4 } [ 5 <iota> maximum ] unit-test
 { 0 } [ 5 <iota> minimum ] unit-test
+[ 0 <iota> minimum ] must-fail
 
 { 4 } [ 5 <iota> [ ] maximum-by ] unit-test
 { 0 } [ 5 <iota> [ ] minimum-by ] unit-test


mrjbq7:
Hmm, I'm not sure about this. Minimum on ranges of length zero should be well defined since they have a start value even if no sequence length.

mrjbq7:
Although I guess with no members of the sequence, even if mathematically there's a minimum, I could see throwing an error as reasonable.

rgrinberg:
I found it surprising that `n <iota> minimum` doesn't agree with `n <iota> >array minimum`. Although if this is the desired behavior, it might be best to take the test and discard the fix. As the tests demonstrates, others rely on this behavior.

core/sequences/sequences.factor
```diff
@@ -1139,7 +1139,8 @@ M: repetition sum [ elt>> ] [ length>> ] bi * ; inline
 
 GENERIC: minimum ( seq -- elt )
 M: sequence minimum [ ] [ min ] map-reduce ; inline
-M: iota minimum drop 0 ; inline
+M: iota minimum
+    dup empty? [ call-next-method ] [ drop 0 ] if ; inline
```
mrjbq7: perhaps instead something like

```factor
M: iota minimum first ;
```

Current:

```factor
IN: scratchpad M\ iota minimum optimized.
[ drop 0 ]
```

Your proposal:

```factor
IN: scratchpad M\ iota minimum optimized.
[
    dup 2 slot 0 dupd eq?
    [ drop \ t ] [ dup tag 0 eq? [ drop f ] [ 0 bignum= ] if ]
    if [
        dup >R 0 swap 2dup >R >R 2 slot dup tag 0 eq?
        [ fixnum< ] [ >R fixnum>bignum R> bignum< ] if
        [ \ t ] [ f ] if R> R> rot
        [ ( 10541395 10541396 -- 10541395 ) ] [ bounds-error ]
        if R> >R R> >R R> 2 slot >R 1 R> \ ( gensym ) [
            over over 2dup both-fixnums?
            [ fixnum< ] [
                dup tag 0 eq?
                [ fixnum>bignum bignum< ] [
                    over tag 0 eq?
                    [ >R fixnum>bignum R> bignum< ] [ bignum< ]
                    if
                ] if
            ] if [
                dupd >R >R swap >R R> swap min
                R> R> >R 1 over tag 0 eq?
                [ fixnum+ ] [ fixnum>bignum bignum+ ] if
                R> ( gensym )
            ] [ 2drop ] if
        ] label
    ] [ drop 0 ] if
]
```

My version (``first``):

```factor
IN: scratchpad M\ iota minimum optimized.
[
    0 swap 2dup >R >R 2 slot dup tag 0 eq?
    [ fixnum< ] [ >R fixnum>bignum R> bignum< ] if
    [ \ t ] [ f ] if R> R> rot
    [ ( 10542960 10542961 -- 10542960 ) ] [ bounds-error ] if
]
```

mrjbq7: Also, ``M\ iota maximum`` for empty iota has an issue too. Should fix those consistently.

Maybe

```factor
M: iota maximum last ;
```

mrjbq7:
Fixed in cdc003119b543dab7d1b4bcc36fbcc3b31618882.

Thanks!

PR #3191, opened 2026-08-17
uuid: Set RFC variant bits

rgrinberg:
UUIDv1, UUIDv7, and UUIDv8 currently emit variant bits other than the RFC UUID variant, producing fourth groups outside the required `8`, `9`, `a`, or `b` range.

Use a shared private `0b10` variant constant and add regression tests that inspect the two most significant bits of octet 8. RFC 9562 §4.1 requires these bits to be set to 1 and 0: https://www.rfc-editor.org/rfc/rfc9562.html#section-4.1


mrjbq7:
Thanks!

PR #3192, opened 2026-08-17
bit-arrays: Ignore padding bits when converting to integer

rgrinberg:
`bit-array>integer` currently decodes the entire backing byte array, allowing nonzero storage padding to change the logical value. This is observable after `set-bits` and for bit arrays populated through their raw storage.

Mask the decoded integer to the bit array's logical length, with regressions covering both cases.


mrjbq7:
Thanks!

PR #3193, opened 2026-08-17
http.server.cgi: Set CGI content type from metadata

rgrinberg:
CGI POST environments currently put the request body bytes in `CONTENT_TYPE`, because `cgi-variables` unwraps the `post-data` tuple before extracting both CGI values.

Keep the tuple long enough to read `content-type>>` for `CONTENT_TYPE` and use `data>>` only when calculating `CONTENT_LENGTH`. A regression test covers both values for a four-byte `text/plain` request.


mrjbq7:
Thanks!

PR #3194, opened 2026-08-18
http.server.requests: Require decimal Content-Length

rgrinberg:
The HTTP server currently passes `Content-Length` directly to Factor's general numeric parser, so values such as `0x7`, `+7`, and `7.0` can be accepted even though the HTTP field grammar permits decimal digits only.

Require every character to be an ASCII digit before parsing the value. Existing nonnegative and upload-limit validation remains unchanged. Regression tests cover the accepted non-decimal forms.


mrjbq7:
Thanks!

PR #3195, opened 2026-08-18
math.statistics: Apply ddof to covariance in corr-ddof

rgrinberg:
`corr-ddof` always used population covariance while applying the requested degrees-of-freedom adjustment to both variances. This caused `sample-corr` to scale correlation coefficients by `(n - 1) / n`.

Apply the same `ddof` to covariance so the scaling remains consistent, and correct the existing perfect-correlation expectations to `1.0` and `-1.0`.


mrjbq7:
Thanks!

PR #3198, opened 2026-08-18
json: Reject unknown escapes

rgrinberg:
The JSON reader currently accepts unknown backslash escapes such as `\q` and silently removes the backslash. JSON defines a closed set of valid string escapes, so unknown escapes must be rejected.

Return `f` from the escape dispatch's default branch so the existing `if*` failure path throws `json-error`. A regression test covers an unknown escape.

### DIFF
--- a/basis/json/json-tests.factor
+++ b/basis/json/json-tests.factor
@@ -39,6 +39,7 @@ IN: json.tests
 ! unicode is allowed in json
 { "ß∂¬ƒ˚∆" } [ "  \"ß∂¬ƒ˚∆\""  json> ] unit-test
 ${ { 8 9 10 12 13 34 47 92 } >string } [ " \"\\b\\t\\n\\f\\r\\\"\\/\\\\\" " json> ] unit-test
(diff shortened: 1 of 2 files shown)


mrjbq7:
I dunno.

Is that more strict than web browsers?

```javascript
> JSON.parse('{"name":"\q"}')
< {name: "q"}
```

rgrinberg:
Browsers are quite permissive and will accept a lot of JSON definitions that factor does not already. Single quoted strings and unquoted key names are some examples.

Do we want to relax our parser to accommodate those? Or is something that is done opportunistically?

mrjbq7:
I don't know. I might do a survey of what well known libraries do.

NoahTheDuke:
To help in the discussion: https://seriot.ch/security/parsing_json.html

mrjbq7:
If someone cares about this change, perhaps we can in the future add ``json-permissive-escapes?`` parameter or something. But I like it now. Thanks!

PR #3199, opened 2026-08-18
calendar: Fix start of quarter

rgrinberg:
`start-of-quarter` currently computes quarter start months as 0, 3, 6, and 9, and resets only the time of day. For example, February 15 becomes month zero on day 15 instead of January 1.

Use `start-of-month` to reset the day as well as the time, and add one to the zero-based quarter-month calculation to obtain months 1, 4, 7, and 10. A regression test covers both errors with a non-midnight timestamp.


mrjbq7:
Thanks!

PR #3200, opened 2026-08-18
math.statistics: Sum absolute errors

rgrinberg:
`sum-of-absolute-errors` currently sums signed deviations from the mean, causing positive and negative errors to cancel. For `{ 1 3 }`, it returns zero instead of two.

Take the absolute value of each deviation before `map-sum`. A regression test covers the cancellation case.


mrjbq7:
Thanks!

PR #3201, opened 2026-08-18
math.matrices: Use squared magnitudes in matrix L2 norm

rgrinberg:
`matrix-l2-norm` currently squares complex entries directly, which can make the resulting norm complex. Use squared magnitudes so the Frobenius norm remains a nonnegative real value, and clarify the documentation accordingly.


mrjbq7:
Thanks!

PR #3202, opened 2026-08-19
bencode: Reject trailing data

rgrinberg:
`bencode>` currently decodes the first bencoded object and silently drops any remaining input, accepting values such as `i42ei43e` and `i42eJUNK` as `42`.

Make the one-shot string and byte-array decoders verify EOF after `read-bencode`, while preserving `read-bencode` for stream-oriented parsing. Add regression coverage for concatenated objects and malformed suffixes, and document the stricter behavior.


mrjbq7:
Thanks!

PR #3204, opened 2026-08-19
ini-file: Escape vertical tabs correctly

rgrinberg:
Correct the INI escape table's vertical-tab entry and add regression coverage for backspace and vertical-tab serialization.


mrjbq7:
Oops! Thanks!!

PR #3205, opened 2026-08-19
bson: Reject trailing data

rgrinberg:
`bson>` silently ignored bytes after the first document. Require EOF after decoding one document while retaining `bson-sequence>` for concatenated documents.


mrjbq7:
Thanks!

PR #3206, opened 2026-08-19
cbor: Decode tag 33 as Base64url

rgrinberg:
Decode CBOR tag 33 text as unpadded Base64url and add a regression test.


mrjbq7:
Thanks!

PR #3207, opened 2026-08-19
bson: Reject invalid boolean bytes

rgrinberg:
Reject BSON boolean bytes other than 0 and 1, with regression coverage.


mrjbq7:
Thanks!

PR #3208, opened 2026-08-19
bencode: Reject truncated strings

rgrinberg:
`bencode>` accepted strings shorter than their declared length. Make payload reads throw `stream-exhausted` on premature EOF for string and byte-array inputs.


mrjbq7:
Thanks!

PR #3209, opened 2026-08-20
hex-strings: Reject invalid digits

rgrinberg:
Reject non-hexadecimal characters before decoding and add regression coverage.

### DIFF
--- a/basis/hex-strings/hex-strings.factor
+++ b/basis/hex-strings/hex-strings.factor
@@ -22,9 +22,12 @@ IN: hex-strings
 : sha512-string? ( str -- ? ) { [ length 128 = ] [ hex-string? ] } 1&& ;
 
 ERROR: invalid-hex-string-length n ;
+ERROR: invalid-hex-string string ;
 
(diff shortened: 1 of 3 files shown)


basis/hex-strings/hex-strings.factor
```diff
@@ -22,9 +22,12 @@ IN: hex-strings
...
 ERROR: invalid-hex-string-length n ;
+ERROR: invalid-hex-string string ;
 
 : hex-string>bytes ( hex-string -- bytes )
-    dup length dup even? [ invalid-hex-string-length ] unless 2/ <byte-array> [
+    dup length dup even? [ invalid-hex-string-length ] unless
+    over hex-string? [ drop invalid-hex-string ] unless
+    2/ <byte-array> [
         [
             [ digit> ] 2dip over even? [
```
mrjbq7: maybe better doing something here?

so it doesn't iterate twice over the sequence?

mrjbq7: do we care?

PR #3210, opened 2026-09-11
shuffle: move $shuffle help out of shuffle.factor

keiravillekode:
By using word-help* in shuffle-docs.factor, we remove shuffle.factor's dependency on the help system.


keiravillekode:
This will allow Exercism to retire the patch
https://github.com/exercism/factor-test-runner/blob/main/patches/shuffle-drop-help.patch
when we move to the next Factor tag.

mrjbq7:
Thanks!

PR #3212, opened 2026-09-13
zig vm: re-derive word in modify-code-heap

keiravillekode:
Previously we had

    if (word_after.pic_def != f)      jitCompileQuotation(word_after.pic_def)
    if (word_after.pic_tail_def != f) ...

jitCompileQuotation allocates and can therefore run a collection.

Found using Claude


mrjbq7:
Fixed in 8e16e7b2f1.

PR #3213, opened 2026-09-13
zig vm: write barrier in passArgsToFactor

keiravillekode:
The passArgsToFactor re-derives the OBJ_ARGS array from special_objects after each allotAlien, but it stored each alien with a plain

    data[i] = alien;

and no write barrier.

Found using Claude


keiravillekode:
When a unit test was requested:

> The passArgsToFactor test is in main.zig, and it found something important: the function re-fetches the args array before the allocation that can collect, so when that collection promotes the array the argument is stored into the dead nursery copy and lost. PR 3213's barrier is applied to that same stale slot, so it doesn't cover this case. I verified the test fails with 3213's line alone and passes when the array is re-derived after the allocation. I'd amend 3213 to allocate the alien first, then re-derive the array, store, and barrier.


Expect an update tomorrow.

mrjbq7:
Fixed in 8e16e7b2f1.

PR #3216, opened 2026-09-16
zig vm: fix format-float locale mask on glibc

### LINKED ISSUE #3215: Zig VM: format-float returns an empty string on Linux (glibc)
With Linux and glibc, examples like

```text
USING: formatting prettyprint ;
3.14159 "%.2f" sprintf .
```

give the expected results with the C++ VM but not with the Zig VM.

The suspected cause is that glibc `newlocale()` rejects  mask `0xFFF`

[LLM investigation notes](https://keiravillekode.github.io/factor-lang/zig-vm-format-float-locale)

keiravillekode:
Take LC_ALL_MASK from the system locale.h via @cImport instead of guessing it per platform (glibc: 0x1FBF, BSD/macOS: 0x3F).

Float formatting unit tests by Claude

Resolves #3215


mrjbq7:
Would you mind rebasing on latest master?

mrjbq7:
Thanks!

PR #3217, opened 2026-09-16
vm: fix bignum xor of a positive and a negative losing its top digit

keiravillekode:
bignum_posneg_bitwise_op sized its result as max(len1, len2 + 1) digits, but xor of a positive with a negative can need one digit more than the longer operand: (2^124 - 1) bitxor -1 is -2^124, whose magnitude needs three 62-bit digits while both operands fit in two. The two's complement result was all zeros over the allocated digits, so bignum_negate_magnitude dropped the carry out of the top digit and the primitive returned 0.

Found by Claude


mrjbq7:
Thanks!

PR #3218, opened 2026-09-16
vm: fix conversion to bignum of -2^63

keiravillekode:
We compute the magnitude of negative values by negating, and this led to overflows when negating -2^63 using signed types.

Found using Claude


mrjbq7:
Would you mind rebasing on latest master?

mrjbq7:
Thanks!

PR #3219, opened 2026-09-16
zig vm: remove two unused layouts accessors

keiravillekode:
Callstack.bottom and stringCapacity were unused.

If they had been called, type checking would have failed: `untagFixnumUnsigned` was needed instead of `untagFixnum`


keiravillekode:
Callstack.bottom is ported from an unused C++ VM member function `callstack::bottom()`

mrjbq7:
Thanks!

PR #3220, opened 2026-09-17
zig vm: retire unused label jump and call emitters

keiravillekode:
src/cpu.zig defines Arch enum containing x86 but
src/jit.zig does not handle the x86 enumeration value in
- emitJumpDirect()
- emitCallDirect()
- emitJumpPlaceholder()
- emitCallPlaceholder()

Compile errors would occur if there were any calls to their clients
- emitJumpToLabel()
- emitCallToLabel()

We retire all six functions.


keiravillekode:
Claude found this when asked to write unit tests.

mrjbq7:
Thanks!

PR #3221, opened 2026-09-17
zig vm: file primitives pop all for null handle

keiravillekode:
If the file handle is null, perhaps in a Windows GUI process, we still need to pop all operands from the stack.

- primitive_fread
- primitive_fwrite
- primitive_fputc
- primitive_fseek


mrjbq7:
Thanks!

PR #3222, opened 2026-09-17
vm: resume partial reads/writes at correct offset

keiravillekode:
In fread/fwrite retry loops, the offset `items * size` is in bytes, not in sizeof(int) steps.

- raw_fread
- factor_vm::safe_fwrite


keiravillekode:
Claude found and verified this while writing the unit tests that are in branch

https://github.com/keiravillekode/factor-lang/tree/vm-fread-fwrite-retry

`$ make -j8 CC=gcc CXX=g++ DEBUG=1 factor-test CONFIG=vm/Config.linux.x86.64 && ./factor-test --filter resumes`

Passes with commit 6dd0b39171 and fails without it.

mrjbq7:
Fixed in f62720cf97.

PR #3224, opened 2026-09-21
zig vm: link dynamically when targeting musl

keiravillekode:
`zig build -Dtarget=x86_64-linux-musl` previously produced a statically linked VM, because that is Zig's default for musl. Resolving FFI symbols involves `dlopen` calls; these failed.

glibc, macOS and Windows builds are dynamic already and are unaffected.


keiravillekode:
Note that various library-call changes are also needed, common across the C++ and Zig VMs.

An experiment using Claude:
https://github.com/factor/factor/compare/master...keiravillekode:factor-lang:linux-library-lookup

mrjbq7:
Thanks!

PR #3225, opened 2026-09-21
fixed find-library in freebsd

ronnie-w:
old behaviour: could not find library ie. libncursesw.so.9
errored while finding "lncursesw.9.so"

now matches "ncursesw" and gets the library name from the path


mrjbq7:
Thanks!

